PR #1, opened 2019-03-23
Initial wrapper work

rmelian:
### What?
This creates an initial Node.js parser. It's a wrapper for the C shared object exported in the Go parser.

### Notes
Currently, the binaries for Windows, Mac (darwin), and Linux are included in the `bin` directory. This makes the package size increase a lot, which is fine for now but we should look for another option in the future. Here are my suggestions:

* Compile things at `npm install` time using `node-gyp`. This is the first approach that I've tried first but didn't succeed. The downside of this approach is that people will need to have Python and Xcode (Mac), gcc in Linux, etc.
* Build 3 different npm packages for each platform. Potentially more than 3 if we add support for more platforms and archs (x32, x86, etc.)
* Download the right binary when installing the package (`npm install`) using a `pre-install` hook. The downside is that we might have people wanting to use this package in an environment that's not connected to Internet and use their own internal npm registry, like some CI systems.

Food for thought.


rmelian:
@fmvilas @MikeRalphson do you have an idea how to publish parser C libraries so that we can pull it from the parser?

What I'm thinking is to publish cparser libraries to a public place, then when installing dependencies for npm create a custom script to download those dependencies to local env.

we could use travis or similar tool to build platform-specific C libraries but then we need a place to store it.  maybe Github??

fmvilas:
To the best of my knowledge you can put them along with the npm package files. However, that would make the npm package very large and heavy to download, because you'll have all the platform binaries there. I've seen some packages have a compilation process during `npm install` but honestly I've no idea how this works or if it's even a good practice.

In any case, we could have all the binaries stored somewhere. Like you said, maybe a Github page and a CDN in front of it would be enough.

fmvilas:
That said, putting all the binaries in the package is not a bad idea as a first step. We can improve this part later.

fmvilas:
How is this going @rmelian?

fmvilas:
Updated the description of the PR with the following:

### What?
This creates an initial Node.js parser. It's a wrapper for the C shared object exported in the Go parser.

### Notes
Currently, the binaries for Windows, Mac (darwin), and Linux are included in the `bin` directory. This makes the package size increase a lot, which is fine for now but we should look for another option in the future. Here are my suggestions:

* Compile things at `npm install` time using `node-gyp`. This is the first approach that I've tried first but didn't succeed. The downside of this approach is that people will need to have Python and Xcode (Mac), gcc in Linux, etc.
* Build 3 different npm packages for each platform. Potentially more than 3 if we add support for more platforms and archs (x32, x86, etc.)

Food for thought.

rmelian:
developers will hate us if we make them to do this `Compile things at npm install time using node-gyp. This is the first approach that I've tried first but didn't succeed. The downside of this approach is that people will need to have Python and Xcode (Mac), gcc in Linux, etc.`

What about the idea of hosting ourselves the binaries and when compiling node we can crate a hook to download the corresponding binaries: I think you told me exatly this once.

fmvilas:
Yeah, forgot about this option too. We can have a pre-install hook that will download the right binary. Not sure, however, how will it affect people that want to run their CI systems without an Internet connection and using their own internal npm registry.

fmvilas:
Updated the description with your suggestion. Thanks, @rmelian!

PR #6, opened 2019-08-01
fix some issue encountered while testing on a browser

rmelian:
- fixed when no sending options parameter error "Error: Expected a file path, URL, or object. Got undefined    at Object.parse (parser.js:49)"
- Without this, it always breaks when finding ref to json-schema draft 04 for example
```javascript
"title": {
      "$ref": "http://json-schema.org/draft-04/schema#/properties/title"
 },
```
so I fixed it doing this
```javascript
const ajv = new Ajv({
    allErrors: true,
    schemaId: 'auto',
  });
  ajv.addMetaSchema(await getData('http://json-schema.org/draft-04/schema'));
```
I'm loading dynamically draft 04. I wanted to just show why this is needed here, so we can elaborate a better solution like packaging the schema similar to asyncapi .json files

- In the index I commented the raml imports as is causing an infinite loop at compile time and I can't figure it yet why(of course this is not intended to be merged this way)


fmvilas:
I've merged this branch with the original one because some of the issues have been fixed.

lib/parser.js
```diff
@@ -15,7 +15,7 @@ module.exports.parse = parse;
 module.exports.parseFromUrl = parseFromUrl;
 module.exports.registerSchemaParser = registerSchemaParser;
 
-async function parse(asyncapiYAMLorJSON, options = {}) {
+async function parse(asyncapiYAMLorJSON, options) {
```
fmvilas: This should default to `{}`. Check out the original (updated) branch. It checks for `options.path` instead of simply checking for `options`. There's now a unit test covering this case.

lib/parser.js
```diff
@@ -35,15 +35,26 @@ async function parse(asyncapiYAMLorJSON, options = {}) {
...
+    } catch (error) {
+      console.log(error);
+    }
+  };
+
   const ajv = new Ajv({
     allErrors: true,
+    schemaId: 'auto',
   });
+  ajv.addMetaSchema(await getData('http://json-schema.org/draft-04/schema'));
```
fmvilas: Can you provide a full example of when this is failing? I can't manage to make it fail.

fmvilas: Also, we should not go to the Internet to grab the schema. Ajv already got this packaged we just have to do `ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-04.json'));`. Otherwise, the parser will not work when offline.

fmvilas: Oh! I just saw it failing on Travis.

fmvilas: It's already fixed on the base branch. Thanks for raising the issue! I guess this PR can be closed now.

rmelian:
yes, I'm closing this

PR #12, opened 2019-08-14
Implement API

fmvilas:
Fixes #8 

**Note:** I recommend to start reviewing the PR from the README.md file, then the API.md, and then the code.

PR #13, opened 2019-08-14
add unit test for oneOf failing

rmelian:
@fmvilas this is to show you the issue with oneOf


fmvilas:
Not sure what the problem is but I just tested it on the #12 branch and it's not failing anymore. Thanks for this PR anyway 👍

PR #16, opened 2019-08-15
Add tests for errors and forward parsed JSON on errors

fmvilas:
Creates 2 new types of errors:

* `ParserErrorNoJS`: It means the parser was not able to convert the document to a JS object. It doesn't forward the parsedJSON object because it couldn't convert it.
* `ParserErrorUnsupportedVersion`: It means the document version is not supported. It forwards the parsedJSON object.

...and adds `parsedJSON` object to the existing `ParserError` error.


fmvilas:
Well, the reason we need to pass errors in `err.errors` is that some errors will be related to ajv and some will be related to other things, like converting YAML to JS for instance. Do you have another solution in mind?

aerfio:
No, now that `error.message` contains only your custom text like "Invalid AsyncAPI document" it makes perfect sense

LGTM

aerfio:
No, now that `error.message` contains only your custom text like "Invalid AsyncAPI document" it makes perfect sense

LGTM

PR #17, opened 2019-08-15
Export bundle.js

fmvilas:
Adds `browser` field to package.json.


aerfio:
LGTM from code point of view, make sure it doesn't crash in browser (using create-react-app for example)

I'd add some kind of prepublish hook to build `bundle.js` 
https://docs.npmjs.com/misc/scripts probably `prepublishOnly`

fmvilas:
That makes me think if we should just remove the bundle from version control. I think so, what do you think?

aerfio:
I would remove it, as it is generated on demand

PR #27, opened 2019-10-24
Add allMessages and allSchemas methods to AsyncAPIDocument model

### LINKED ISSUE #26: asyncapi.components().messages() is undefined when message ref'ed is in other file
I have in a template as follows:
```
{%- for messageName, message in asyncapi.components().messages() -%}
{{-messageName|log-}}
{%- endfor -%}
```
When executed on the following localref.yaml file:
```
asyncapi: '2.0.0-rc2'
id: 'urn:hello_world_publisher'
info:
  title: Hello world publisher application
  version: '0.1.0'
channels:
  hello:
    publish:
      message:
        $ref: '#/components/messages/hello-msg'
    x-protocolInfo:
      nats:
        pubSub:
components:
  messages:
    hello-msg:
      name: helloMsg
      payload:
        type: string
        pattern: '^hello .+$'
```
the output is:
```
hello-msg
```

But when executed on a globalref.yaml file, that references a message defined in another file, I get an error.
Content of the globalref.yaml file is:
```
asyncapi: '2.0.0-rc2'
id: 'urn:hello_world_publisher'
info:
  title: Hello world publisher application
  version: '0.1.0'
channels:
  hello:
    publish:
      message:
        $ref: './common.yaml#/components/messages/hello-msg'
```
Content of the common.yaml file is:
```
components:
  messages:
    hello-msg:
      name: helloMsg
      payload:
        type: string
        pattern: '^hello .+$'
```
The error shown is:
```
Something went wrong:
Template render error: (unknown path) [Line 3, Column 35]
  Error: Unable to call `the return value of (asyncapi["components"])["messages"]`, which is undefined or falsey
    at Object._prettifyError (/usr/lib/node_modules/asyncapi-generator/node_modules/nunjucks/src/lib.js:36:11)
```
I would expect the asyncapi.components().messages() to contain all messages, whether they are defined in the same file or in a different file.

There is a similar issue with asyncapi.components().schemas().

Using ag v0.16.0

fmvilas:
Adds the `allMessages` method to the AsyncAPIDocument model. This method returns a map with all the messages used in the parsed document. To uniquely identify the messages, it uses the property `name` or `x-parser-message-name` (autogenerated). If both properties are missing, it generates a base64 representation of the message.

Adds the `allSchemas` method to the AsyncAPIDocument model. This method returns a map with all the schemas used in the parsed document. To uniquely identify the schemas, it uses the property `$id` or `x-parser-schema-id` (autogenerated).

Fixes #26


fmvilas:
FYI @jdall

PR #30, opened 2019-12-06
Fixing message and schema keys

### LINKED ISSUE #29: allSchemas and allMessages does not return expected keys
Given the following input:
```yml 
 schemas:
    lightMeasuredPayload:
      type: object
      properties:
        lumens:
          type: integer
          minimum: 0
          description: Light intensity measured in lumens.
          x-pi: false
        sentAt:
          $ref: "#/components/schemas/sentAt"
    turnOnOffPayload:
      type: object
      properties:
        command:
          type: string
          enum:
            - on
            - off
          description: Whether to turn on or off the light.
          x-pi: false
        sentAt:
          $ref: "#/components/schemas/sentAt"
```
I would expect them to have the key `lightMeasuredPayload` and `turnOnOffPayload` however, they get these `anonymous-schema-x` names. 

The unit test for the method should also be changed. 
https://github.com/asyncapi/parser-js/blob/abf06b18e9942825e5342f386a57ef5e6101ec30/test/models/asyncapi_test.js#L172-L181

This problem is also present when calling `allMessages` function.

jonaslagoni:
Fixes #29


WaleedAshraf:
@jonaslagoni can you also bump version of `asyncapi` to `2.6.1` in this PR?

jonaslagoni:
Will do yea :)

jonaslagoni:
Not sure which project you mean, if it is the spec then you might need Fran to do it :)

WaleedAshraf:
This package: "asyncapi": "^2.6.0", to "asyncapi": "^2.6.1"

fmvilas:
Let's not do this in the same PR. If we end up having to rollback this PR we'll also lose the upgrade. It takes no time to create another one just for that.

fmvilas:
Sorry, closed the PR by mistake. Mobile UI is clunky.

jonaslagoni:
So I am unsure whether `outputWithOpenAPI` is correct in the test file `parse_test.js` for the test with `should parse OpenAPI schemas`. The payload contain the old `"x-parser-schema-id": "<anonymous-schema-1>"` and not the expected `"x-parser-schema-id": "testSchema"`. Where the `x-parser-original-payload` contain the new `"x-parser-schema-id": "testSchema"`. But I guess this comes from another npm package.

fmvilas:
Yes, this is coming from another package but it's fine.

fmvilas:
Looks great man! :+1:

PR #34, opened 2020-02-09
Add support for RAML data types

fmvilas:
It allows you to use RAML data types as message payloads. See `test/asyncapi-raml.yaml` to get an example.


fmvilas:
Yes, it's done here: https://github.com/asyncapi/asyncapi/pull/289.

PR #41, opened 2020-03-05
fix: allSchemas to return all schemas

jonaslagoni:
Solving https://github.com/asyncapi/parser-js/issues/39

PR #43, opened 2020-03-06
fix: add missing anonymous schema id's

### LINKED ISSUE #42: Schemas are not assigned correct uid's
Nested schemas without manually specifying the `$id` attribute never get assigned correct id's.

**Example**
```json
{
	"channels": {
		"test": {
			"parameters": {
				"testParam1": {
					"schema": {
						"type": "object",
						"properties": {
							"testParamNestedSchemaProp": {
								"type": "object",
								"properties": {
									"testParamNestedNestedSchemaProp2": {
										"type": "string"
									}
								}
							}
						}
					}
				}
			},
                        ...
		}
	}
}

```
given the above JSON I would expect the `$uid` of the schema to be `<anonymous-schema-1>`, `testParamNestedSchemaProp`, `testParamNestedNestedSchemaProp2`, however it is `<anonymous-schema-1>`, `undefined`, `undefined`.

One question, should the parameter schema not get assigned the uid `testParam1`?

jonaslagoni:
fixes #42


jonaslagoni:
I actually have not checked how it handles circle references should I create a test for that 🤔 ?

jonaslagoni:
Sorry, didn't see conflicts, give me 2 seconds.

jonaslagoni:
I absolutely hate rebasing 🗡

derberg:
@jonaslagoni is it like ready ready for review or better to leave it for Monday 😄 I say you pushed some additional things after requesting review. Maybe you want to work more on it over the weekend 😆

jonaslagoni:
@derberg it is ready, I just saw in the changes here on GitHub it had some weird formatting but seems like I couldn't change it 🤔 let me know if it is crucial and I'll take another look :)

lib/models/asyncapi.js
```diff
@@ -256,43 +272,96 @@ function assignNameToAnonymousMessages(doc) {
...
 }
 
+/**
+ * Gives schemas id to all anonymous schemas.
+ * 
+ * @param {AsyncAPIDocument} doc 
+ */
+function assignIdToAnonymousSchemas(doc) {
+  var anonymousSchemaCounter = 0;
+  let callback = (schema) => {
```
derberg, line 359:
```suggestion
  const callback = (schema) => {
```

lib/models/asyncapi.js
```diff
@@ -256,43 +272,96 @@ function assignNameToAnonymousMessages(doc) {
...
   }
 }
 
+/**
+ * Gives schemas id to all anonymous schemas.
+ * 
+ * @param {AsyncAPIDocument} doc 
+ */
+function assignIdToAnonymousSchemas(doc) {
+  var anonymousSchemaCounter = 0;
```
derberg, line 358:
```suggestion
  let anonymousSchemaCounter = 0;
```

lib/models/asyncapi.js
```diff
@@ -256,43 +272,96 @@ function assignNameToAnonymousMessages(doc) {
...
+            recursiveSchema(arraySchema, callback);
+          });
+        } else {
+          recursiveSchema(schema.items(), callback);
+        }
+        break;
+    }
+  } else {
+    //check for allOf, oneOf, anyOf
+    let checkCombiningSchemas = (combineArray) => {
```
derberg, line 306:
```suggestion
    const checkCombiningSchemas = (combineArray) => {
```

lib/models/asyncapi.js
```diff
@@ -256,43 +272,96 @@ function assignNameToAnonymousMessages(doc) {
...
+ *         the function that is called foreach schema found.
+ *         schema {Schema}: the found schema.
+ */
+function recursiveSchema(schema, callback) {
+  if (schema == null) return;
+  callback(schema);
+  if (schema.type() != null) {
+    switch (schema.type()) {
+      case 'object':
+        let props = schema.properties();
```
derberg, line 289:
```suggestion
        const props = schema.properties();
```

lib/models/asyncapi.js
```diff
@@ -256,43 +272,96 @@ function assignNameToAnonymousMessages(doc) {
...
+ * 
+ * @param {Schema} schema found.
+ * @param {Function} callback(schema)
+ *         the function that is called foreach schema found.
+ *         schema {Schema}: the found schema.
+ */
+function recursiveSchema(schema, callback) {
+  if (schema == null) return;
+  callback(schema);
+  if (schema.type() != null) {
```
derberg, line 286:
```suggestion
  if (schema.type() !== null) {
```

lib/models/asyncapi.js
```diff
@@ -256,43 +272,96 @@ function assignNameToAnonymousMessages(doc) {
...
+/**
+ * Recursively go through each schema and call callback.
+ * 
+ * @param {Schema} schema found.
+ * @param {Function} callback(schema)
+ *         the function that is called foreach schema found.
+ *         schema {Schema}: the found schema.
+ */
+function recursiveSchema(schema, callback) {
+  if (schema == null) return;
```
derberg, line 284:
```suggestion
  if (schema === null) return;
```

lib/models/asyncapi.js
```diff
@@ -256,43 +272,96 @@ function assignNameToAnonymousMessages(doc) {
...
+  callback(schema);
+  if (schema.type() != null) {
+    switch (schema.type()) {
+      case 'object':
+        let props = schema.properties();
+        for (const [_, propertySchema] of Object.entries(props)) {
```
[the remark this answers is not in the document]
jonaslagoni: I've started using `_` for variables that is not used such as here. Is that so bad? 🤔

jonaslagoni:
Thanks for the review!

> * extend `should return an array with all the schemas used in the document` with "nested json schema" case we discussed under the issue. or maybe a separate test?

I am thinking about implementing this in https://github.com/asyncapi/parser-js/pull/41

> * would be also great so see some tests that makes sure `assignUidToParameterSchemas` works as expected

I agree, I'll add this now.

derberg:
sorry @jonaslagoni that you had to ping me again and instead of approval I'm again back with a question, but didn't we agree in the issue that we should not assume id to be the same as the property key `"x-parser-schema-id": "streetlightId"` but all should be anonymous?

jonaslagoni:
@derberg no worries 😄 Yes, and I fail to see where the properties are being assigned to the property name instead of anonymous id? 

The current implementation should:
1. Assign anonymous id's recursively for all schemas regardless of it being a property in an object (which is as discussed in the issue).
2. Assign parameter names as ids which is what the new test checks. We do this for the same reason [schema](https://github.com/asyncapi/parser-js/blob/243a976844a3261db344278da7bc57357cd879fc/lib/models/asyncapi.js#L244) and [message](https://github.com/asyncapi/parser-js/blob/243a976844a3261db344278da7bc57357cd879fc/lib/models/asyncapi.js#L251) components gets their names assigned as ids.

derberg:
@jonaslagoni you are so right man, I should limit 🥃 in the late evenings

lib/models/asyncapi.js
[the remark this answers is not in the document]
derberg: you can just leave it empty afaik `[, propertySchema]`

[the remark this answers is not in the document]
jonaslagoni: Ohh, actually didn't knew that was possible 😄 Gonna change it.

jonaslagoni:
@derberg added your last changes 😄

derberg:
:shipit:

PR #44, opened 2020-03-09
Extract RAML and OpenAPI schema parsers to their own package

### LINKED ISSUE #48: ramldt2jsonschema package has link to fs module
The 'ramldt2jsonschema' module you are using has a link to the 'fs' module in the code. As a result, it turned out to be impossible to use parser version 0.14 with the asyncapi-react project when it is built only for the browser.

Is there anything you can do about this?

fmvilas:
It makes the bundle size of the parser much bigger. Next one could be the OpenAPI parser too.

The new package is here: https://www.github.com/asyncapi/raml-dt-schema-parser/

fixes #48


lib/index.js
```diff
@@ -19,8 +18,4 @@ parser.registerSchemaParser([
   'application/vnd.oai.openapi+yaml;version=3.0.0',
 ], openapiSchemaParser);
 
-parser.registerSchemaParser([
-  'application/raml+yaml;version=1.0',
-], ramlDtParser);
-
```
jonaslagoni: I can understand why we move the functionality for parsing RAML. However why do we remove the parser completely :thinking: ?

derberg: @jonaslagoni this is optional parser, so doesn't have to be here, it actually makes a lot sense to have all the additional parsers used as plugins. Look on readme here how you would use it https://github.com/asyncapi/raml-dt-schema-parser/

This change makes the parser very nice as you can decide not to use our parser for RAML but write your own and plug it in. Plus, the bundle size is getting smaller as you have in it only the required minimum

@fmvilas it shoudl definitely happen with openapi parser too, please do not forget to create issue.

package.json
```diff
@@ -15,6 +15,7 @@
   "author": "Fran Mendez <fmvilas@gmail.com> (fmvilas.com)",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@asyncapi/raml-dt-schema-parser": "^1.0.4",
```
derberg: why do we still test the RAML parser here and not https://github.com/asyncapi/raml-dt-schema-parser/ ?

lib/index.js
jonaslagoni: Alright understandable, do we have/need an issue in the generator which includes this new parser 🤔?

package.json
jonaslagoni: I actually don't mind it being testing here, great to have some integration tests. I guess we could actually duplicate this test in the RAML parser as well?

derberg: maybe as a short term solution, but ideally there should be a dummy parser there in the repo used for testing that also could be used as a reference in docs. So we at least need to have a followup issue for that to improve

jonaslagoni: :thinking: Yea you are right, would probably be best to use a dummy parser here and then in the RAML parser add this test :)

lib/index.js
fmvilas: What new parser, @jonaslagoni?

package.json
fmvilas: Good point. Moving the test.

lib/index.js
jonaslagoni: The optional RAML Schema parser that is being separated, I would imagine we had to include that [here ](https://github.com/asyncapi/generator/blob/master/lib/generator.js#L209) right?

package.json
fmvilas: @derberg @jonaslagoni https://github.com/asyncapi/raml-dt-schema-parser/pull/2

lib/index.js
fmvilas: Oh, good point! You're right. Doing it now. Thanks, mate!

package.json
fmvilas: Ok, finished moving tests (for OpenAPI too).

lib/index.js
fmvilas: Took the opportunity to extract the OpenAPI schema parser too: github.com/asyncapi/openapi-schema-parser.

derberg:
@fmvilas oh man, you already did OpenAPI plugin extraction. Well done!

lib/index.js
fmvilas: @jonaslagoni https://github.com/asyncapi/generator/pull/247 😄

fmvilas:
@jonaslagoni do you think it's fine now?

Issue #52, opened 2020-03-19
Automate release of parser and other packages


derberg:
- automate release as it is done already in generator https://github.com/asyncapi/generator/issues/225
- Important to remember:
  - no need for docker
  - remember to use npm scope
  - remember to mark in package.json that package is public
  - for browser distributions make sure that during the release bundles are attached to GitHub Release
  - I think CDN is not needed for now unless someone from community thinks different. In the age of HTTP/1 I would not use CDN but bundle the library but in HTTP/2 I would personally reconsider it. 
- except of parser we need to have automation in:
  - https://github.com/asyncapi/raml-dt-schema-parser
  - https://github.com/asyncapi/openapi-schema-parser
  - https://github.com/asyncapi/asyncapi-node
  - https://github.com/asyncapi/asyncapi-react/
  - https://github.com/asyncapi/parser-go
  - https://github.com/asyncapi/converter-go
  Generator templates:
  - https://github.com/derberg/html-template
  - https://github.com/derberg/markdown-template
  - https://github.com/derberg/nodejs-template
  - https://github.com/derberg/nodejs-ws-template
  - https://github.com/derberg/java-spring-template

derberg:
we still need to decide on:
- what about slack notifications on releases
- have twitter updates on releases
- incluse sonar cube as release step

derberg:
all the leftovers will be handled in separate issues

PR #53, opened 2020-03-19
feat: add configuration for automated releases

derberg:
Same as https://github.com/asyncapi/generator/pull/242 except of:
- I removed `dist` dir and added to ignore as since we will add it to release artifact, I do not think we should commit it to repository
- no docker release
- added step to run `npm test`
- avoided mistakes done with generator - package marked as public, readme updated, travis removed
- generating bundle.js and attaching to the GitHub release as an artifact

See also https://github.com/asyncapi/parser-js/issues/52


fmvilas:
> generating bundle.js and attaching to the GitHub release as an artifact

It's definitely more elegant 👍

PR #62, opened 2020-04-14
chore: add docs generation to release pipeline

derberg:
**Description**

- add docs generation to release pipeline so we do not have to remember about it on our own. Generated dosc will be pushed to repo with the same PR that will be created for package.json bump

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,8 @@ jobs:
         run: npm ci
       - name: Run tests
         run: npm test
+      - name: Regenerate docs
+        run: npm run docs
       - name: Generate bundle.js for the browser
         run: npm run prepublishOnly
       - name: Get version from package.json before release step
--- a/README.md
+++ b/README.md
@@ -84,9 +84,8 @@ Head over to [asyncapi/raml-dt-schema-parser](https://www.github.com/asyncapi/ra
 1. Run tests with `npm test`
 1. Write code and tests.
 1. Make sure all tests pass `npm test`
-1. Update bundle for client-side parser `npm run bundle`
 
-Release regenerates API documentation so you do not have to regenerate it manually with `npm run docs`.
+Release regenerates API documentation and browser bundle, so you do not have to regenerate it manually with `npm run docs` and `npm run prepublishOnly`.
 
 ## Contributing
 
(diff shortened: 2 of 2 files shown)


derberg:
@fmvilas @jonaslagoni done as discussed in other PR

README.md
```diff
@@ -84,9 +84,10 @@ Head over to [asyncapi/raml-dt-schema-parser](https://www.github.com/asyncapi/ra
 1. Run tests with `npm test`
 1. Write code and tests.
 1. Make sure all tests pass `npm test`
-1. Generate new API docs `npm run docs`
 1. Update bundle for client-side parser `npm run bundle`
```
jonaslagoni, line 87: Cant we include this in the workflow as well?

.github/workflows/release.yml
```diff
@@ -20,6 +20,8 @@ jobs:
         run: npm ci
       - name: Run tests
         run: npm test
+      - name: Regenerate docs
+        run: npm run docs
```
jonaslagoni, line 24: Is this enough? Shouldn't the workflow commit the changes caused by the command? 🤔

README.md
derberg: release one? it is already there, or?

.github/workflows/release.yml
derberg: this step will handle it like a boss
https://github.com/asyncapi/parser-js/pull/62/files#diff-56c30d9224897b08c28075052d4595c5R43

README.md
jonaslagoni: sorry, the `Update bundle for client-side parser `npm run bundle``

fmvilas: Then we can probably delete this line too, right?

jonaslagoni: Yea, it can be removed https://github.com/asyncapi/parser-js/blob/e5ffb4a01c5932c180aac969df29edfceda0253a/.github/workflows/release.yml#L23

.github/workflows/release.yml
jonaslagoni: Perfect 😄

README.md
derberg: yes, you are right, removing this line 
`npm run prepublishOnly` runs in release pipeline and it runs `npm run bundle`

derberg:
@jonaslagoni @fmvilas I pushed fix, please approve again as your approvals got dismissed after my push

PR #69, opened 2020-04-17
fix: dereference is failing with url references

fmvilas:
This is a temporary fix meanwhile the folks at @apidevtools merge [this pull request](https://github.com/APIDevTools/json-schema-ref-parser/pull/163) or another one solving the same problem.

The problem is that when a $ref like `https://myserver.com/mySchema.json` fails to download (either because it doesn't exist or another kind of errors), the information we get about the error is not that much. We get a message, a status, and some other information but we don't know which $ref provoked this error, i.e., the `https://myserver.com/mySchema.json`.

We need to know which $ref provoked the error so that we're able to locate it in the document and give a nice hint to the user on how and where the error is.

## Update

I closed my PR because it doesn't make sense anymore. However, I can't still stop pointing to my fork because of this: https://github.com/APIDevTools/json-schema-ref-parser/issues/164.


package.json
```diff
@@ -43,11 +43,11 @@
     "semantic-release": "^17.0.4"
   },
   "dependencies": {
+    "@apidevtools/json-schema-ref-parser": "github:fmvilas/json-schema-ref-parser",
```
fmvilas, line 46: This is pointing to my fork until the other PR gets merged.

jonaslagoni:
Already notified @fmvilas but the tests does not succeed in Windows. Just wanted to make sure it won't get merged untill fixed 😄

fmvilas:
@derberg Good call. I pointed to the branch on the fork. And agree about the TCK. It would be awesome to have it running on Github Actions too.

@jonaslagoni Tests are now fixed and taking into account EOL's format (`/r/n` or `/n`).

---

I also closed my PR because the Stoplight folks added a new feature to this library and, on the way, solved the problem I was fixing (perfect!). However, I'm still pointing to my fork because there's a bug on their implementation: https://github.com/APIDevTools/json-schema-ref-parser/issues/164. Therefore, I'll keep using my fork until it gets fixed and released.

PR #72, opened 2020-04-23
Fix: additional items and properties are never discovered and tests failing on windows

### LINKED ISSUE #71: Map/Dictionary Parsing Failing
#### Describe the bug
I am not 100% sure this is a bug but I've been able the whittle this down to a simple document to duplicate it.  I am only able to duplicate it using the CLI for generator, but I believe the issue is in the parser, hence I'm posting the issue here.  The issue does not show up when using the playground online.

When I try to generate either an html or MD template from the YML file below, I get the following error:
```
Something went wrong:
TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at recursiveSchema (/Users/ryan/.nvm/versions/node/v13.13.0/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/models/asyncapi.js:272:49)
    at /Users/ryan/.nvm/versions/node/v13.13.0/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/models/asyncapi.js:327:11
    at Array.forEach (<anonymous>)
    at /Users/ryan/.nvm/versions/node/v13.13.0/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/models/asyncapi.js:325:40
    at Array.forEach (<anonymous>)
    at schemaDocument (/Users/ryan/.nvm/versions/node/v13.13.0/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/models/asyncapi.js:311:24)
    at assignIdToAnonymousSchemas (/Users/ryan/.nvm/versions/node/v13.13.0/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/models/asyncapi.js:346:3)
    at new AsyncAPIDocument (/Users/ryan/.nvm/versions/node/v13.13.0/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/models/asyncapi.js:25:5)
    at parse (/Users/ryan/.nvm/versions/node/v13.13.0/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/parser.js:128:10)
```

The generator I'm using is 0.39.1.  I was able to hit the same issue with version 0.37.0.

I understand it very well could be that I'm not using the standard correctly.  The documentation on how to do a map/dictionary, while present, is hard to find.  And I believe there is not much regarding doing list/arrays.  In the end I need to do a `Map<String,List<Object>>` but this example is just `Map<String, String>` to keep it simple.

#### How to Reproduce
```yml
asyncapi: '2.0.0'
defaultContentType: application/json
info:
  title: Map Object Test
  version: '1'

channels:
  some_channel:
    subscribe:
      message:
        name: some_map
        payload:
          type: object
          additionalProperties:
            type: string
          #properties:
          #  name:
          #    type: string
```
If I comment the `additionalProperties` branch and uncomment the `properties` branch, it works.

jonaslagoni:
- fixes #71
- Solves a problems with additional items are never checked by the recursive schema function.
- Solved tests failing on windows but not on linux.

### DIFF
--- a/lib/models/asyncapi.js
+++ b/lib/models/asyncapi.js
@@ -172,6 +172,9 @@ class AsyncAPIDocument extends Base {
   allSchemas() {
     const schemas = new Map();
     const callback = (schema) => {
+      if(!schema){
+        console.log("shit")
+      }
       if (schema.uid()) {
         schemas.set(schema.uid(), schema);
       }
@@ -268,18 +271,31 @@ function recursiveSchema(schema, callback) {
   if (schema.type() !== undefined) {
     switch (schema.type()) {
       case 'object':
-        const props = schema.properties();
-        for (const [, propertySchema] of Object.entries(props)) {
-          recursiveSchema(propertySchema, callback);
+        if(schema.additionalProperties() !== undefined && typeof schema.additionalProperties() !== "boolean"){
+          var additionalSchema = schema.additionalProperties()
+          recursiveSchema(additionalSchema, callback);
+        }
+        if(schema.properties() !== null){
+          const props = schema.properties();
+          for (const [, propertySchema] of Object.entries(props)) {
+            recursiveSchema(propertySchema, callback);
+          }
         }
         break;
       case 'array':
-        if (Array.isArray(schema.items())) {
-          schema.items().forEach(arraySchema => {
-            recursiveSchema(arraySchema, callback);
-          });
-        } else {
-          recursiveSchema(schema.items(), callback);
+        if(schema.additionalItems() !== undefined){
+          var additionalArrayItems = schema.additionalItems()
+          recursiveSchema(additionalArrayItems, callback);
+        }
+
+        if(schema.items() !== null){
+          if (Array.isArray(schema.items())) {
+            schema.items().forEach(arraySchema => {
+              recursiveSchema(arraySchema, callback);
+            });
+          } else {
+            recursiveSchema(schema.items(), callback);
+          }
         }
         break;
     }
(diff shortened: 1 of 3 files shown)


jonaslagoni:
@ryanskinner do you have time to check if this fixes your overall use-case?

lib/models/asyncapi.js
```diff
@@ -268,18 +268,31 @@ function recursiveSchema(schema, callback) {
   if (schema.type() !== undefined) {
     switch (schema.type()) {
       case 'object':
-        const props = schema.properties();
-        for (const [, propertySchema] of Object.entries(props)) {
-          recursiveSchema(propertySchema, callback);
+        if(schema.additionalProperties() !== undefined && typeof schema.additionalProperties() !== "boolean"){
+          var additionalSchema = schema.additionalProperties()
```
fmvilas, line 272: Always prefer `let` or `const`. The scope of `var`s is the whole function which may lead to undesired behaviors in the future when someone "touches" this code.

```suggestion
          const additionalSchema = schema.additionalProperties()
```

lib/models/asyncapi.js
```diff
@@ -268,18 +268,31 @@ function recursiveSchema(schema, callback) {
...
         break;
       case 'array':
-        if (Array.isArray(schema.items())) {
-          schema.items().forEach(arraySchema => {
-            recursiveSchema(arraySchema, callback);
-          });
-        } else {
-          recursiveSchema(schema.items(), callback);
+        if(schema.additionalItems() !== undefined){
+          var additionalArrayItems = schema.additionalItems()
```
fmvilas, line 284:
```suggestion
          const additionalArrayItems = schema.additionalItems()
```

PR #82, opened 2020-05-26
update: tags() to map subscribe/publish tags.

sebastian-palma:
The current code to get the tags from a document only takes into account those tags in the root of the document:

```js
this._json.tags.map(t => new Tag(t))
```

Here it's supposed `tags` is at the same level of `info`, `channels`, `servers`,  etc. So, tags in channels aren't being retrieved when asking for them, nor does `hasTags` when checking if the document contains tags.

I've updated the code for `tags()` and `hasTags()` - the latter uses `tags()` to check if any tag is present in the document (which I don't know if it's okay).

By now tags are only being "extracted" from publish/subscribe entries in channels, so we can see where else could be tags and add them to `tags()` as well.

### DIFF
--- a/lib/models/asyncapi.js
+++ b/lib/models/asyncapi.js
@@ -124,15 +124,26 @@ class AsyncAPIDocument extends Base {
    * @returns {boolean}
    */
   hasTags() {
-    return !!(this._json.tags && this._json.tags.length);
+    return !!(this.tags() && this.tags().length);
   }
 
   /**
    * @returns {Tag[]}
    */
   tags() {
-    if (!this._json.tags) return [];
-    return this._json.tags.map(t => new Tag(t));
+    const tags = new Set();
+
+    this.channelNames().forEach(channelName => {
+      const channel = this.channel(channelName);
+      if (channel.hasPublish()) {
+        (channel._json.publish.tags || []).forEach(tag => tags.add(tag.name));
+      }
+      if (channel.hasSubscribe()) {
+        (channel._json.subscribe.tags || []).forEach(tag => tags.add(tag.name));
+      }
+    })
+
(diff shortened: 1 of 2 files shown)


derberg:
@sebastian-palma hey, sorry it took so long to look at your PR.
Are you sure it is needed? 
In your PR you are extending https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js which is a Model for root elements of the spec, and `tags` related method should return only root level tags.

Messages, and also Operations that you now want to add support for, already have proper methods for `tags` and they work as expected. For example, have a look at https://github.com/asyncapi/parser-js/blob/master/lib/models/operation.js#L10 that extends an object with `tags` related methods. You can see it used in action https://github.com/asyncapi/html-template/blob/master/partials/operation.html#L64

what I'm trying to say is that `asyncapi.tags()` should stay as it is and return only root level tags.
we need to find a different solution here 🤔 

I honestly think the issue is only in https://github.com/asyncapi/html-template/blob/master/partials/sidebar.html#L33. We just need to get rid of those `asyncapi.tags()` and `asyncapi.hasTags()`  related things as it should not be there at all. It is about building left navigation for operations, so it should be only dependent on operations and the fact if they have tags or not.

What are your thoughts on that?

sebastian-palma:
Hey @derberg, that completely makes sense. If the asyncapi model is for root elements, then it should remain as it's, as you say.

We can close this PR and open a new related one in the html-template repo. I'm gonna dig a bit more on the problem.

Thanks!

derberg:
@sebastian-palma awesome man, lemme know whenever you need a hand, you can also ping me on one of our channels on slack if ya want

PR #85, opened 2020-06-12
feat: Add hasMessages function to main model

### LINKED ISSUE #84: Add method .hasMessages()
Add method `asyncapi.hasMessages()` like we have others, `asyncapi.hasChannels()`.
This will be used to check if messages are defined, and then show them in a separate section.

To be used here:
https://github.com/asyncapi/html-template/pull/31#pullrequestreview-422044695

Related PR: https://github.com/asyncapi/html-template/pull/31

derberg:
**Description**

- add `hasMessages()` function to main model. You can easily evaluate if specification file has any messages, in components or inside the channels operations

**Related issue(s)**
Resolves #84

### DIFF
--- a/lib/models/asyncapi.js
+++ b/lib/models/asyncapi.js
@@ -136,6 +136,13 @@ class AsyncAPIDocument extends Base {
   }
 
   /**
+   * @returns {boolean}
+   */
+  hasMessages() {
+    return !!this.allMessages().size
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


fmvilas:
I guess a `.components().hasMessages()` would be great too. In other PR, of course.

derberg:
@fmvilas should then there be also `hasSchemas`, `hasSecuritySchemas` etc?

PR #91, opened 2020-06-19
feat: add server variables and channel parameters validation

### LINKED ISSUE #24: MIssing validation of channel parameters
It would be nice if the parser could validate channel parameters
Having e.g.
```
channels
  playground/testParams/{param1}
```   
without the parameters being specified like e.g.
```
    parameters:
      - name: param1
        schema:
          type: string
```
should result in the parser bailing out - with a useful error message.
Please consider this.
Thanks.

derberg:
**Description**

- Add validation for server.url if it has variables that are not described in the variable object
- Add validation for channel name if it has parameters that are not described in the parameter object

**Related issue(s)**
Resolves #24


package.json
```diff
@@ -50,11 +50,12 @@
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "github:fmvilas/json-schema-ref-parser#add-more-info-about-errors",
     "@asyncapi/specs": "^2.7.1",
-    "@fmvilas/pseudo-yaml-ast": "^0.3.0",
+    "@fmvilas/pseudo-yaml-ast": "^0.3.1",
     "ajv": "^6.10.1",
     "js-yaml": "^3.13.1",
     "json-to-ast": "^2.1.0",
     "node-fetch": "^2.6.0",
+    "re2": "^1.15.0",
```
derberg, line 58: added to have secure handling of regex as suggested by sonar

lib/utils.js
```diff
@@ -219,16 +219,67 @@ utils.findRefs = (json, absolutePath, relativePath, initialFormat, asyncapiYAMLo
...
   if (!ast) return { jsonPointer };
+
   return findLocationOf(jsonPointerToArray(jsonPointer), ast, initialFormat);
 };
 
 utils.improveAjvErrors = (errors, asyncapiYAMLorJSON, initialFormat) => {
   const ast = getAST(asyncapiYAMLorJSON, initialFormat);
   return errors.map(error => {
     const defaultLocation = { jsonPointer: error.dataPath || '/' };
+    console.log(error.dataPath);
```
fmvilas, line 230: 🚜 ...

lib/utils.js
```diff
@@ -219,16 +219,67 @@ utils.findRefs = (json, absolutePath, relativePath, initialFormat, asyncapiYAMLo
...
     };
   });
-};
\ No newline at end of file
+};
+
+/**
+ * Returns an array with all strings in between curly braces, including braces
+ */
+utils.getCurlyBracesEl = str => {
```
fmvilas, line 241: Can we come up with a better name for this function? Outside the context of this PR, I'd not really understand what's doing.

lib/utils.js
```diff
@@ -219,16 +219,67 @@ utils.findRefs = (json, absolutePath, relativePath, initialFormat, asyncapiYAMLo
...
+  if (typeof str !== 'string') return;
+  const regex = new RE2(/{(.+?)}/g);
+
+  return regex.match(str); 
+};
+
+/**
+ * Returns an array with values that map to the keys of provided object
+ */
+utils.groupExistingElements = (arr, obj) => {
```
fmvilas, line 251: Same as above. It's not clear what's the purpose of this function. Also, it seems you named this function when you thought it was going to be only an array of existing elements but now it also includes `noEl`, which I don't understand what it is.

lib/utils.js
```diff
@@ -219,16 +219,67 @@ utils.findRefs = (json, absolutePath, relativePath, initialFormat, asyncapiYAMLo
...
+ */
+utils.getCurlyBracesEl = str => {
+  if (typeof str !== 'string') return;
+  const regex = new RE2(/{(.+?)}/g);
+
+  return regex.match(str); 
+};
+
+/**
+ * Returns an array with values that map to the keys of provided object
```
fmvilas, line 249: It says it returns an array but it never does it.

lib/utils.js
```diff
@@ -219,16 +219,67 @@ utils.findRefs = (json, absolutePath, relativePath, initialFormat, asyncapiYAMLo
...
+  });
+
+  return groupedElements;
+};
+
+/**
+ * Returns array of errors for channel params and server variables validation
+ */
+utils.groupValidationErrors = (root, errorMessage, errorElements, asyncapiYAMLorJSON, initialFormat) => {
+  const error = [];
```
fmvilas, line 274:
```suggestion
  const errors = [];
```

lib/utils.js
```diff
@@ -219,16 +219,67 @@ utils.findRefs = (json, absolutePath, relativePath, initialFormat, asyncapiYAMLo
...
+  const regex = new RE2(/\//g);
+
+  Object.keys(errorElements).forEach((el) => {
+    error.push({
+      title: `${el} ${errorMessage}: ${errorElements[el]}`,
+      location: utils.getLocationOf(root + el.replace(regex, '~1'), asyncapiYAMLorJSON, initialFormat)
+    });
+  });
+
+  return error;
```
fmvilas, line 284:
```suggestion
  return errors;
```

lib/utils.js
derberg: 😱

lib/utils.js
derberg: I hate reading my code a week after writing it and not understanding my intentions...
anyway 😅  fixed the code that was overcomplicated

lib/customValidators.js
```diff
@@ -0,0 +1,62 @@
...
+  if (!srvs) return true;
+
+  const notProvidedVariables = {};
+
+  Object.keys(srvs).forEach((server) => {
+    const variables = parseUrlVariables(srvs[server].url);
+    if (!variables) return;
+
+    const missingServerVariables = getMissingProps(variables, srvs[server].variables);
+    if (missingServerVariables.length) notProvidedVariables[server] = notProvidedVariables[server] ? notProvidedVariables[server].concat(missingServerVariables) : missingServerVariables;
```
fmvilas, line 15: I'm surprised Sonar is not warning you about this. Assigning properties this way is not a good practice because you can unexpectedly override "special" object properties. For instance, consider we have a variable called "constructor". This line would be doing:

```js
notProvidedVariables['constructor'] = missingServerVariables;
```

You would be changing the constructor of the object and that can cause undesired problems. See [The Dangers of Square Bracket Notation](https://github.com/nodesecurity/eslint-plugin-security/blob/master/docs/the-dangers-of-square-bracket-notation.md).

fmvilas: That said, I'd use arrays and leverage Array.find which is performat enough, especially for small arrays.

lib/customValidators.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  return true;
+}
+  
+function validateChannelParams(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
+  const chnls = parsedJSON.channels;
+  if (!chnls) return true;
+
+  //let providedParams = [];
```
fmvilas, line 34: 🚜 ...

lib/customValidators.js
```diff
@@ -0,0 +1,62 @@
...
+  const notProvidedParams = {};
+    
+  Object.keys(chnls).forEach((channel) => {
+    const variables = parseUrlVariables(channel);
+    if (!variables) return;
+
+    const missingChannelParams = getMissingProps(variables, chnls[channel].parameters);
+  
+    //if (declaredChannelParams.existEl.length) providedParams = providedParams.concat(declaredChannelParams.existEl);
```
fmvilas, line 43: 🚜 ...

lib/customValidators.js
```diff
@@ -0,0 +1,62 @@
...
+  const notProvidedParams = {};
+    
+  Object.keys(chnls).forEach((channel) => {
+    const variables = parseUrlVariables(channel);
+    if (!variables) return;
+
+    const missingChannelParams = getMissingProps(variables, chnls[channel].parameters);
+  
+    //if (declaredChannelParams.existEl.length) providedParams = providedParams.concat(declaredChannelParams.existEl); 
+    if (missingChannelParams.length) notProvidedParams[channel] = notProvidedParams[channel] ? notProvidedParams[channel].concat(missingChannelParams) : missingChannelParams;
```
fmvilas, line 44: Same as above.

lib/utils.js
```diff
@@ -219,16 +219,58 @@ utils.findRefs = (json, absolutePath, relativePath, initialFormat, asyncapiYAMLo
...
+
+  if (!obj) return arr;
+  
+  return arr.filter(val => {
+    return !obj.hasOwnProperty(val);
+  });
+};
+
+/**
+ * Returns array of errors for channel params and server variables validation
```
fmvilas, line 262: This function is no longer **only** for channel params and server variables.

lib/customValidators.js
fmvilas: We can probably add this: https://github.com/nodesecurity/eslint-plugin-security.

lib/utils.js
derberg: what ya mean? at the moment it is used only for channels params and server variables. If we see it is useful in other validations then we should change it, or I'm missing the point?

fmvilas: My point is this is the description of the function. It says what it does. IMO it doesn't have to say where it's used, given this can change in the future and it's easy to forget changing this description. I'd just make the description say "what this function does". Also, the `utils.js` file is for generic functions. If it's not generic, it should probably be in the `lib/customValidators.js` file. What do you think?

lib/customValidators.js
derberg: `notProvidedVariables` is an internal object of this function and doesn't have any constructors, it's purpose is to provide a structured information about not provided params/variables so later it is much easier to extract the information in  `groupValidationErrors` function. Using array here adds unnecessary complexity.

the plugin you suggested throws 
Screenshot 2020-06-30 at 17 06 46
and I believe it will be more annoying than useful, its fails to bring good info.

Even if someone would really provide values in AsyncAPI document that are hidden functions or something, looking at my code I don't see what could go wrong really.

lib/utils.js
derberg: ah, I know what you mean now, you're right, description should only say what function is doing of course. Will change, thanks. As if it comes to location, `utils.js` is fine as this function is already reused in two other functions and will be easier to notice in `utils.js`

derberg: done

lib/customValidators.js
fmvilas: That's ok but consider it a code smell (something that in the future could potentially become problematic). Right now the variable is contained in this function and we only return true or false but a change in the future might make us think it's better to return the object instead of a boolean and suddenly it goes beyond the scope of the function.

> `notProvidedVariables` is an internal object of this function and doesn't have any constructors

Just a note: every variable in JS has a constructor.
Screen Shot 2020-07-01 at 13 06 06

If you want to get rid of this "issue", consider using a Map: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map.

Anyway, not a nazi of these things myself, just consider it.

> and I believe it will be more annoying than useful, its fails to bring good info.

It's annoying because I'm doing the same thing in other places and I probably shouldn't. We can always add `// eslint-ignore` to the line.

derberg: @fmvilas you are right, I just pushed a change and after working on refactoring I have to say I prefer working with Map than an Object really.

I would not introduce the linter though as it is still failing in many other places and I should not fix them in this pr and I'm skeptical if it comes to `ignoring`. I would prefer to have another issue for that, and if it is not in scope for 1.0 it could be marked with `good first issue` as because of the size of the changes it is a good task to get familiar with the project

derberg: https://github.com/asyncapi/parser-js/issues/103

PR #92, opened 2020-06-23
feat: registation of parsers without providing mime types

### LINKED ISSUE #47: Make parser auto-discover schema parsers mime-types
While I was working on extracting RAML and OpenAPI schema parsers (#44), I realized that it's a bit painful to remember all the mime types that a given schema parser supports. It would be great to make the parser auto-discover these mime types.

For instance, instead of writing this:

```js
parser.registerSchemaParser([
  'application/vnd.oai.openapi;version=3.0.0',
  'application/vnd.oai.openapi+json;version=3.0.0',
  'application/vnd.oai.openapi+yaml;version=3.0.0',
], openapiSchemaParser);
```

You write this:

```js
parser.registerSchemaParser(openapiSchemaParser);
```

This will improve a few things:
1. You will no longer have to remember the list of mime types.
1. It will help with evolution. E.g., if a parser supports a new mime type, you'll only have to update to the latest version.
1. It's less error-prone.

## How?

By enforcing a common interface on schema parsers, e.g.:

```js
openapiSchemaParser.parse = function ({ message, defaultSchemaFormat }) { ... }
openapiSchemaParser.getMimeTypes = function () { ... }
```

derberg:
**Description**

- enforce parsers to provide 2 functions, `parse` and `getMimeTypes`
- added new error type
- solved bug that parser registration would throw a meaningless error in case someone did not provide the function to registration
- some super basic test to make sure errors are thrown properly

**OtherPRs**
- https://github.com/asyncapi/raml-dt-schema-parser/pull/6
- https://github.com/asyncapi/openapi-schema-parser/pull/4
- https://github.com/asyncapi/generator/pull/384 -> all parser changes were manually tested here to make sure all works well

**Breaking Change**
This is a breaking change. Custom message parsers must now implement `parse` and `getMimeTypes` functions:

```
myParser.parse = function ({ message, defaultSchemaFormat }) { ... }
myParser.getMimeTypes = function() { return [ ... ] }
```

so the registration looks just like this:
```
parser.registerSchemaParser(openapiSchemaParser);
```

**Related issue(s)**
Resolves #47


README.md
```diff
@@ -79,6 +79,28 @@ Head over to [asyncapi/openapi-schema-parser](https://www.github.com/asyncapi/op
 
 Head over to [asyncapi/raml-dt-schema-parser](https://www.github.com/asyncapi/raml-dt-schema-parser) for more information.
 
+### Custom message parsers
+
+AsyncAPI doesn't enforce one schema format for messages. You can have payload of your messages described with OpenAPI or Avro. This parser by default parses only AsyncAPI schema format. You can extend it by creating a custom parser and registering it withing the parser:
```
fmvilas, line 84: Looks like you only have 2 options. What about this:

```suggestion
AsyncAPI doesn't enforce one schema format for messages. You can have payload of your messages described with OpenAPI, Avro, etc. This parser by default parses only AsyncAPI schema format. You can extend it by creating a custom parser and registering it withing the parser:
```

README.md
```diff
@@ -79,6 +79,28 @@ Head over to [asyncapi/openapi-schema-parser](https://www.github.com/asyncapi/op
...
+
+1. Create custom parser module that exports two functions:
+    ```js
+    module.exports = {
+      parse: ({ message, defaultSchemaFormat }) => { //custom parsing logic},
+      getMimeTypes: () => [
+        '//mime types that will be used in the AsyncAPI document to specify mime type of given message',
```
fmvilas, line 91:
```suggestion
        '//mime types that will be used as the `schemaFormat` property of the message to specify its mime type',
```

README.md
```diff
@@ -79,6 +79,28 @@ Head over to [asyncapi/openapi-schema-parser](https://www.github.com/asyncapi/op
...
+        'application/vnd.custom.type;version=1.0.0',
+        'application/vnd.custom.type+json;version=1.0.0',
+      ]
+    }
+    ```
+2. Before parsing AsyncAPI document with a parser, register additional custom schema parser:
```
fmvilas, line 97:
```suggestion
2. Before parsing an AsyncAPI document with a parser, register the additional custom schema parser:
```

PR #94, opened 2020-06-26
feat: handling circular references

### LINKED ISSUE #83: Recursive messages cause the parser to stack overflow
#### Describe the bug
I need to define messages that are recursive tree structures. When I do this and attempt to use the `@asyncapi/generator`, to generate documentation (or code) for my spec, the parser fails with a `Error: Maximum call stack size exceeded` message.

#### How to Reproduce
Here's a minimal asyncapi spec that reproduces the issue:
```yaml
asyncapi: '2.0.0'
info:
  title: Example
  version: 0.1.1
channels:
  recursive:
    subscribe:
      message:
        payload:
          $ref: '#/components/schemas/Recursive'
components:
  schemas:
    Recursive:
      type: object
      properties:
        children:
          type: array
          items:
            $ref: '#/components/schemas/Recursive'
        # This also causes a stack overflow
        # child:
        #   $ref: '#/components/schemas/Recursive'
        something:
          type: string
```
```
$ npx -p @asyncapi/generator ag ./recursive.yaml @asyncapi/html-template --force-write -o docs
npx: installed 213 in 8.774s
Something went wrong:
Error: Maximum call stack size exceeded
    at parse (~/.npm/_npx/32509/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/parser/lib/parser.js:121:11)
    at async Generator.generateFromString (~/.npm/_npx/32509/lib/node_modules/@asyncapi/generator/lib/generator.js:233:21)
    at async /~/.npm/_npx/32509/lib/node_modules/@asyncapi/generator/cli.js:150:9
```

#### Expected behavior
I would expect the parser to be able to handle recursive messages.

derberg:
**Description**

- handling circular references

for file like 
```
asyncapi: 2.0.0
info:
  title: My Circular API
  version: '1.0.0'
channels:
  recursive:
    subscribe:
      message:
        payload:
          $ref: '#/components/schemas/RecursiveSelf'
  external/file:
    publish:
      message:
        payload:
          $ref: './good/circular-refs-file-ref.yaml#/ExternalFile'
components:
  schemas:
    RecursiveSelf:
      type: object
      properties:
        children:
          type: array
          items:
            $ref: '#/components/schemas/RecursiveSelf'
        something:
          type: object
          properties:
            test: 
               $ref: '#/components/schemas/RecursiveAncestor'
    RecursiveAncestor:
      type: object
      properties:
        children:
          type: array
          items:
            $ref: '#/components/schemas/RecursiveSelf'
        something:
          type: string
```

`AsyncAPIDocument.components().schema('RecursiveAncestor').properties().children._json.items` returns `'$.components.schemas.RecursiveSelf'`

**Related issue(s)**
Resolves #83


nsnichols:
I tried these changes. It's not blowing up anymore so that's good. However, if you have a specification with non-recursive messages and recursive messages, even full structures with no recursion whatsoever don't get dereferenced.

Try parsing the following specs and observe the difference in results when recursive schemas are present vs when they are not.

```
asyncapi: 2.0.0
info:
  title: My Circular API
  version: '1.0.0'
channels:
  recursive:
    subscribe:
      message:
        payload:
          $ref: '#/components/schemas/RecursiveSelf'
  nonRecursive:
    subscribe:
      message:
        payload:
          $ref: '#/components/schemas/NonRecursive'
components:
  schemas:
    NonRecursive:
      type: object
      properties:
        child:
          $ref: '#/components/schemas/NonRecursiveChild'
    NonRecursiveChild:
      type: object
      properties:
        value:
          type: string
    RecursiveSelf:
      type: object
      properties:
        children:
          type: array
          items:
            $ref: '#/components/schemas/RecursiveSelf'
        something:
          type: object
          properties:
            test:
               $ref: '#/components/schemas/RecursiveAncestor'
    RecursiveAncestor:
      type: object
      properties:
        children:
          type: array
          items:
            $ref: '#/components/schemas/RecursiveSelf'
        something:
          type: string
```
```
asyncapi: 2.0.0
info:
  title: My Circular API
  version: '1.0.0'
channels:
  nonRecursive:
    subscribe:
      message:
        payload:
          $ref: '#/components/schemas/NonRecursive'
components:
  schemas:
    NonRecursive:
      type: object
      properties:
        child:
          $ref: '#/components/schemas/NonRecursiveChild'
    NonRecursiveChild:
      type: object
      properties:
        value:
          type: string
```
Note that there is no overlap between the recursive and non-recursive messages. I can understand why having recursive messages could cause their structures to change, but I wouldn't expect completely unrelated schema objects to also be affected.

Another question: Is `"key": "$.components.schema.SomeName"` really better than just leaving the ref alone (`"key": { "$ref": "#/components/schema/SomeName" } `). Having an object containing some property is consistent with how other field types are handled. Won't it be easier for the generator and templates if they could just look for a `$ref` property instead of having to check if it's either a `string` or an `object`?

derberg:
@nsnichols thanks a lot for the example. I'm starting another round here 😅  and `payload: '$.components.schemas.RecursiveSelf'` approach is not possible since I added https://github.com/asyncapi/parser-js/releases/tag/v0.24.0 and with such custom handling, where instead of object I put string the validation of the payload against AsyncAPI format obviously fails

test/parse_test.js
```diff
@@ -377,6 +379,19 @@ describe('parse()', function() {
...
+
+  it('should properly mark circular references', async function() {
+    const result = await parser.parse(inputYAMLCircular, { path: __filename });
+    expect(result.components().schema('RecursiveAncestor').properties().children._json.items[xParserCircle]).to.equal(true);
+    expect(result.components().schema('RecursiveSelf').properties().something._json.properties.test.properties.children.items[xParserCircle]).to.equal(true);
+    expect(result.channel('external/file').publish().json().message.payload.properties.test.properties.children.items[xParserCircle]).to.equal(true);
+    expect(result.json()[xParserCircle]).to.equal(true);
+    //not testing on a model level as required xParserCircle value is added before model construction so we need to test through calling parser function
+    expect(result.hasCircular()).to.equal(true);
+    expect(result.components().schema('RecursiveSelf').isCircular()).to.equal(true);
```
fmvilas, line 391: Ok so now I see where we have different understandings of the meaning of "circular". IMO, this should evaluate to `false` and instead `hasCircular()` should evaluate to `true`. Why?

1. This schema is not circular, it's recursive. It **contains** circular properties. More specifically, the `children` property **is** circular. Even more, the `items` schema **is** circular.
1. As a developer, I'd love to know when I'm in front of a recursive schema and when I'm in front of a circular property of that schema. As an example, in the HTML renderer, I want to render the schema as usual, with all its properties but once I get into the circular property, I don't want to continue rendering nested properties but instead, I want to just render a link to another part of the document (where the referenced schema is rendered). The same would be applied to the generation of types/structs/etc. If we don't do it like this, developers would have to check themselves if that's the first time they see this schema or not.

Following what I said, tests should look like:

```js
    expect(result.components().schema('RecursiveSelf').isCircular()).to.equal(false);
    expect(result.components().schema('RecursiveSelf').hasCircular()).to.equal(true);
    expect(result.components().schema('RecursiveSelf').properties()['children'].items().isCircular()).to.equal(true);
```

derberg:
```
    expect(result.components().schema('RecursiveSelf').isCircular()).to.equal(false);
    expect(result.components().schema('RecursiveSelf').hasCircular()).to.equal(true);
```

makes total sense, totally agree as what you wrote, it has, not is

derberg: @fmvilas do you think `expect(result.components().schema('RecursiveSelf').hasCircular()).to.equal(true);` is a must? like useful really. I can see a use case for `expect(result.hasCircular()).to.equal(true);` as the tool can quickly throw to the user that circular refs are not supported, but `expect(result.components().schema('RecursiveSelf').hasCircular()).to.equal(true);` doesn't help much as it means that some properties might be circular and some not and I anyway have to go throught them and react only in case of `isCircular()`

or you provided a use case above but I just didn't fully get it?

fmvilas: That's a good point. I can't see it being used anywhere. You can probably drop `hasCircular` from Schema model.

derberg: @fmvilas ok man, it's completed. Works like a charm.

I want to admit, that after looking at the implementation now I agree now that this `isCircular` function can be very helpful for tools developers!

lib/parser.js
```diff
@@ -148,6 +133,34 @@ function parseFromUrl(url, fetchOptions = {}, options) {
...
+  }
+}
+
+/*
+ * In case of circular refs, this function dereferences the spec again to dereference circular dependencies
+ * Last step is to discover those and mark properly with information whey they are pointing
+*/
+async function handleCircularRefs(parsedJSON, initialFormat, asyncapiYAMLorJSON, options) {
+  options.dereference = {circular: true};
+  await dereference(parsedJSON, initialFormat, asyncapiYAMLorJSON, options);
```
fmvilas, line 159: Not a big issue but `options.dereference` is modifying the input parameter. Let's try to avoid mutating parameters as much as possible:

```suggestion
  await dereference(parsedJSON, initialFormat, asyncapiYAMLorJSON, { ...options, circular: true });
```

lib/parser.js
```diff
@@ -42,6 +44,10 @@ async function parse(asyncapiYAMLorJSON, options = {}) {
   let initialFormat;
 
   options.path = options.path || `${process.cwd()}${path.sep}`;
+  //because of Ajv lacks support for circular refs, parser should not resolve them before Ajv validation and first needs to ignore them and leave circular $refs to successfully validate the document
+  //this is done pair to advice from Ajv creator https://github.com/ajv-validator/ajv/issues/1122#issuecomment-559378449
+  //later we perform full dereference of circular refs if they occure
+  options.dereference = { circular: 'ignore'};
```
fmvilas, line 50: I think it's safe to say now that we can remove `dereference` from the `options` object. I mean not allowing the user to customize its value because we need it ourselves and we're not using the passed value anyway.

lib/parser.js
```diff
@@ -74,39 +80,17 @@ async function parse(asyncapiYAMLorJSON, options = {}) {
...
-
-  const ajv = new Ajv({
-    jsonPointers: true,
-    allErrors: true,
-    schemaId: 'id',
-    logger: false,
-  });
-
-  ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-04.json'));
+    await dereference(parsedJSON, initialFormat, asyncapiYAMLorJSON, options);
```
fmvilas, line 83: Following my comment above (saying `dereference` shouldn't be an option anymore):

```suggestion
    await dereference(parsedJSON, initialFormat, asyncapiYAMLorJSON, { ...options, dereference: 'ignore' });
```

lib/models/asyncapi.js
```diff
@@ -263,55 +271,71 @@ function addNameToKey(messages, number) {
   });
 }
 
+/**
+ * Add anonymous name to key if no name provided.
```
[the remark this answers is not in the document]
derberg: how do you know, don't believe everything you see 😄

lib/models/asyncapi.js
```diff
@@ -263,55 +271,71 @@ function addNameToKey(messages, number) {
...
+ * 
+ * @param {Schema} schema schema that is currently accessed and need to be checked if it is a first time
+ * @param {Array} seenObjects list of object that were already seen during recursion 
+ */
+function isCircular(schema, seenObjects) {
+  const json = schema.json();
+
+  if (!seenObjects.includes(json)) return false;
+  return true;
```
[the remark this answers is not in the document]
derberg: oh, damn, this method change so many times that at the end I did not notice it can just be a one-liner 😄

lib/parser.js
derberg: good point, done

derberg:
@fmvilas nah, I want to close it before holidays and merge so people can try it out if they want and report bugs, so I can work on them once I'm back 😄 

anyway, all fixes provided, ready for final review

PR #104, opened 2020-07-03
feat: add payload validation against asyncapi schema format

### LINKED ISSUE #75: Parser inconsistently applying additionalProperties schema validation
#### Describe the bug

Depending on how the schema is constructed, additionalProperties is sometimes allowed to deviate from the schema without a parse error.

#### How to Reproduce

The following is parsed successfully:

```
asyncapi: 2.0.0
info:
  title: Demo API
  version: '2-0-0'

channels:
  /someChannel:
    publish:
      operationId: someOp
      message:
        payload:
          type: object
          additionalProperties: [invalid_array]
```

However, the following is rejected:

```
asyncapi: 2.0.0
info:
  title: Demo API
  version: '2-0-0'

channels:
  /someChannel:
    publish:
      operationId: someOp
      message:
        payload:
          $ref: #/components/schemas/someOp
components:
  schemas:
    someOp:
      type: object
      additionalProperties: [invalid_array]
```

#### Expected behavior
I would expect the parser to reject both formats, given it does not comply with the schema

derberg:
**Description**

- add payload validation against asyncapi schema format

**Related issue(s)**
Fixes #75


test/asyncapiSchemaFormatParser_test.js
```diff
@@ -0,0 +1,41 @@
...
+      await parser.parse(invalidAsyncapi);
+    } catch (e) {
+      console.log('papapapapa', e.validationErrors);
+      expect(e.type).to.equal('https://github.com/asyncapi/parser-js/validation-errors');
+      expect(e.title).to.equal('There were errors validating the message payload against the AsyncAPI schema format.');
+      expect(e.parsedJSON).to.deep.equal(JSON.parse(invalidAsyncapi));
+      expect(e.validationErrors).to.deep.equal([
+        {
+          title: '/additionalProperties should be object,boolean',
+          location: { jsonPointer: '/additionalProperties' }
```
derberg, line 20: I'm not happy with this information, it is limited to the message payload only, but how someone can know what payload, for which message in what channel. 

I need to improve it and also check if maybe I can also get info about the error location in the file (line/column)

lib/asyncapiSchemaFormatParser.js
```diff
@@ -0,0 +1,46 @@
...
+
+async function parse({ message, oryginalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument }) {
+  const ajv = new Ajv({
+    jsonPointers: true,
+    allErrors: true,
+    schemaId: 'id',
+    logger: false,
+  });
+  const payloadSchema = preparePayloadSchema(asyncapi[parsedAsyncAPIDocument.asyncapi]);
+  const validate = ajv.compile(payloadSchema);
```
fmvilas, line 19: It would be great to move these lines outside the `parse` function. Compiling the schema is a relatively slow process and just needs to be done once per the AsyncAPI version. If we leave it here, it's compiling the schema on every `parse` call.

Another solution would be to cache the `validate` variable so it's only compiled once. For instance:

```js
let compiledSchemasCache = new Map();

async function parse({ message, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument }) {
  let validate;
  if (compiledSchemasCache.get(parsedAsyncAPIDocument.asyncapi)) {
    validate = compiledSchemasCache.get(parsedAsyncAPIDocument.asyncapi);
  } else {
    const ajv = new Ajv({
      jsonPointers: true,
      allErrors: true,
      schemaId: 'id',
      logger: false,
    });
    const payloadSchema = preparePayloadSchema(asyncapi[parsedAsyncAPIDocument.asyncapi]);
    validate = ajv.compile(payloadSchema);
    compiledSchemasCache.set(parsedAsyncAPIDocument.asyncapi, validate);
  }

  const valid = validate(message.payload);
  ...
}
```

lib/asyncapiSchemaFormatParser.js
```diff
@@ -0,0 +1,46 @@
...
+  const valid = validate(message.payload);
+
+  if (!valid) throw new ParserError({
+    type: 'validation-errors',
```
fmvilas, line 23: I'd be more specific here. Something like:

```suggestion
    type: 'schema-validation-errors',
```

lib/asyncapiSchemaFormatParser.js
```diff
@@ -0,0 +1,46 @@
...
+    title: 'There were errors validating the message payload against the AsyncAPI schema format.',
```
fmvilas, line 24: What about:

```suggestion
    title: 'This is not a valid AsyncAPI Schema Object.',
```

This way we use the same vocabulary as the spec ([Schema Object](https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#schemaObject)) and make it clear these are user errors and not other kinds of errors like the parsed had a problem parsing it. In short, it's user's fault and they should fix it.

lib/parser.js
```diff
@@ -165,20 +165,23 @@ async function customDocumentOperations(js, asyncapiYAMLorJSON, initialFormat, o
...
-        if (msg) await validateAndConvertMessage(msg);
+        if (msg) await validateAndConvertMessage(msg, asyncapiYAMLorJSON, initialFormat, js);
       }
     });
     await Promise.all(convert);
   }
 }
 
-async function validateAndConvertMessage(msg) {
+async function validateAndConvertMessage(msg, oryginalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument) {
```
fmvilas, line 175:
```suggestion
async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument) {
```

lib/parser.js
```diff
@@ -165,20 +165,23 @@ async function customDocumentOperations(js, asyncapiYAMLorJSON, initialFormat, o
...
   const schemaFormat = msg.schemaFormat || DEFAULT_SCHEMA_FORMAT;
 
   await PARSERS[schemaFormat]({
     schemaFormat,
     message: msg,
     defaultSchemaFormat: DEFAULT_SCHEMA_FORMAT,
+    oryginalAsyncAPIDocument,
```
fmvilas, line 182:
```suggestion
    originalAsyncAPIDocument,
```

test/asyncapiSchemaFormatParser_test.js
```diff
@@ -0,0 +1,41 @@
...
+const path = require('path');
+const expect = chai.expect;
+
+describe('asyncapiSchemaFormatParser', function() {
+  it('should throw an error because of invalid schema', async function() {
+    const invalidAsyncapi = fs.readFileSync(path.resolve(__dirname, './wrong/invalid-payload-asyncapi-format.json'), 'utf8');
+    try {
+      await parser.parse(invalidAsyncapi);
+    } catch (e) {
+      console.log('papapapapa', e.validationErrors);
```
fmvilas, line 13: https://www.youtube.com/watch?v=RtvqVbg7lII

test/asyncapiSchemaFormatParser_test.js
fmvilas: Agree. It would be great to be able to have some context.

lib/asyncapiSchemaFormatParser.js
derberg: @fmvilas I do not think we have to worry as according to readme, ajv already caches it for us:
> Ajv compiles schemas to functions and caches them in all cases (using schema serialized with fast-json-stable-stringify or a custom function as a key), so that the next time the same schema is used (not necessarily the same object instance) it won't be compiled again.

derberg:
@fmvilas ready for another round of review. I applied your suggestions and also added much smarter errors that point where the error is in the AsyncAPI document and not just the payload

lib/asyncapiSchemaFormatParser.js
fmvilas: Oh! I didn't know they added this. That's great then. Forget about it.

lib/asyncapiSchemaFormatParser.js
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * To validate schema of the payload we just need a small portion of official AsyncAPI spec JSON Schema, the definition of the schema must be
+ * a main part of the JSON Schema
+ * 
+ * @param  {Object} asyncapiSchema AsyncAPI specification JSON Schema
+ * @returns {Object} valid JSON Schema document describing format of AsyncAPI-valid schema for message payload
+ */
+function preparePayloadSchema(asyncapiSchema) {
+  return {
+    $ref: '#/definitions/schema',
```
fmvilas, line 50: That's smart 👏

lib/asyncapiSchemaFormatParser.js
```diff
@@ -0,0 +1,69 @@
...
+ * 
+ * @param  {Array<Object>} errors Ajv errors
+ * @param  {String} path Path to location of the payload schema in AsyncAPI Document
+ * @returns {Array<Object>} same object as received in input but with modified datePath property so it contain full path relative to AsyncAPI document
+ */
+function addFullPathToDataPath(errors, path) {
+  errors.forEach((val) => {
+    val.dataPath = path + val.dataPath;
+  });
+  return errors;
```
fmvilas, line 67: I have a preference for not mutating the input parameters when possible. I know it's not a problem in this case but consider using `map` instead of `forEach`:

```js
return errors.map((err) => ({
  ...err,
  ...{
    dataPath: `${path}${err.dataPath}`
  }
}));
```

derberg: sure thing changed, I do not use spread operators quite often so do not consider them much during dev, but they make a lot of sense

derberg:
@fmvilas what would it be useful for? I added full path as it is super useful for editors, to highlight the exact line that has error. What would `/additionalProperties` path be useful for? for someone that runs a parser it is a kind of out of context

fmvilas:
Can't think of a use case right now but thought it's cheap to add and maybe someone would find it useful.

derberg:
@fmvilas adding is cheap, but as you said, no use case, and tbh I think at the moment that this information is completely useless, and that is why I took the effort to provide a full path

PR #105, opened 2020-07-03
feat: validation checking if operationIds are duplicated

### LINKED ISSUE #10: List down manual validations
The JSON Schema version of the spec is capable of validating many things, however, many others are impossible or very complicated to express. We should carefully read the spec and create a list of all the restrictions/validations that should be applied aside from the JSON Schema ones.

## List

- [~~Message examples are valid messages. Not sure if ajv is performing this validation on the schema `examples`~~](https://github.com/asyncapi/parser-js/issues/99)
- [~~Server variable examples are valid according to the schema.~~](https://github.com/asyncapi/parser-js/issues/100)
- Operation Ids are unique.
- Check that a server security scheme name matches an object on components/securitySchemes.

derberg:
**Description**

- validation checking if operationIds are duplicated

**Related issue(s)**
Resolves #10


lib/customValidators.js
```diff
@@ -43,25 +62,72 @@ function validateChannelParams(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
...
+    
+    //isOperationIdDuplicated always holds one record and it is an array of paths, the one that is a duplicate and the one that is duplicated 
+    duplicatedOperations.set(operationPath, isOperationIdDuplicated[0][1]); 
+  };
+
+  chnlsMap.forEach((val,key) => {
+    const sub = val.subscribe;
+    const pub = val.publish;
+
+    if (sub) addDuplicateToMap(sub, key, 'subscribe');
```
derberg, line 121: I should probably use OPERATIONS array from parser.js?

lib/customValidators.js
```diff
@@ -43,25 +62,72 @@ function validateChannelParams(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
...
+  if (!chnls) return true;
+  const chnlsMap = new Map(Object.entries(chnls));
+  //it is a map of paths, the one that is a duplicate and the one that is duplicated
+  const duplicatedOperations = new Map();
+  //is is a 2-dimentional array that holds information with operationId value and its path
+  const allOperations = [];
+  
+  const addDuplicateToMap = (op, k, opName) => {
+    const operationId = op.operationId;
+    const operationPath = `${ k.replace(regex, '~1') }/${ opName }/operationId`;
```
derberg, line 109: I'll check if I can use existing tilde/untilde functions

PR #106, opened 2020-07-06
fix: introduce security related linter and fix warnings

### LINKED ISSUE #103: Introduce security related linter and fix existing problems
#### Reason/Context
There are some common security issues that can be found statically and we should make sure parser is 100% free of those.

#### Description
- add https://github.com/nodesecurity/eslint-plugin-security and configure with eslint
- fix all the issues reported by the linter. There are many but trivial where mainly it is about switching form object to map. Good issue to familiarize with the project structure

magicmatatjahu:
**Description**
- Add `eslint-plugin-security` package and configure it in project,
- Bump `eslint` package to `7.X.X`,
- Introduce `getMapKey` util function for retrieve data by key from object - use it in models,
- Fix warnings returned by `eslint-plugin-security`.

**Related issue(s)**
Resolves #103


lib/models/operation.js
```diff
@@ -30,10 +30,10 @@ class Operation extends OperationTraitable {
    * @returns {Message}
    */
   message(index) {
-    if (!this._json.message) return null;
+    if (typeof index !== 'number' || !this._json.message) return null;
     if (!this._json.message.oneOf) return new Message(this._json.message);
     if (index > this._json.message.oneOf.length - 1) return null;
-    return new Message(this._json.message.oneOf[index]);
+    return new Message(this._json.message.oneOf[+index]);
```
derberg, line 36: how is this change related to security warnings?

lib/models/schema.js
```diff
@@ -240,6 +230,7 @@ class Schema extends Base {
    * @returns {Object<string, Schema|string[]>}
    */
   dependencies() {
+    /* eslint-disable security/detect-object-injection */
```
derberg, line 233: why disable, we we cannot just use Map here. Which part linter doesn't like?

lib/parser.js
```diff
@@ -154,6 +154,7 @@ async function customDocumentOperations(js, asyncapiYAMLorJSON, initialFormat, o
   
   validateChannelParams(js, asyncapiYAMLorJSON, initialFormat);
 
+  /* eslint-disable security/detect-object-injection */
```
derberg, line 157: `js.channels` can be converted into Map -> https://github.com/asyncapi/parser-js/pull/105/files#diff-c3d5db07c30dd8c16e1da041b6f90850R115

of course depends what part is a problem for the linter

lib/utils.js
```diff
@@ -157,31 +159,38 @@ utils.createMapOfType = (obj, Type) => {
...
+};
+
+utils.getMapKey = (obj, key) => {
+  return utils.getMapKeyOfType(obj, key);
 };
 
 utils.addExtensions = (obj) => {
   obj.prototype.extensions = function () {
     const result = {};
     Object.keys(this._json).forEach(key => {
```
derberg, line 183: same as with others, we could make `this._json` a Map

lib/utils.js
```diff
@@ -157,31 +159,38 @@ utils.createMapOfType = (obj, Type) => {
...
 
   Object.keys(obj).forEach(key => {
+    // eslint-disable-next-line security/detect-object-injection
     result[key] = new Type(obj[key]);
   });
 
   return result;
 };
 
 utils.getMapKeyOfType = (obj, key, Type) => {
```
derberg, line 169: I have some naming issues here, now this function is basically not only for getting a typed map, so name is no longer `getMapKeyOfType` and hiding it with `getMapKey` seems strange too me 🤔 
why not try renaming `getMapKeyOfType` to something more generic? with good jsdoc?

lib/models/operation.js
magicmatatjahu: `eslint-plugin-security` warns that index is custom variable passed by argument, and it could be string etc. `+` before variable, casts it to number: `"1" => 1`. If you pass something like `"FOO"`, then it will be `NaN`, so it removes malicious code written as string in variable.

lib/models/schema.js
magicmatatjahu: Linter doesn't like the `key` variable. If you use `Object.keys()` then you have 100% sure, than variable is string and we don't do things like `_json.dependencies[key]()`, so we avoid eval malicious code, so I decided to disable plugin in those lines. I could use utils helper `createMapOfType`, but as you can see, in one case we create Schema Type, in another we write raw value.

lib/parser.js
magicmatatjahu: The problem is with retrieving value by `channelName` and `opName` name. As you see, this variable will be in 100% strings.

lib/utils.js
magicmatatjahu: But then users will have to get data via the `get ()` method and not via the array index, and for me this is a breaking change. If we decided to use Map here, then we should also adjust whole methods, which return plain JS Object, to Map. Or probably I don't understand your suggestion :D

lib/utils.js
magicmatatjahu: I will try make it more generic :)

lib/parser.js
magicmatatjahu: And as I remember `forEach` method from ES6 Map cannot handle async functions, and in this function we must handle async conversion of Messages.

lib/utils.js
magicmatatjahu: Also in `162` line, without `eslint-disable`, we should cast key to String like `result[String(key)] = new Type(obj[String(key)]);`

lib/models/schema.js
derberg: I don't like to disable rules unless it is a test or there is really nothing else we can do.
values are accessed by a key and with Map it is not the case so it is better, so I don't really get your explanation about `_json.dependencies[key]()`

lib/parser.js
derberg: you are right about `forEach` but then you can just use `for of`

thing is that we should not assume that what we have now (100% string) will always be like that if the future, so better just to fix than ignore

lib/utils.js
derberg: yeap, do the casting

did you make changes here? I don't think so

lib/utils.js
derberg: you change it to a map only for `forEach` but in 193 still return pure json

PR #108, opened 2020-07-06
fix: apply correctly traits to channels

### LINKED ISSUE #96: Message traits are not applied when using oneOf on channels
#### Describe the bug
Message traits are not applied when using `oneOf` on channels.

#### How to Reproduce
Using the following code for testing:

```js
const asyncapi = await parser.parse(<FILE>);
  
const result = asyncapi.channels()['test'].subscribe().messages()[0].contentType();
```

I get the expected result (`contentType: application/json`) with the following YAML file:

```yaml
asyncapi: 2.0.0

info:
  title: Test
  version: 1.0.0

channels:
  test:
    subscribe:
      message:
        $ref: "#/components/messages/test"

components:
  messages:
    test:
      payload:
        type: string
      traits:
        - $ref: "#/components/messageTraits/commons"

  messageTraits:
    commons:
      contentType: application/json
```

But I get the wrong result (`contentType: undefined`) with the following YAML file:

```yaml
asyncapi: 2.0.0

info:
  title: Test
  version: 1.0.0

channels:
  test:
    subscribe:
      message:
        oneOf:
          - $ref: "#/components/messages/test"

components:
  messages:
    test:
      payload:
        type: string
      traits:
        - $ref: "#/components/messageTraits/commons"

  messageTraits:
    commons:
      contentType: application/json
```

I have tried with more than one message, and other traits, like headers, but they are missing from the message too.

#### Expected behavior
Message traits should be applied.

magicmatatjahu:
**Description**
- Apply correctly traits to channels, also for `oneOf` messages.

**Related issue(s)**
Resolves #96


test/parse_test.js
```diff
@@ -51,11 +50,6 @@ describe('parse()', function() {
     const result = await parser.parse(inputYAML, { path: __filename });
     await expect(JSON.stringify(result.json())).to.equal(outputJSON);
   });
-
-  it('should not apply traits', async function() {
-    const result = await parser.parse(inputYAML, { path: __filename, applyTraits: false });
-    await expect(JSON.stringify(result.json())).to.equal(outputJsonNoApplyTraits);
-  });
```
jonaslagoni, line 58: Is there a reason this test is removed?

lib/parser.js
```diff
@@ -159,13 +159,21 @@ async function customDocumentOperations(js, asyncapiYAMLorJSON, initialFormat, o
     const convert = OPERATIONS.map(async (opName) => {
       const op = channel[opName];
       if (op) {
+        let messages = [];
+        const msg = op.message;
+        if (msg) {
+          if (msg.oneOf) {
```
jonaslagoni, line 165: Should we just leave the other options (`allOf` and `anyOf`) for later issues? And are there even use-cases for them? 🤔

lib/parser.js
```diff
@@ -159,13 +159,21 @@ async function customDocumentOperations(js, asyncapiYAMLorJSON, initialFormat, o
...
+
         if (options.applyTraits) {  
           applyTraits(op);
-          applyTraits(op.message);
+          messages.forEach(m => applyTraits(m));
         }
-        
-        const msg = op.message;
-        if (msg) await validateAndConvertMessage(msg);
+        messages.forEach(m => validateAndConvertMessage(m));
```
jonaslagoni, line 176: Since `validateAndConvertMessage` is async `forEach` is not possible to use here is it? 🤔

test/parse_test.js
magicmatatjahu: It's not removed, its moved down with new test case.

lib/parser.js
magicmatatjahu: For messages in Operation Type, AsyncAPI have two only two options: single message or `oneOf` messages. Pic from documentation 

image

`allOf` and `anyOf` is in Schema Type :)

lib/parser.js
magicmatatjahu: I missed it. I will correct it. Thanks for review! :)

test/parse_test.js
jonaslagoni: Ahh my bad, missed that yesterday 🧐

derberg:
lgtm 
thanks @magicmatatjahu @jonaslagoni

PR #109, opened 2020-07-06
refactor: use .createMapOfType() util method in schema.model

### LINKED ISSUE #107: Fix 3 bugs reported by sonarcube
#### Describe the bug
SonarCube reported 3 trivial bugs that we should solve before 1.0 release

#### Expected behavior
Fix https://sonarcloud.io/project/issues?id=asyncapi_parser-js&resolved=false&types=BUG

magicmatatjahu:
**Description**
- use `.createMapOfType()` util method in Schema model.
- refactor `.findLocationOf()` method.

**Related issue(s)**
Resolves #107

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -55,10 +55,10 @@ const findNodeInAST = (ast, location) => {
 };
 
 const findLocationOf = (keys, ast, initialFormat) => {
-  let node;
-  let info;
   if (initialFormat === 'js') return { jsonPointer: `/${keys.join('/')}` };
   
+  let node;
+  let info;
   if (initialFormat === 'yaml') {
     node = findNode(ast, keys);
     if (!node) return { jsonPointer: `/${keys.join('/')}` };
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -69,6 +69,10 @@ const findLocationOf = (keys, ast, initialFormat) => {
     info = node.loc;
   }
 
+  if (!info || !info.start) {
+    throw new Error('Invalid `info` object.');
```
derberg, line 73: please throw ParserError here. All the types are mentioned here https://github.com/asyncapi/parser-js/#error-types

although I'm not sure error is a good solution here. If info is missing we are just not able to return in the error information about the error location, but jsonPointer can be returned. And if you check `start` then why not `end`?

why did you choose throwing an Error?

magicmatatjahu: Firstly I want use ParserError and I didn't know which type choose.

You're right, return `jsonPointer` is a better idea than throw error, sorry for that. In this solution I only check start, because then in code `info.end` is checked in ternary operators.

I will adjust method to return jsonPointer.

derberg: @magicmatatjahu lgtm, great!

PR #120, opened 2020-07-14
refactor: improve code basing on linters warnings

derberg:
**Description**

- make eslint throw error also on warnings. If we don't do it, many security issues or the sonarjs things will be left and we will have to from time to time fix them which makes no sense
- fix issues from https://sonarcloud.io/dashboard?id=asyncapi_parser-js
- fix sonarjs cognitive complexity warnings


lib/json-parse.js
```diff
@@ -1,35 +1,48 @@
+const RE2 = require('re2');
+
 module.exports = (txt, reviver, context = 20) => {
+  const regexToken = new RE2(/^Unexpected token.*position\s+(\d+)/i);
+  const regexEndof = new RE2(/^Unexpected end of JSON.*/i);
```
magicmatatjahu, line 5: Those regex can be extracted outside the function. It's unnecessary to re-compiled those const in every call of this function.

lib/json-parse.js
```diff
@@ -1,35 +1,48 @@
...
     e.offset = errIdx;
     const lines = txt.substr(0, errIdx).split('\n');
     e.startLine = lines.length;
     e.startColumn = lines[lines.length - 1].length;
     throw e;
   }
 };
+
+function handleJsonNotString(txt) {
+  if (typeof txt !== 'string') {
```
magicmatatjahu, line 24: Here, is possible to change the conditional order and make something like:

```js
if (typeof txt === 'string') {
  return;
}

const isEmptyArray = Array.isArray(txt) && txt.length === 0;
const errorMessage = `Cannot parse ${ 
      isEmptyArray ? 'an empty array' : String(txt)}`;
throw new TypeError(errorMessage);
```

But this is only a suggestion.

lib/json-parse.js
```diff
@@ -1,35 +1,48 @@
...
+    const isEmptyArray = Array.isArray(txt) && txt.length === 0;
+    const errorMessage = `Cannot parse ${ 
+      isEmptyArray ? 'an empty array' : String(txt)}`;
+    throw new TypeError(errorMessage);
+  }
+}
+
+function handleErrIdxNotNull(e, txt, errIdx, context) {
+  if (errIdx !== null) {
```
magicmatatjahu, line 33: Identically as above comment.

lib/models/asyncapi.js
```diff
@@ -369,4 +347,50 @@ function assignIdToAnonymousSchemas(doc) {
...
+/**
+ * Recursively go through schema of object type and execute callback.
+ * 
+ * @param {Schema} schema Object type.
+ * @param {Function} callback(schema)
+ *         the function that is called foreach schema found.
+ *         schema {Schema}: the found schema.
+ */
+function recursiveSchemaObject(schema, callback) {
+  if (schema.additionalProperties() !== undefined && typeof schema.additionalProperties() !== 'boolean') {
```
magicmatatjahu, line 359: `schema.additionalProperties()` could be write to const. I think call 3 times in a row `schema.additionalProperties()` function is unnecessary.

```js
const additionalProperties = schema.additionalProperties();
if (additionalProperties !== undefined && typeof additionalProperties !== 'boolean') {
    recursiveSchema(additionalProperties, callback);
}

// rest of logic identically.
```

lib/models/asyncapi.js
```diff
@@ -369,4 +347,50 @@ function assignIdToAnonymousSchemas(doc) {
...
+
+/**
+ * Recursively go through schema of array type and execute callback.
+ * 
+ * @param {Schema} schema Array type.
+ * @param {Function} callback(schema)
+ *         the function that is called foreach schema found.
+ *         schema {Schema}: the found schema.
+ */
+function recursiveSchemaArray(schema, callback) {
```
magicmatatjahu, line 379: Identically as above comment about save common value (here is a `schema.additionalItems()` and `schema.items()`) in const.

lib/parser.js
```diff
@@ -138,7 +138,12 @@ async function parse(asyncapiYAMLorJSON, options = {}) {
  * @param {Object} [options] Configuration to pass to the {@link module:Parser#parse} method.
  * @returns {Promise<AsyncAPIDocument>} The parsed AsyncAPI document.
  */
-function parseFromUrl(url, fetchOptions = {}, options) {
+function parseFromUrl(url, fetchOptions, options) {
+  //Why not just addinga default to the arguments list?
+  //All function parameters with default values should be declared after the function parameters without default values. Otherwise, it makes it impossible for callers to take advantage of defaults; they must re-specify the defaulted values or pass undefined in order to "get to" the non-default parameters.
+  //To not breat the API by changing argument position and to silet the linter it is just better to move adding
+  if (!fetchOptions) fetchOptions = {};
```
magicmatatjahu, line 145: `if (fetchOptions === undefined) fetchOptions = {};`

I know that null, or false is also falsey, but I think, that this code should support native defaulting operation in argument (in this only works with `undefined`)

lib/json-parse.js
derberg: I did not change the logic when not needed on purpose

lib/json-parse.js
derberg: I did not change the logic when not needed on purpose

lib/models/asyncapi.js
derberg: I did not change the logic when not needed on purpose, I just wanted to get rid of cognitive complexity

lib/models/asyncapi.js
derberg: I did not change the logic when not needed on purpose, I just wanted to get rid of cognitive complexity

lib/parser.js
derberg: why? I saw people approach passing arguments with `null` in case they do not need to pass anything

magicmatatjahu: Yes, but only `undefined` works with default value of argument, as you wrote in comment to this line. If it allows false and null, why not just allow undefined?

derberg: hmmm, not sure I follow, the comment from my side is about removing default value from the argument and handling it inside function. With your approach, if someone passes null, it won't work:

given:
```
function lysy(m) {
  if (!m) m = "lysy jest fajny"
  console.log(m)
}

function lysyUndefined(m) {
  if (m === undefined) m = "lysy jest fajny"
  console.log(m)
}

lysy("lysy jest niefajny")
lysy()
lysy(undefined)
lysy(null)

lysyUndefined("lysy jest niefajny")
lysyUndefined()
lysyUndefined(undefined)
lysyUndefined(null)
```

results in
```
"lysy jest niefajny"

"lysy jest fajny"

"lysy jest fajny"

"lysy jest fajny"

"lysy jest niefajny"

"lysy jest fajny"

"lysy jest fajny"

null
```

I do not want to be `null` 😆

magicmatatjahu: I wrote about it :D native defaulting works only with undefined, but if you want support also `null` or `false`, ok :)

PR #127, opened 2020-07-23
Docs: Update README.md with requirements

jonaslagoni:
**Description**
Since we have a github link as a dependency we must define Git as a requirement.

**Related issue(s)**
This originates from https://github.com/asyncapi/generator/issues/389

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,11 @@ Use this package to parse and validate AsyncAPI documents —either YAML or JSON
 
 > This package doesn't support AsyncAPI 1.x.
 
+## Requirements
+
+* Git v1+
+
+
 ### Install
 
 ```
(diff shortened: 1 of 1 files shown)


jonaslagoni:
> Not sure we should add anything here. The dependency is temporary and will soon be changed by a regular npm package. What about adding it if someone faces this issue?

Hmm, I mean now that we know its a requirement why waste someone's time debugging when we have the answer. I know it's temporary, but I also know temporary things have a way of staying temporary for quite some time. And if we can save a person an hour debugging etc with a temporary requirement in the README I feel like its worth it 😄

fmvilas:
Fair point

derberg:
I added a comment on this one here https://github.com/asyncapi/generator/issues/389#issuecomment-664475557 that we should not update the readme imho

derberg:
@jonaslagoni closing since we are not using fork anymore https://github.com/asyncapi/parser-js/pull/134

PR #132, opened 2020-07-29
fix: message function returns message and not null if index not provided

derberg:
**Description**

- message function returns message and not null if index not provided

**Related issue(s)**
See also https://github.com/asyncapi/java-spring-template/issues/75 where the issue was noticed

### DIFF
--- a/lib/models/operation.js
+++ b/lib/models/operation.js
@@ -31,8 +31,9 @@ class Operation extends OperationTraitable {
    * @returns {Message}
    */
   message(index) {
-    if (typeof index !== 'number' || !this._json.message) return null;
+    if (!this._json.message) return null;
     if (!this._json.message.oneOf) return new Message(this._json.message);
+    if (typeof index !== 'number') return null;
     if (index > this._json.message.oneOf.length - 1) return null;
     return new Message(this._json.message.oneOf[+index]);
   }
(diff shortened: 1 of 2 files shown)


lib/models/operation.js
```diff
@@ -31,8 +31,9 @@ class Operation extends OperationTraitable {
    * @returns {Message}
    */
   message(index) {
-    if (typeof index !== 'number' || !this._json.message) return null;
+    if (!this._json.message) return null;
     if (!this._json.message.oneOf) return new Message(this._json.message);
+    if (typeof index !== 'number') return null;
```
fmvilas, line 36: I don't get these changes. What's different?

derberg: `if (typeof index !== 'number') return null;` is a security requirement to always make sure provided index is a number

but when it was like this `if (typeof index !== 'number' || !this._json.message) return null;`, even if you did not provide index (this was possible with `message` function) you were always getting `null` becouse `undefined` is also not a `number`

so now I check for the number right before the part of the message where index is used and most important is that this part `    if (!this._json.message.oneOf) return new Message(this._json.message);` now works as before

fmvilas: Oh, I see! Thanks for the explanation 🙏

PR #134, opened 2020-07-31
feat: use latest json references parser instead of the fork

derberg:
**Description**

- use latest json references parser instead of the fork
- because during this work I discovered issues with tests that are green even though they should not, I decided to do the full cleanup in `parse_test.js` and not just a single test. I did not refactor tests from `asyncapiSchemaFormatParser_test.js` and `customValidators_test.js` as they are totally not related but must be refactored at some point to use `checkErrorWrapper`. I think this can be a followup task, but no strong opinion here so if you think I should do it now, just lemme know


lib/parser.js
```diff
@@ -134,23 +134,24 @@ function parseFromUrl(url, fetchOptions = {}, options) {
 async function dereference(parsedJSON, initialFormat, asyncapiYAMLorJSON, options) {
   try {
     return await refParser.dereference(options.path, parsedJSON, {
+      continueOnError: true,
```
derberg, line 137: new flag from [release v9](https://github.com/APIDevTools/json-schema-ref-parser/blob/master/CHANGELOG.md#v900-2020-04-21) that allows you to get all errors rather than just the first one

lib/parser.js
```diff
@@ -134,23 +134,24 @@ function parseFromUrl(url, fetchOptions = {}, options) {
...
       parse: options.parse,
       resolve: options.resolve,
       dereference: options.dereference,
     });
   } catch (err) {
     throw new ParserError({
       type: 'dereference-error',
-      title: err.message,
+      title: err.errors[0].message,
```
derberg, line 145: looking at the current structure of the ParserError I just pick up the first message from the list of errors. We might consider a change here in case we see a need in the future. I can create a follow-up issue if you think it is needed, but I'm affraid it will get stale pretty quickly and think it is better to leave it as it is and go back to it once we really need more details here for some UX purposes

lib/parser.js
```diff
@@ -134,23 +134,24 @@ function parseFromUrl(url, fetchOptions = {}, options) {
...
       parsedJSON,
-      refs: findRefs(parsedJSON, err.originalPath || err.path, options.path, initialFormat, asyncapiYAMLorJSON),
+      refs: findRefs(err.errors, initialFormat, asyncapiYAMLorJSON),
```
derberg, line 147: no need to pass `parsedJSON` as we no longer need to traverse through it to find a location where the error happened

lib/parser.js
```diff
@@ -134,23 +134,24 @@ function parseFromUrl(url, fetchOptions = {}, options) {
...
     });
   }
 }
 
 /*
  * In case of circular refs, this function dereferences the spec again to dereference circular dependencies
- * Last step is to discover those and mark properly with information whey they are pointing
+ * Special property is added to the document that indicates it contains circular refs
```
derberg, line 154: not related to this PR, just noticed this description was old and not matching reality

lib/utils.js
```diff
@@ -76,15 +76,6 @@ const findLocationOf = (keys, ast, initialFormat) => {
   };
 };
 
-const traverse = function (o, fn, scope = []) {
```
derberg, line 79: no need to traverse anymore, json ref parser now gives the information about where the error happened

test/parse_test.js
```diff
@@ -18,28 +18,30 @@ const invalidAsyncpiJSON = fs.readFileSync(path.resolve(__dirname, './wrong/inva
...
-const invalidAsyncAPI = { asyncapi: '2.0.0', info: {} };
+const outputJsonWithRefs = '{"asyncapi":"2.0.0","info":{"title":"My API","version":"1.0.0"},"channels":{"mychannel":{"publish":{"traits":[{"externalDocs":{"url":"https://company.com/docs"}}],"externalDocs":{"x-extension":true,"url":"https://irrelevant.com"},"message":{"traits":[{"x-some-extension":"some extension"}],"payload":{"type":"object","properties":{"name":{"type":"string"},"test":null}}}}},"oneOfMessageChannel":{"publish":{"message":{"oneOf":[{"traits":[{"x-some-extension":"some extension"}],"payload":{"type":"object","properties":{"name":{"type":"string"},"test":null}}}]}}}},"components":{"messages":{"testMessage":{"traits":[{"x-some-extension":"some extension"}],"payload":{"type":"object","properties":{"name":{"type":"string"},"test":null}}}},"schemas":{"testSchema":{"type":"object","properties":{"name":{"type":"string"},"test":null}}},"messageTraits":{"extension":{"x-some-extension":"some extension"}},"operationTraits":{"docs":{"externalDocs":{"url":"https://company.com/docs"}}}}}';
```
derberg, line 21: notice an interesting change in the way ref parser works. In the past it was failing with one broken ref and did not dereference the rest. Now all valid refs are dereferenced and the one that is invalid is replaced with `null`

test/parse_test.js
```diff
@@ -18,28 +18,30 @@ const invalidAsyncpiJSON = fs.readFileSync(path.resolve(__dirname, './wrong/inva
...
-  try {
-    await fn();
-    throw Error('should not be reachable');
-  } catch (e) {
-    expect(e instanceof ParserError).to.equal(true);
-    expect(e).to.have.own.property('type', type);
-    expect(e).to.have.own.property('message', message);
-  }
-};
+const checkErrorWrapper = async (fn, validationObject) => {
```
derberg, line 26: I have no idea how to reduce cognitive complexity of this function, I would ignore here this linter rule

derberg: `checkErrorTypeAndMessage` is now `checkErrorWrapper` as it is universal now and tests many different combinations

test/parse_test.js
```diff
@@ -18,28 +18,30 @@ const invalidAsyncpiJSON = fs.readFileSync(path.resolve(__dirname, './wrong/inva
...
   try {
     await fn();
-    throw Error('should not be reachable');
+    throw Error('This error should not be reachable. If you reached it, it means the function did not throw a proper error and executed successfully.');
   } catch (e) {
-    expect(JSON.stringify(e.parsedJSON)).to.equal(parsedJSON);
+    const isProperError = e instanceof ParserError;
+    if (!isProperError) console.log(e);
+
+    if (isProperError) expect(e instanceof ParserError).to.equal(true);
```
derberg, line 36: I added this part to make sure that when something really goes wrong and the error from line 31 is reached a clear and meaningfull error is actually visible to the user. Now it wasn't

test/parse_test.js
```diff
@@ -52,331 +54,432 @@ describe('parse()', function() {
...
+      ]
+    };
+    await checkErrorWrapper(async () => {
+      await parser.parse(inputYAML);
+    }, expectedErrorObject);
   });
 
   it('should offer information about YAML line and column where $ref errors are located', async function() {
-    try {
-      await parser.parse(inputYAML, { path: __filename });
```
derberg, line 204: `{ path: __filename }` here and in other places dragged my attention when I noticed something is wrong with tests. Path was provided so the parser could not fail, the test was green even though in the `catch` clouse there were checks that should fail.

PR #150, opened 2020-08-06
refactor: Makes checkErrorWrapper a standalone util function

### LINKED ISSUE #140: Extract checkErrorWrapper and use it in other tests
#### Reason/Context
Tests that do `try/catch` to evaluate thrown error must be wrapped in `checkErrorWrapper` to make sure tests always work, and they fail even if the test doesn't jump into the `catch`.

#### Description
- move `checkErrorWrapper` from `parse_test.js` to separate file, `utils.js` ?
- refactor `asyncapiSchemaFormatParser_test.js` and `customValidators_test.js` to use wrapper in test cases where `try/catch` is used

McZenith:
**Description**

- moved checkErrorWrapper to utils
- Refactor other tests using async/await to use the util function

**Related issue(s)**
Resolves #140


test/parse_test.js
```diff
@@ -5,7 +5,7 @@ const fs = require('fs');
 const path = require('path');
 const parser = require('../lib');
 const ParserError = require('../lib/errors/parser-error');
```
derberg, line 7: you can remove this line, not needed anymore

test/parse_test.js
```diff
@@ -31,28 +31,6 @@ const eolLength = EOL.length;
  * @param  {Function} fn Function that you want to test
  * @param  {Object} validationObject Error object to evaluate against the error thrown by fn()
 */
```
derberg, line 33: please move above jsdoc to utils to checkErrorWrapper function too

test/utils.js
```diff
@@ -0,0 +1,30 @@
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+const ParserError = require('../lib/errors/parser-error');
+chai.use(chaiAsPromised);
+const expect = chai.expect;
+
+
```
derberg, line 7: please remove this extra line, I suggest running locally `npm run lint` to see all the linter errors

derberg:
Also please remove yarn.lock from this PR

McZenith:
Thank you @derberg, I am actually trying to make daily incremental progress towards the issue. I will be pushing daily progress until I am able to finally resolve the issue. Thanks for the guides they are really helpful!

derberg:
@McZenith sounds good, then please convert it to a draft PR so no one bothers you and then submit it once ready

Screenshot 2020-08-06 at 17 27 23

McZenith:
Will do that immediately, thank you.

PR #151, opened 2020-08-06
feat: Validate oauth2 or openIdConnect have scopes in server sec

derberg:
**Description**

- Validate oauth2 or openIdConnect have scopes in server sec
- This PR will make these test cases to pass https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Security%20Requirement%20Object/invalid-oauth2-without-scopes.yaml and https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Security%20Requirement%20Object/invalid-openIdConnect-without-scopes.yaml

I think such validation would be also handlable with if/else statement in AsyncAPI JSON Schema (didn't check but I believe it is doable) but because we already had logic for validation of server security, it was easy to plug in there

### DIFF
--- a/lib/customValidators.js
+++ b/lib/customValidators.js
@@ -135,6 +135,8 @@ function validateOperationId(parsedJSON, asyncapiYAMLorJSON, initialFormat, oper
   return true;
 }
 
+/* eslint-disable sonarjs/cognitive-complexity */
+/* spliting it because it is 18 and not 15 lines would only make it more complex */
 /**
  * Validates if server security is declared properly and the name has a corresponding security schema definition in components with the same name
  * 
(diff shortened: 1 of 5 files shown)


lib/customValidators.js
```diff
@@ -135,6 +135,8 @@ function validateOperationId(parsedJSON, asyncapiYAMLorJSON, initialFormat, oper
   return true;
 }
 
+/* eslint-disable sonarjs/cognitive-complexity */
```
derberg, line 138: disabled and instead added extensive comments to code

lib/customValidators.js
```diff
@@ -235,6 +252,25 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
...
+/**
+ * Validates if given server security is not an empty array when security type requires it
+ * @private
+ * @param  {String} schemaType security type, like httpApiKey or userPassword
+ * @param  {String[]} specialSecTypes list of special types that do not have to be an empty array
+ * @param  {Object} secObj server security object
+ * @param  {String} secName name os server security object
+ * @returns {String[]} there are 2 elements in array, index 0 is the name of the security schema object and index 1 is it's type
+ */
+function hasSrvrSecScopes(schemaType, specialSecTypes, secObj, secName) {
```
derberg, line 264: I decided to have separate function instead of trying to mix it with `isSrvrSecProperArray` because code seems clearer with such an approach, and we can also nicely throw specific meaningful error

test/customValidators_test.js
```diff
@@ -543,6 +583,66 @@ describe('validateServerSecurity()', function() {
...
+      expect(e.title).to.equal('Server security value must not be an empty array if corresponding security schema type is oauth2 or openIdConnect. Add list of required scopes.');
+      expect(e.parsedJSON).to.deep.equal(parsedInput);
+      expect(e.validationErrors).to.deep.equal([
+        {
+          title: 'dummy/security/oauthsec security info must not have an empty array because its corresponding security schema type is: oauth2',
+          location: {
+            jsonPointer: '/servers/dummy/security/oauthsec',
+            startLine: 12,
+            startColumn: 28,
+            startOffset: 251,
```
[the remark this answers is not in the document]
derberg: I just pushed a fix, made sure offset is used all over the place

derberg:
@fmvilas it is possible, environments for windows and macOS are available afaik
I updated this issue with info about windows testing https://github.com/asyncapi/asyncapi/issues/423

PR #160, opened 2020-08-11
chore: Bump to release candidate and readme improvement

derberg:
**Description**

- Bump to release candidate and readme improvement


CONTRIBUTING.md
```diff
@@ -9,6 +9,14 @@ Contributions are more than welcome. If you want to contribute, please make sure
...
 4. A maintainer will review and, eventually, merge your Pull Request. Please, be patient as most of us are doing this in our spare time.
 
+### Develop
+
+1. Write code and tests.
+1. Make sure all tests pass `npm test`
+1. Make sure code is well formatted and secure `npm run lint`
+
+Release regenerates API documentation and browser bundle, so you do not have to regenerate it manually with `npm run docs` and `npm run prepublishOnly`.
+
```
fmvilas, line 19: Somehow unrelated to this PR but shouldn't we use the CONTRIBUTING file on the .github repo? https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md

Maybe we can just get rid of this file altogether.

README.md
```diff
@@ -2,35 +2,47 @@
   <br>
   <a href="https://www.asyncapi.org"><img src="https://github.com/asyncapi/parser-nodejs/raw/master/assets/logo.png" alt="AsyncAPI logo" width="200"></a>
   <br>
-  JS Parser
+  JavaScript Parser
 </h5>
-<h4 align="center">Parse and validate AsyncAPI documents</h4>
+<p align="center">
```
fmvilas, line 7: I liked the previous approach. Short and concise tagline. However, it's true we could improve the description as you did. What about having both? The tagline and the description below.

README.md
```diff
@@ -2,35 +2,47 @@
...
+  <em>Use this package to parse and validate AsyncAPI documents —either YAML or JSON— in your Node.js or browser application. Updated bundle for the browser is always attached to the GitHub Release.</em>
+</p>
 
----
+![npm](https://img.shields.io/npm/v/@asyncapi/parser?style=for-the-badge) ![npm](https://img.shields.io/npm/dt/@asyncapi/parser?style=for-the-badge)
 
-## :loudspeaker: ATTENTION:
+<br><br>
```
fmvilas, line 13: Why double `<br>` instead of a heading? I think we're losing clarity here.

Screenshot 2020-08-11 at 11 16 54

Also, the AsyncAPI 1.x notice feels very attached to the ToC which is completely a different section. What about moving the notice **after** the ToC and adding a heading saying "Table of Contents" or similar?

README.md
```diff
@@ -2,35 +2,47 @@
...
+  * [Example passing a URL](#example-passing-a-url)
+  * [Example using OpenAPI schemas](#example-using-openapi-schemas)
+  * [Example using RAML data types](#example-using-raml-data-types)
+  * [Example using Avro schemas](#example-using-avro-schemas)
+- [API documentation](#api-documentation)
+- [Custom message parsers](#custom-message-parsers)
+- [Error types](#error-types)
+- [Circular references](#circular-references)
+- [Contributing](#contributing)
+- [Contributors ✨](#contributors-%E2%9C%A8)
```
fmvilas, line 33: This link is not going to work because of the emoji

README.md
```diff
@@ -71,15 +83,23 @@ console.log(doc.info().title());
...
 
 Head over to [asyncapi/openapi-schema-parser](https://www.github.com/asyncapi/openapi-schema-parser) for more information.
 
-##### Example using RAML data types
+### Example using RAML data types
 
 Head over to [asyncapi/raml-dt-schema-parser](https://www.github.com/asyncapi/raml-dt-schema-parser) for more information.
 
-### Custom message parsers
+### Example using Avro schemas
```
fmvilas, line 94: I think we should move Avro to become the first one. In messaging world, it's much more extended than OpenAPI and RAML.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/parser",
-  "version": "0.33.1",
+  "version": "1.0.0-rc.1",
```
fmvilas, line 3: Looks like you're missing a merge with master?

derberg: no, rc releases go from my local, this PR is `bump to release candidate` 😉

README.md
derberg: ToC is there so you do not go below ToC but can just jump to a specific section. I've put it there as this is where we have notice in Generator readme, same with `<br>`. I also want to avoid writing `table of contents`, don't like to write obvious things

if I remove those brs, I think it looks better

Screenshot 2020-08-11 at 14 37 18

README.md
derberg: i removed it for 2 reasons, to be consistent with generator and because parser is a name that kind of suggests it is a parser

CONTRIBUTING.md
derberg: you are so right, removing

derberg:
@fmvilas I pushed some changes and also made some comments, have a look please

PR #161, opened 2020-08-13
chore: Improve circular refs support with circularProps and hasCircularProps

derberg:
**Description**

Unfortunately, the functionality that I added was useless in action. I tried it in 2 templates and I realized that marking the object as circular only if one of its props is circular doesn't make sense.

[BREAKING CHANGE] now object doesn't have the `x-parser-circular` is some props are circular, instead, it gets `x-parser-circular-props` with proper functions on schema model (`circularProps` and `hasCircularProps`). Thanks to it later in the template I can easily eliminate traversing circular props, and still go through those that are not circular, not listed in `x-parser-circular-props`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ For more information about the `ParserError` class, [check out the documentation
 Parser dereferences all circular references by default. In addition, to simplify interactions with the parser, the following is added:
 - `x-parser-circular` property is added to the root of the AsyncAPI document to indicate that the document contains circular references. Tooling developer that doesn't want to support circular references can use the `hasCircular()` function to check the document and provide a proper message to the user.
 - `x-parser-circular` property is added to every schema of array type that is circular. To check if schema is circular or not, you should use `isCircular()` function on a Schema model like `document.components().schema('RecursiveSelf').properties()['selfChildren'].isCircular()`.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -133,8 +133,9 @@ For more information about the `ParserError` class, [check out the documentation
...
-- `x-parser-circular` property is added to every schema where circular reference starts. You should use `isCircular` method on a Schema model like `document.components().schema('RecursiveSelf').properties()['selfChildren'].isCircular()`.
+- `x-parser-circular` property is added to the root of the AsyncAPI document to indicate that the document contains circular references. Tooling developer that doesn't want to support circular references can use the `hasCircular()` function to check the document and provide a proper message to the user.
+- `x-parser-circular` property is added to every schema of array type that is circular. To check if schema is circular or not, you should use `isCircular()` function on a Schema model like `document.components().schema('RecursiveSelf').properties()['selfChildren'].isCircular()`.
+- `x-parser-circular-props` property is added to every schema of object type with a list of properties that are circular. To check if schema has properties with circular references, you should use `hasCircularProps()` function. To get a list of properties with circular references, you should use `circularProps()` function.
```
fmvilas, line 138:
```suggestion
- `x-parser-circular-props` property is added to every schema of object type with a list of properties that are circular. To check if a schema has properties with circular references, you should use `hasCircularProps()` function. To get a list of properties with circular references, you should use `circularProps()` function.
```

PR #167, opened 2020-08-27
fix: Resolve concurrency bug during dereferencing

mauriceackel:
**Description**
By reusing the same refParser instance for all calls of the parse-function, it can come to concurrency issues if two parse-calls are executed at the same time (i.e. inside a Promise.all() call). This PR resolves the issue by creating a new json-schema-ref-parser for each call of the parse function.


derberg:
@mauriceackel I'm gonna close it in favour of https://github.com/asyncapi/parser-js/pull/168. Thanks a lot for good investigation

PR #168, opened 2020-08-28
chore: References resolved wrong when parsing is done inside Promise.all()

derberg:
**Description**

- References resolved wrong when parsing is done inside Promise.all(). It was caused by the fact that `$RefParser` was initialized globally and shared in concurrent parsing


magicmatatjahu (approved):

mauriceackel (approved):
LGTM!

PR #180, opened 2020-10-08
chore: Cleanup GitHub Actions workflows

DanielChuDC:
**Description**

1. Added `sentiment-analysis.yml` to .github/workflows

2. Rename pull_request from line 4 to pull_request_target in .github/workflows/welcome-first-time-contrib.yml file

3. Rename the `automerge-release-pr-bump.yml`to `automerge.yml` and modify the if statement according to the instruction

**Related issue(s)**
Fixes #174

PR #182, opened 2020-10-08
chore: Update automerge workflow to support dependabot

DanielChuDC:
Update automerge.yml

**Related issue(s)**
add-on fix for  #180


derberg:
@DanielChuDC thanks for the follow up!

derberg (approved):

PR #191, opened 2020-10-29
feat: add validation of query params in channel name

### LINKED ISSUE #145: Add validation of query params in channel name
#### Reason/Context
channel should not contain query parameters. The parser should throw a validation error once a query param is spotted.

#### Description
- add validation and test it against such an example https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Channels%20Object/invalid-query-param-used.yaml

juergenbr:
**Description**

- added validateChannelNames test suite
- added validateChannelName validator
- added method to check for URL parameters in string with RegEx

**Related issue(s)**
Fixes #145

### DIFF
--- a/lib/customValidators.js
+++ b/lib/customValidators.js
@@ -44,49 +44,6 @@ function validateServerVariables(parsedJSON, asyncapiYAMLorJSON, initialFormat) 
 }
 
 /**
- * Validates if parameters specified in the channel have corresponding parameters object defined
- * 
- * @private
- * @param  {Object} parsedJSON parsed AsyncAPI document
- * @param  {String} asyncapiYAMLorJSON AsyncAPI document in string
- * @param  {String} initialFormat information of the document was oryginally JSON or YAML
- * @returns {Boolean} true in case the document is valid, otherwise throws ParserError
- */
-function validateChannelParams(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
-  const chnls = parsedJSON.channels;
-  if (!chnls) return true;
-
-  const chnlsMap = new Map(Object.entries(chnls));
-  const notProvidedParams = new Map();
-    
-  chnlsMap.forEach((val, key) => {
-    const variables = parseUrlVariables(key);
-    const notProvidedChannelParams = notProvidedParams.get(tilde(key));
-    if (!variables) return;
-
-    const missingChannelParams = getMissingProps(variables, val.parameters);
-  
-    if (!missingChannelParams.length) return;
-
-    notProvidedParams.set(tilde(key), 
-      notProvidedChannelParams 
-        ? notProvidedChannelParams.concat(missingChannelParams) 
-        : missingChannelParams);
-  });
-
-  if (notProvidedParams.size) {
-    throw new ParserError({
-      type: validationError,
-      title: 'Not all channel parameters are described with parameter object',
-      parsedJSON,
-      validationErrors: groupValidationErrors('channels', 'channel does not have a corresponding parameter object for', notProvidedParams, asyncapiYAMLorJSON, initialFormat)
-    });
-  }
-
-  return true;
-}
-
-/**
  * Validates if operationIds are duplicated in the document
  * 
  * @private
@@ -240,27 +197,52 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
 }
 
 /**
- * Validates if channel name does not contain parameters
+ * Validates if parameters specified in the channel have corresponding parameters object defined and if name does not contain url parameters
+ * 
  * @private
  * @param  {Object} parsedJSON parsed AsyncAPI document
  * @param  {String} asyncapiYAMLorJSON AsyncAPI document in string
  * @param  {String} initialFormat information of the document was oryginally JSON or YAML
  * @returns {Boolean} true in case the document is valid, otherwise throws ParserError
  */
-function validateChannelName(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
+function validateChannels(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
   const chnls = parsedJSON.channels;
   if (!chnls) return true;
 
   const chnlsMap = new Map(Object.entries(chnls));
+  const notProvidedParams = new Map();
   const invalidChannelName = new Map();
-
+    
   chnlsMap.forEach((val, key) => {
-    const variables = parseUrlParameters(key);
-    if (!variables) return;
+    const variables = parseUrlVariables(key);
+    const parameters = parseUrlParameters(key);
+    if (!variables && !parameters)
+      return;
 
-    invalidChannelName.set(key, 
-      variables);
+    if (variables) {
+      const notProvidedChannelParams = notProvidedParams.get(tilde(key));
+      const missingChannelParams = getMissingProps(variables, val.parameters);
+  
+      notProvidedParams.set(tilde(key), 
+        notProvidedChannelParams 
+          ? notProvidedChannelParams.concat(missingChannelParams) 
+          : missingChannelParams);
+    }
+    
+    if (parameters) {
+      invalidChannelName.set(key, 
+        parameters);
+    }
   });
+
+  if (notProvidedParams.size) {
+    throw new ParserError({
+      type: validationError,
+      title: 'Not all channel parameters are described with parameter object',
+      parsedJSON,
+      validationErrors: groupValidationErrors('channels', 'channel does not have a corresponding parameter object for', notProvidedParams, asyncapiYAMLorJSON, initialFormat)
+    });
+  }
 
   if (invalidChannelName.size) {
     throw new ParserError({
@@ -275,9 +257,8 @@ function validateChannelName(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
 }
 
 module.exports = {
-  validateChannelParams,
   validateServerVariables,
   validateOperationId,
   validateServerSecurity,
-  validateChannelName
+  validateChannels
 };
--- a/lib/customValidators.js
+++ b/lib/customValidators.js
@@ -217,8 +217,6 @@ function validateChannels(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
     const variables = parseUrlVariables(key);
     const notProvidedChannelParams = notProvidedParams.get(tilde(key));
     const queryParameters = parseUrlQueryParameters(key);
-    if (!variables && !queryParameters)
-      return;
     
     //channel variable validation: fill return obeject with missing parameters
     if (variables) {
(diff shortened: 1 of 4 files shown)


derberg:
@juergenbr looks good. 2 questions:

- as you can see our linter doesn't like regex 😄 (btw, you can run in locally with `npm run lint`. What do you think about using URL API instead? like 
    ```
    const channelName = '/user/signedup?foo=1';
    const url = new URL('http://'+channelName);
    console.log(url.search) //logs ?foo=1;
    ```
- I think it would be possible to integrate new function with `validateChannelParams` (of course name would have to be changed to `validateChannel`. Asking because this way validator would iterate over list of channels only onece, not twice 🤔 what do you think?

juergenbr:
@derberg good points, yes. I thought that using a similar expression as parseUrlVariables would not be an issue, sorry.
How should a case look like when both checks fail? Should I handle them via a try catch block and output both error messages?

derberg:
@juergenbr you can keep regex, I just suggested different approaches as then you do not have issues with a linter. You can also just modify your function to have a similar approach as in `parseUrlVariables` and that is fine too. Just do not create new regex instance, my guess it is the issue.

regarding case when 2 issues show up at the same time, don't worry about it, they do not have to be thrown at the same time. We have such a case here already https://github.com/asyncapi/parser-js/blob/master/lib/customValidators.js#L180-L196

juergenbr:
@derberg  oh okay, no problem I'll fix it.
Thanks for the reference, then I'll handle it similarly!

juergenbr:
Test cases now work as expected, but due to the merge the new method got a little too complex so I split part of the logic. Please let me know if this is fine or if I should find another way.

lib/customValidators.js
```diff
@@ -239,9 +196,74 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
...
+
+  const chnlsMap = new Map(Object.entries(chnls));
+  const notProvidedParams = new Map();
+  const invalidChannelName = new Map();
+
+  chnlsMap.forEach((val, key) => {
+    const variables = parseUrlVariables(key);
+    const notProvidedChannelParams = notProvidedParams.get(tilde(key));
+    const parameters = parseUrlParameters(key);
+    if (!variables && !parameters)
```
derberg, line 220: because of the following ifs, I don't think we need this one

lib/customValidators.js
```diff
@@ -239,9 +196,74 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
...
+  if (!chnls) return true;
+
+  const chnlsMap = new Map(Object.entries(chnls));
+  const notProvidedParams = new Map();
+  const invalidChannelName = new Map();
+
+  chnlsMap.forEach((val, key) => {
+    const variables = parseUrlVariables(key);
+    const notProvidedChannelParams = notProvidedParams.get(tilde(key));
+    const parameters = parseUrlParameters(key);
```
derberg, line 219: maybe rename `parameters` to `queryParameters` and `parseUrlParameters` to `parseUrlQueryParameters`?

this will make code slightly clearer. Problem with current naming is that this function validates, query params and also if variables have params object. So you see "parameters" in 2 different contexts with totally different meaning. This is why I think we need to make super clear that it is `query param` here

lib/customValidators.js
```diff
@@ -239,9 +196,74 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
...
+    if (!variables && !parameters)
+      return;
+    if (variables) {
```
derberg, line 222: again regarding your concerns with code readability, just add a comment to this `if` statement that it concerns channel variables validation, and to another `if` add a comment that it is about validating query params

test/customValidators_test.js
```diff
@@ -300,7 +300,65 @@ describe('validateChannelParams()', function() {
...
 
-    expect(() => validateChannelParams(parsedInput, inputString, input)).to.throw('Not all channel parameters are described with parameter object');
+    expect(() => validateChannels(parsedInput, inputString, input)).to.throw('Not all channel parameters are described with parameter object');
+  });
+
+  it('should successfully validate channel name without variable', async function() {
+    const inputString = `{
+      "channels": {
+        "test/test01": {
+          "parameters": {
```
derberg, line 310: you are validating here a channel name, so instead of providing a parameter here, I think you could just pass empty object

test/customValidators_test.js
```diff
@@ -300,7 +300,65 @@ describe('validateChannelParams()', function() {
...
+    const parsedInput = JSON.parse(inputString);
+
+    expect(validateChannels(parsedInput, inputString, input)).to.equal(true);
+  });
+
+  it('should throw error that one of provided channel name is invalid', async function() {
+    const inputString = `{
+      "channels": {
+        "/user/signedup?foo=1": {
+          "subscribe": {
```
derberg, line 329: same as with the other test case, I think you can easily just pass an empty channel object

lib/customValidators.js
```diff
@@ -239,9 +196,74 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
...
+      type: validationError,
+      title: 'Channel names with parameters exist',
+      parsedJSON,
+      validationErrors: groupValidationErrors('channels', 'channels contain invalid name with url parameters ', invalidChannelName, asyncapiYAMLorJSON, initialFormat)
+    });
+  }
+
+  return true;
+
+  function setNotProvidedParams(variables, val, key, notProvidedChannelParams) {
```
derberg, line 252: I think you can move this function outside this function, as a separate private helper function, because it has the potential to be used in other validations, just don't forget about jsdoc.

Why? because I'm pretty sure you will need to use it in query params validation too, like here
```
    if (parameters) {
      invalidChannelName.set(key,
        parameters);
    }
```

this logic is here for handling use case where more that once the error of the same title appear during validation. In other words, I'm pretty sure the current implementation is not able to handle a case where you have 2 different channels containing query parameters. Please add such a test case and you will see what I mean.

derberg (changes requested):
it is getting a very good shape. I added some comments, long story short -> you need to add one more test case there that I do not think is covered with current implementation

juergenbr:
@derberg thanks again for your detailed feedback. I refactored the code to pull the additional method into the utils file. I also found a way to combine exceptions from both validations so that none "overrules" the other.

Regarding the case with 2 or more channels that contain query parameters: since the code you mentioned is inside the loop this case is covered. I added a test case and it is working as expected, returning 2 validation errors inside the array and returning them with correct messages. Same works if one channel has a missing parameter and one contains a query parameter.

lib/utils.js
```diff
@@ -304,3 +317,25 @@ utils.groupValidationErrors = (root, errorMessage, errorElements, asyncapiYAMLor
...
   return errors;
 };
+
+/**
+ * return all missing properties for a channel
+ *
+ * @function setNotProvidedParams
+ * @private
+ * @param  {String} variables array of all identified URL variables in a channel name
+ * @param  {Object} val The channel object for which to identify the missing parameters
```
derberg, line 327:
```suggestion
 * @param  {Object} val the channel object for which to identify the missing parameters
```

lib/utils.js
```diff
@@ -304,3 +317,25 @@ utils.groupValidationErrors = (root, errorMessage, errorElements, asyncapiYAMLor
 
   return errors;
 };
+
+/**
+ * return all missing properties for a channel
+ *
+ * @function setNotProvidedParams
+ * @private
+ * @param  {String} variables array of all identified URL variables in a channel name
```
derberg, line 326:
```suggestion
 * @param  {Array<String>} variables array of all identified URL variables in a channel name
```

lib/utils.js
```diff
@@ -304,3 +317,25 @@ utils.groupValidationErrors = (root, errorMessage, errorElements, asyncapiYAMLor
 
   return errors;
 };
+
+/**
+ * return all missing properties for a channel
```
derberg, line 322:
```suggestion
 * extend map with channel params missing corresponding param object
```

lib/utils.js
```diff
@@ -304,3 +317,25 @@ utils.groupValidationErrors = (root, errorMessage, errorElements, asyncapiYAMLor
...
+ *
+ * @function setNotProvidedParams
+ * @private
+ * @param  {String} variables array of all identified URL variables in a channel name
+ * @param  {Object} val The channel object for which to identify the missing parameters
+ * @param  {String} key the channel name. 
+ * @param  {Array<Object>} notProvidedChannelParams concatinated list of missing parameters for all channels
+ * @param  {Map} notProvidedParams result map of all missing parameters
```
derberg, line 330:
```suggestion
 * @param  {Map} notProvidedParams result map of all missing parameters extended by this function
```

derberg (changes requested):
I added some comments regarding naming of errors and jsdoc.
Great idea with displaying errors from different validations, I like it.

What is missing:
- test case for channel that has no parameter object and name of channel is `user/{userId}/signedup?foo=1`. So 2 different errors for the same channel
- in tests, I cannot see in error information about the channel (the one that failed validation) locatin in the file. Things like `endColumn` and others. Notice that channel variables validaton uses `tilde` function to make it work

lib/customValidators.js
```diff
@@ -239,9 +196,61 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
...
+    //channel name validation: fill return object with channels containing query parameters
+    if (queryParameters) {
+      invalidChannelName.set(tilde(key),
+        queryParameters);
+    }
+  });
+
+  //combine validation errors of both checks and output them as one array
+  const parameterValidationErrors = groupValidationErrors('channels', 'channel does not have a corresponding parameter object for', notProvidedParams, asyncapiYAMLorJSON, initialFormat);
+  const nameValidationErrors = groupValidationErrors('channels', 'channel contains invalid name with url query parameters ', invalidChannelName, asyncapiYAMLorJSON, initialFormat);
```
derberg, line 235:
```suggestion
  const nameValidationErrors = groupValidationErrors('channels', 'channel contains invalid name with url query parameters', invalidChannelName, asyncapiYAMLorJSON, initialFormat);
```

derberg: all is good ready to merge, just please remove this empty space as later in the error it shows with empty space before colon `channel contains invalid name with url query parameters : ` which looks strange and is not consistent with other errors

derberg (approved):
LGTM. Great work @juergenbr !

derberg:
@all-contributors please add @juergenbr for code

WaleedAshraf:
This is failing with error
```
err { Error: Invalid URL: http:///
        at parse (/Users/waleedashraf/projects/waleed/asyncapi-validator/node_modules/@asyncapi/parser/lib/parser.js:108:11)
        at process._tickCallback (internal/process/next_tick.js:68:7)
      type: 'https://github.com/asyncapi/parser-js/unexpected-error',
      title: 'Invalid URL: http:///',
      parsedJSON:
       { asyncapi: '2.0.0',
         id: 'urn:com:slack:rtm:api',
         info: { title: 'Slack Real Time Messaging API', version: '1.0.0' },
         servers: { production: [Object] },
         channels: { '/': [Object] },
         components:
          { securitySchemes: [Object],
            schemas: [Object],
            messages: [Object] } },
      message: 'Invalid URL: http:///' } Invalid URL: http:///
```

Schema: https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/slack-rtm.yml

Because the channel name is just `/`

derberg:
@WaleedAshraf fix provided https://github.com/asyncapi/parser-js/pull/196

PR #196, opened 2020-11-13
fix: channels with name / fail on validation

derberg:
**Description**

- channels with name / fail on validation

See also https://github.com/asyncapi/parser-js/pull/191#issuecomment-726451384


WaleedAshraf (approved):
🎉

PR #198, opened 2020-11-17
feat: add a traverse schema function

### LINKED ISSUE #166: Parser API should expose functions for traversing through schemas
#### Reason/Context
Traversing through JSON schemas with nested structure is not an easy task and requires a good knowledge of JSON Schema. The parser should make it easy especially that it already has a needed logic for anonymous id's assignment to messages and schemas, or detection of circular references. 

#### Description
- expose something like this on the API level `asyncapi.traverseSchemas(callback, schemaType)`
  - `schemaType` would by default traverse all schemas, but you could specify that you for example only want `payload` schemas or only `parameters` schemas
  - `callback` would be a function that you can pass into the traverse, that will be invoked on each schema found during 
the traverse
 - would be great if the callback execution result could be returned by the function
- refactor existing logic for traversing schemas. We cannot just expose 1:1 what we have. Current traversing logic cares directly for circular refs detection and marking schemas as circular during parsing. Would be good to not just repeat everything again when traversing through the document after it is parsed
- this is where the current traversing logic is located
  - https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L378
  - https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L323

jonaslagoni:
**Description**

The PR contains the following updates
- Exposed a traverseSchema function 
- Split out the mark recursive functionality
- Added test to the traverseSchema function
- Added special iterator type to include certain schemas
- Revisited all the code which recursively tries to access schemas

**Related issue(s)**
Fixes #166


derberg:
@jonaslagoni awesome you picked it up. Let me know once you are ready for some hands on, I'd like to test it on the template for templates, and basically  remove https://github.com/asyncapi/template-for-generator-templates/blob/main/filters/mermaidDiagram.js#L75-L177

jonaslagoni:
Hmm, this is not as straight forward as I first assumed since it is implemented with `markcircular` on the previous schema not the currently crawled schema. And its not possible to directly overwrite the schema because of the nature of JavaScript passes objects as copy of a reference, have to do some sort of recursive overwrite in that case :thinking:

derberg:
@jonaslagoni oh yes, I think some refactor might be needed, not trivial one, this is why I've put it here and didn't work on it immediately

lib/models/asyncapi.js
```diff
@@ -276,121 +351,139 @@ function addNameToKey(messages, number) {
...
  * @private
  * @param {AsyncAPIDocument} doc 
- * @param {FoundSchemaCallback} callback
+ * @param {FoundSchemaCallback} cb
+ * @param {SchemaTypesToIterate[]} schemaTypesToIterate
  */
-function schemaDocument(doc, callback) {
+function traverseAsyncApiDocument(doc, cb, schemaTypesToIterate) {
+  if (!schemaTypesToIterate) {
+    schemaTypesToIterate = [
```
magicmatatjahu, line 421: We can simplify it by:

```js
schemaTypesToIterate = Object.values(SchemaTypesToIterate)
```

:)

lib/models/asyncapi.js
```diff
@@ -276,121 +351,139 @@ function addNameToKey(messages, number) {
...
+ * @param {FoundSchemaCallback} cb
+ * @param {SchemaTypesToIterate[]} schemaTypesToIterate
+ */
+function traverseChannel(channel, cb, schemaTypesToIterate) {
+  if (schemaTypesToIterate.includes(SchemaTypesToIterate.parameters)) {
+    Object.values(channel.parameters()).forEach(p => {
+      traverseSchema(p.schema(), cb, null, schemaTypesToIterate);
+    });
+  }
+  if (channel.hasPublish()) {
```
magicmatatjahu, line 460: As I remember (but I can be wrong), AsyncAPI only support one publish or subscribe per channel. You cannot combine publish/subscribe. @derberg I am right?

```js
if (channel.hasPublish() || channel.hasSubscribe()) {
    channel.publish().messages().forEach(m => {
      traverseMessage(m, cb, schemaTypesToIterate);
    });
}
```

will be enough.

magicmatatjahu (commented):
Awesome! I only have a suggestion to change `cb` name to more verbose `callback` :)

lib/models/asyncapi.js
jonaslagoni: The spec does not specify only one operation is allowed. Also this combination is used in request/reply (since its not support yet) that both gets specified :slightly_smiling_face:

lib/models/asyncapi.js
jonaslagoni: I remember trying different solutions with for loops but yes, this is way better :smile:

lib/models/asyncapi.js
```diff
@@ -276,121 +351,139 @@ function addNameToKey(messages, number) {
...
+  if (schemaTypesToIterate.includes(SchemaTypesToIterate.oneOfs)) {
     checkCombiningSchemas(schema.oneOf());
   }
-
-  seenObj.pop();
 }
 
 /**
- * Go through each channel and for each parameter, and message payload and headers recursively call the callback for each schema.
+ * Go through each channel and for each parameter, and message payload and headers recursively call the cb for each schema.
```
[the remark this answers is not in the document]
jonaslagoni: Yea agreed

lib/models/asyncapi.js
```diff
@@ -276,121 +351,139 @@ function addNameToKey(messages, number) {
...
-          if (crawl(combineSchema, seenObj, callback)) markCircular(schema);
-        });
-      }
-    };
+  const checkCombiningSchemas = (combineArray) => {
+    if (combineArray !== null && combineArray.length > 0) {
+      combineArray.forEach(combineSchema => {
+        traverseSchema(combineSchema, cb, null, schemaTypesToIterate);
+      });
+    }
```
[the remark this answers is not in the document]
jonaslagoni: perfect :+1:

lib/models/asyncapi.js
```diff
@@ -196,6 +208,69 @@ class AsyncAPIDocument extends Base {
...
+
+  //Mark the schema as recursive
+  const markCircular = (schema, prop) => {
+    if (schema.type() === 'array') return schema.json()[String(xParserCircle)] = true;
+    const circPropsList = schema.json()[String(xParserCircleProps)] || [];
+    if (prop !== undefined) {
+      circPropsList.push(prop);
+    }
+    schema.json()[String(xParserCircleProps)] = circPropsList;
+  };
```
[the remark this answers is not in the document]
jonaslagoni: I can see what you mean, however this reference are only created each time you create a new instance of `AsyncAPIDocument` and are never used since. So moving it outside does not create any optimization in this case :)

lib/models/asyncapi.js
```diff
@@ -196,6 +208,69 @@ class AsyncAPIDocument extends Base {
...
   }
+
+  /**
+   * @callback TraverseSchemas
+   * @param {Schema} schema
+   * @param {SchemaIteratorCallbackType} callbackType
+   * @param {String} propName if the schema is from a property get the name of such
+   * @returns {boolean} should it continue deeper
+   * 
+   */
```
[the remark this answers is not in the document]
jonaslagoni: It is used in the `traverseSchemas` function and are there to document the signature of the callback.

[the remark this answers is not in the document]
magicmatatjahu: You are right, sorry :)

lib/models/asyncapi.js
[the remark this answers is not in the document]
magicmatatjahu: You are right, sorry :)

magicmatatjahu (approved):

magicmatatjahu (dismissed):

derberg:
@jonaslagoni I'll give it a try in template for templates. I'll keep ya posted

derberg:
so far I'm running it locally without introduction any changes to template, like this:
```
../generator/cli.js https://raw.githubusercontent.com/asyncapi/generator/v1.0.1/test/docs/dummy.yml ./ -o output
```

- where `../generator/cli.js` points to my local generator with parser pointing to your fork
- where `./` indicates I use local template which is template for templates

I get this error:

tl;dr property of type string, inside object, doesn't get anonymousId assigned

```
Something went wrong:
Template render error: (/Users/wookiee/sources/generator/node_modules/template-for-generator-templates/template/index.html)
  Template render error: (/Users/wookiee/sources/generator/node_modules/template-for-generator-templates/partials/diagramContent.html) [Line 13, Column 48]
  TypeError: Cannot read property 'startsWith' of undefined
    at Object._prettifyError (/Users/wookiee/sources/generator/node_modules/nunjucks/src/lib.js:36:11)
    at /Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:561:19
    at eval (eval at _compile (/Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:631:18), <anonymous>:49:11)
    at /Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:569:11
    at Template.root [as rootRenderFunc] (eval at _compile (/Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:631:18), <anonymous>:63:3)
    at Template.render (/Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:550:10)
    at eval (eval at _compile (/Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:631:18), <anonymous>:48:10)
    at fn (/Users/wookiee/sources/generator/node_modules/a-sync-waterfall/index.js:26:24)
    at /Users/wookiee/sources/generator/node_modules/a-sync-waterfall/index.js:66:22
    at executeSync (/Users/wookiee/sources/generator/node_modules/a-sync-waterfall/index.js:8:15)
    at /Users/wookiee/sources/generator/node_modules/a-sync-waterfall/index.js:65:11
    at eval (eval at _compile (/Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:631:18), <anonymous>:44:1)
    at createTemplate (/Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:313:9)
    at handle (/Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:325:11)
    at /Users/wookiee/sources/generator/node_modules/nunjucks/src/environment.js:337:9
    at next (/Users/wookiee/sources/generator/node_modules/nunjucks/src/lib.js:280:7)
```

Of cource when I try the same with latest generator, all is good. Something is wrong here https://github.com/asyncapi/template-for-generator-templates/blob/main/filters/mermaidDiagram.js#L13. There is a moment when `https://github.com/asyncapi/template-for-generator-templates/blob/main/filters/mermaidDiagram.js#L35` gets `undefined` in `propSchemaId`. I narrowed it down that problem is with this prop `dummyRecursiveProp2` from `dummyRecursiveObject` schema, it doesn't get anonymousId assigned at all.

I hope this helps. I did not check where exactly is the problem on your side but this investigation should make it easier to find

jonaslagoni:
> There is a moment when `https://github.com/asyncapi/template-for-generator-templates/blob/main/filters/mermaidDiagram.js#L35` gets `undefined` in `propSchemaId`. I narrowed it down that problem is with this prop `dummyRecursiveProp2` from `dummyRecursiveObject` schema, it doesn't get anonymousId assigned at all.

Found the issue, creating some tests and then pushing the change.

jonaslagoni:
@derberg done. Tried testing it with your document and it seems to work. `return` is a dangerous thing inside a loop that should `continue` :roll_eyes:

derberg:
- I wonder if we could get `schemaTypesToIterate` also into API docs, like have a list of supported types?
- I get strange results after your changes, results in diagram. Below you can find diagram before and after your change. Look at `dummyObject`.
After: 
Screenshot 2020-11-30 at 17 11 45
Before: 
Screenshot 2020-11-30 at 18 18 08

`dummyObject.dummyObjectProp2` is a reference to `dummyRecursiveObject`, and we know `dummyRecursiveObject` has recursive property so `dummyObject` should also list `dummyObjectProp2` as circular prop, right? or my thinking is stupid as then it would mean `dummyCreated` also should list `dummyObject` as circular prop . Ok, I think you mixed bug with this PR 😄 

ok, after checking with html template, I'm now actually sure you fixed a bug 👏 
Before:
Screenshot 2020-11-30 at 18 55 57

After:
Screenshot 2020-11-30 at 18 55 39

--- 
Now I will focus on checking if I can actually use this new function to refactor my filter

lib/models/asyncapi.js
```diff
@@ -211,6 +223,69 @@ class AsyncAPIDocument extends Base {
   hasCircular() {
     return !!this._json[String(xParserCircle)];
   }
+
+  /**
+   * @callback TraverseSchemas
```
derberg, line 228: is it correct? I see callback is called like this 👇 
`callback(schema, prop, SchemaIteratorCallbackType.NEW_SCHEMA)`

please also extend descriptions of each property as jsdoc doesn't show things like `NEW_SCHEMA/END_SCHEMA`

derberg (changes requested):
now this is what I call a good feature - when by using it I add just one line and remove 108 -> https://github.com/asyncapi/template-for-generator-templates/pull/5/files 👏 

I only shared my concerns about jsdocs, that there is one small error I think, and that those special types you added are not represented in jsdoc

lib/models/asyncapi.js
jonaslagoni: The jsdocs have been updated to include descriptions and the enums 😄

jonaslagoni:
Sorry for requesting review before fixing the lint errors. It is ready now 😄

lib/models/asyncapi.js
derberg: @jonaslagoni what about this

>is it correct? I see callback is called like this 👇
callback(schema, prop, SchemaIteratorCallbackType.NEW_SCHEMA)

don't you think that below table suggests you get the callback type as second argument, while in fact it is third? I think that order in jsdoc should change
Screenshot 2020-12-07 at 13 16 23

jonaslagoni: 🤔  good catch, did not even catch that in your comment 😄 That is changed.

derberg (approved):
Awesome work man!
Let us not merge yet, I want to first fix the issue with twitter action to not tweet again that it this feature was done by the bot 😄 
I'll work on it today

derberg:
@jonaslagoni I'm almost done with https://github.com/asyncapi/parser-js/pull/208, one repo left to get pr testing there with mandatory option. I want to merge it first too because we will be able to test with your feature how the automated bump works 😄

jonaslagoni:
> @jonaslagoni I'm almost done with #208, one repo left to get pr testing there with mandatory option. I want to merge it first too because we will be able to test with your feature how the automated bump works 😄

Sounds good 👍

PR #205, opened 2020-11-25
chore(ci): dispatch release event to Generator

fmvilas:
**Description**

This is part of an effort to automatically create PRs on the Generator package every time there is a new release on the parser.

This PR introduces a new workflow job that sends an event to the Generator every time there's a new release on this package. The event contains the package name (`@asyncapi/parser-js`) and the release version.

The event looks like the following:

```json
{
  "name": "@asyncapi/parser-js",
  "version": "1.0.1"
}
```

**Related issue(s)**
https://github.com/asyncapi/generator/pull/452


.github/workflows/release.yml
```diff
@@ -66,4 +66,25 @@ jobs:
...
+    needs: release
+    strategy:
+      matrix:
+        repo: ['asyncapi/generator']
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - id: package_details
```
derberg, line 81: we do not need this custom action. We already have this script `npm run get-version` and we can add script that echos `process.env.npm_package_name` too.

so eventually this step will look like this:
```
      - name: Get version and name from package.json
        id: package_details
        run: |
          echo "::set-output name=version::$(npm run get-version --silent)"
          echo "::set-output name=name::$(npm run get-name --silent)"
```

.github/workflows/release.yml
```diff
@@ -66,4 +66,25 @@ jobs:
...
           body: 'Version bump in package.json and package-lock.json for release [${{ steps.extractver.outputs.version }}](https://github.com/${{github.repository}}/releases/tag/v${{ steps.extractver.outputs.version }})'
-          branch: version-bump/${{ steps.extractver.outputs.version }}
\ No newline at end of file
+          branch: version-bump/${{ steps.extractver.outputs.version }}
+  
+  dispatch:
+    needs: release
+    strategy:
+      matrix:
+        repo: ['asyncapi/generator']
```
derberg, line 75: I'm 🤔 if this is the right way to do it, like maintain this matrix on a repo level instead of in central location. 
advantage of central location would be that this could be a separate workflow that reacts on `release` event, would be easy to maintan and update in all the other repositories 🤔

derberg (changes requested):

.github/workflows/release.yml
fmvilas: I actually like the action better. We can publish this action instead of having it here and then we would not need to touch every package.json in every project we want to use it.

I mean, having something like this:

```
      - name: Get version and name from package.json
        id: package_details
        uses: fmvilas/package-details@v1
```

Don't you think it's better? No need to add custom scripts to package.json.

.github/workflows/release.yml
fmvilas: Is this possible? Because honestly, I have no idea how to do it.

.github/workflows/release.yml
derberg: I'm always in position to create custom action, or use community action only when it is really really necessary and saves time (this is why in generator docker release is not done with any special action).

I suggested npm scripts as they are anyway already in use in all the release workflows in all repos. I just prefer to stick to one. If you insist on custom action, then sure but have code separated from this repo, and we need to make sure it is used everywhere and stop using npm scripts.

Also, better than custom action would be to contribute here https://github.com/EndBug/version-check to get then expose package name to output. We already use this action here and in html-template to identify if there is minor or patch release, so we could make it a standard and get rid of npm scripts

.github/workflows/release.yml
derberg: we just need to decide where to store it, no GitHub Actions native functionality
we could have it as a yaml file in `.github` repository for example.

it is still not the best anyway, still pretty manual though

derberg (changes requested):

.github/workflows/release.yml
fmvilas: It's the same effort for me to add another package.json script to get the package name and replace the current methodology with another action. I'd go for my suggestion, really. Happy to contribute to the action you mentioned but it's really a stupid action and there are a few like the one you mentioned. I'd use my action first and contribute to that package. If they don't accept the PR (and it would be expected because it's called `version` check) at least we have it working.

.github/workflows/release.yml
fmvilas: So you mean, every repo will publish an event to every other repo —even if they don't have any kind of dependency between them— and then, on the other side, we only listen for events coming from the dependencies?

PR #209, opened 2020-12-07
ci: fix release workflow step that is responsible for handling twitter

derberg:
**Description**

Since we introduced a step that validates what type of release we have (patch/minor/major) we have a problem with twitter step. It is not run when we do release, and instead, it runs next time we do the push, push commit with pump in package.json. As a result, in tweet, instead of putting the name of the user that is responsible for major/minor release, we put the name of the user that pushed bump commit, always bot user. 

This PR should fix it by checking the version from package.json on master against the version from package.json available in the workflow. Taken from https://github.com/EndBug/version-check#static-checking-with-your-latest-version-on-npm

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,20 +43,16 @@ jobs:
         run: npm run release
       - name: Get version from package.json after release step
         id: extractver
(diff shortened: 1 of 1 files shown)


jonaslagoni (approved):
👍

PR #214, opened 2020-12-10
fix: apply traits to standalone messages from components section

### LINKED ISSUE #210: messageTrait not applied when message not used in a channel
**Description**

We want to integrate the react component into our project for displaying specifications of messages to be implemented. Because the messages are not implemented but only specified by us, we define them only in the `components/messages` section and don't use the `channels` section (passing an empty object).

However, if the message is not referenced from a channel, it doesn't apply the used `messageTraits` to the message.

**Expected result**

Message traits should be applied to the message even if the message is not used in channel. All properties from the used trait should be shown as part of the message.

**Actual result**

Message trait is only applied if the message is used within a channel. In the case the message is used in a channel, you can see the properties from the trait in both the channel, and the message object.

**Steps to reproduce**
Find a minimal example here. The "light measured" message is used in a channel and the "commonHeaders" trait is applied correctly ("my-app-header" property shown in the channel and the message).

The "turn off" message is not used in a channel. The "commonHeaders" trait is NOT applied and you won't find the headers section with the "my-app-header" property.

[message-channel-trait-example.txt](https://github.com/asyncapi/asyncapi-react/files/5671142/message-channel-trait-example.txt)

**Troubleshooting**

c-pius:
**Description**

@derberg I tried to give it a shot but I am not really sure if the provided function is at the right place and also how the `validateAndConvertMessage ` function works. I included the `"x-parser-schema-id":"<anonymous-schema-1>"` properties in the output schemas as far as I understood them. Unfortunately, they are only generated for messages that are used in a channel and not the standalone messages. I assume you would know how to fix those problems?

Besided, `npm run test-browser` was not working on my machine (OSX) from the beginning I cloned the repo. I am on node  `v12.14.1`.

**Related issue(s)**
Resolves #210


lib/parser.js
```diff
@@ -253,4 +254,28 @@ async function customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+ * 
+ * @param  {Object} parsedJSON parsed AsyncAPI document
+ * @param  {String} asyncapiYAMLorJSON AsyncAPI document in string
+ * @param  {String} initialFormat information of the document was originally JSON or YAML
+ * @param  {Object} options Configuration options.
+ */
+async function customMessagesOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options) {
+  const promisesArray = [];
+
+  Object.entries(parsedJSON.components.messages).forEach(([messageName, message]) => {
```
magicmatatjahu, line 272: Before accesing to `parsedJSON.components.messages` please first check that `components.messages` path exists, because either `components` and `messages` in `components` are optional, something like:

```js
async function customMessagesOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options) {
  if (!parsedJSON.components || !parsedJSON.components.messages) return;
  ...your logic
}
```

lib/parser.js
```diff
@@ -253,4 +254,28 @@ async function customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+    if (options.applyTraits) {
+      applyTraits(message);
+    }
+    const pathToPayload = `/components/messages/${messageName}/payload`;
+    promisesArray.push(validateAndConvertMessage(message, asyncapiYAMLorJSON, initialFormat, parsedJSON, pathToPayload));
```
magicmatatjahu, line 277: To your question about `validateAndConvertMessage` function. In AscyncAPI spec you can define in which format you want describe messages, in jsonSchema, avro, raml (at the moment officially we only support them) and we validate and parse message in this function based on format, so your usage is good :) 
Ref in doc: https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#messageObject

magicmatatjahu (changes requested):
One thing to fix :) Thanks for contribution!

magicmatatjahu:
> I included the "x-parser-schema-id":"<anonymous-schema-1>" properties in the output schemas as far as I understood them. Unfortunately, they are only generated for messages that are used in a channel and not the standalone messages. I assume you would know how to fix those problems?

The problem is here https://github.com/asyncapi/parser-js/blob/d35d59a4b423145c0e125436500c0116ce515c3a/lib/models/asyncapi.js#L78 As I see we don't traverse in schemas inside the messages in `components`. We only traverse schemas in `channels.message`, `components.parameters` and `component.schemas`. @derberg should we fix it in this PR or leave it as a followup?

test/parse_test.js
```diff
@@ -20,6 +22,8 @@ const invalidYamlOutput = '{"asyncapi":"2.0.0","info":{"version":"1.0.0"},"chann
...
+const outputJSONNoChannels = '{"asyncapi":"2.0.0","info":{"title":"My API","version":"1.0.0"},"channels":{},"components":{"messages":{"testMessage":{"payload":{"type":"object","properties":{"name":{"type":"string","x-parser-schema-id":"<anonymous-schema-1>"}},"x-parser-schema-id":"testSchema"},"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string","x-parser-schema-id":"<anonymous-schema-3>"}},"x-parser-schema-id":"<anonymous-schema-2>"},"x-parser-original-traits":[{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}],"schemaFormat":"application\/vnd.aai.asyncapi;version=2.0.0","x-parser-message-name":"testMessage"}},"schemas":{"testSchema":{"type":"object","properties":{"name":{"type":"string","x-parser-schema-id":"<anonymous-schema-1>"}},"x-parser-schema-id":"testSchema"}},"messageTraits":{"extension":{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string","x-parser-schema-id":"<anonymous-schema-3>"}},"x-parser-schema-id":"<anonymous-schema-2>"}}}}}';
```
derberg, line 25:
```suggestion
const outputJSONNoChannels = '{"asyncapi":"2.0.0","info":{"title":"My API","version":"1.0.0"},"channels":{},"components":{"messages":{"testMessage":{"payload":{"type":"object","properties":{"name":{"type":"string"}},"x-parser-schema-id":"testSchema"},"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}},"x-parser-original-traits":[{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}],"schemaFormat":"application/vnd.aai.asyncapi;version=2.0.0","x-parser-message-name":"testMessage"}},"schemas":{"testSchema":{"type":"object","properties":{"name":{"type":"string"}},"x-parser-schema-id":"testSchema"}},"messageTraits":{"extension":{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}}}}';
```

test/parse_test.js
```diff
@@ -20,6 +22,8 @@ const invalidYamlOutput = '{"asyncapi":"2.0.0","info":{"version":"1.0.0"},"chann
...
...rs":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}],"schemaFormat":"application\/vnd.aai.asyncapi;version=2.0.0","x-parser-message-name":"channelMessage"},"testMessage":{"payload":{"type":"object","properties":{"name":{"type":"string","x-parser-schema-id":"<anonymous-schema-3>"}},"x-parser-schema-id":"testSchema"},"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string","x-parser-schema-id":"<anonymous-schema-2>"}},"x-parser-schema-id":"<anonymous-schema-1>"},"x-parser-original-traits":[{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}],"schemaFormat":"application\/vnd.aai.asyncapi;version=2.0.0","x-parser-message-name":"testMessage"}},"schemas":{"testSchema":{"type":"object","properties":{"name":{"type":"string","x-parser-schema-id":"<anonymous-schema-3>"}},"x-parser-schema-id":"testSchema"}},"messageTraits":{"extension":{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string","x-parser-schema-id":"<anonymous-schema-2>"}},"x-parser-schema-id":"<anonymous-schema-1>"}}}}}';
```
derberg, line 26:
```suggestion
const outputJSONMessagesChannels = '{"asyncapi":"2.0.0","info":{"title":"My API","version":"1.0.0"},"channels":{"mychannel":{"publish":{"message":{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string","x-parser-schema-id":"<anonymous-schema-2>"}},"x-parser-schema-id":"<anonymous-schema-1>"},"x-parser-original-traits":[{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}],"schemaFormat":"application/vnd.aai.asyncapi;version=2.0.0","x-parser-message-name":"channelMessage"}}}},"components":{"messages":{"channelMessage":{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string","x-parser-schema-id":"<anonymous-schema-2>"}},"x-parser-schema-id":"<anonymous-schema-1>"},"x-parser-original-traits":[{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}],"schemaFormat":"application/vnd.aai.asyncapi;version=2.0.0","x-parser-message-name":"channelMessage"},"testMessage":{"payload":{"type":"object","properties":{"name":{"type":"string"}},"x-parser-schema-id":"testSchema"},"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}},"x-parser-original-traits":[{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}],"schemaFormat":"application/vnd.aai.asyncapi;version=2.0.0","x-parser-message-name":"testMessage"}},"schemas":{"testSchema":{"type":"object","properties":{"name":{"type":"string"}},"x-parser-schema-id":"testSchema"}},"messageTraits":{"extension":{"x-some-extension":"some extension","headers":{"type":"object","properties":{"some-common-header":{"type":"string"}}}}}}}';
```

derberg (changes requested):
> I tried to give it a shot but I am not really sure if the provided function is at the right place and also how the validateAndConvertMessage function works. 

It looks good, thanks a bunch for jumping into it 🙇🏼 
@magicmatatjahu explained well what is the purpose of this function. I'm only afraid that with the current implementation we will have messages (the ones from components.messages that are referred to channels) that are validated and converted twice, like this one -> https://github.com/asyncapi/parser-js/pull/214/files#diff-b335089706650932f2429d8b648707a16777501ad4e410fcbd35be14900707d8R14 which is a performance issue in my opinion because in large componentized specs with channels the validation will simply be doubled 🤔 I'm opened for suggestions 😄 

> I included the "x-parser-schema-id":"<anonymous-schema-1>" properties in the output schemas as far as I understood them. Unfortunately, they are only generated for messages that are used in a channel and not the standalone messages. I assume you would know how to fix those problems?

a simple fix would bring the same issue as in my previous comment, I think. Anonymous schema id goes only to messages in channels. That would be too much for this PR for you to solve, and anyway it is a different topic that requires separate PR. Could you please create a separate issue, please 🙏🏼 

I made code change suggestions that you can apply for now.

> Besides, npm run test-browser was not working on my machine (OSX) from the beginning I cloned the repo. I am on node v12.14.1.

it was not working because of the missing condition @magicmatatjahu mentioned in one of his comments. Spec used in the browser test is as simple as
```
asyncapi: 2.0.0
info:
  title: My API
  version: 1.0.0
channels:
  "/test/tester":
    subscribe:
      message: {}
```
so it covers exactly the same case that @magicmatatjahu described, a spec without messages in components, and in fact, no components object at all. It would be awesome if you could add a test case for such a document to `parse_test` 🙏🏼 

As a side note, I will create a PR to console log errors in the browser test, so nobody has to disable the headless mode of the browser and dig every issue on his/her own

---
last but not least -> you can take this PR out from draft to enable tests as you are pretty advanced with this PR and we anyway jumped into the review

c-pius:
added the guard that `components.messages` is defined and an additional test case for when there is no components object.

> [duplicate application of traits] is a performance issue in my opinion because in large componentized specs with channels the validation will simply be doubled 🤔 I'm opened for suggestions 😄

I tried to give it a thought but I guess that would require to somehow apply the traits in conjunction with resolving the `$ref` properties? Maybe also remember the original `$ref` in a parser extension attribute and if there, replace the message in the operation with the referenced message object where traits have been applied brefore?

derberg:
@c-pius trying to combine it with resolving of `$ref`-s would be quite a refactor and anyway this part is already complicated because of circular refs.

I like your idea about simple remembering that messages were processed already. Just remember it is not just `applyTraits` that is called twice on messages but `validateAndConvertMessage` too.

What if we revert the order? like process Messages from components first, and then Channels. Then in the case of `applyTraits` we do not have to even remember anything, we don't really have a performance issue 🤔 if I'm wrong, blame Monday 😄 but looking at 👇🏼 we are safe because after processing traits, `traits` prop is removed, so there is no risk it will be processed again? right?

```
function applyTraits(js) {
  if (Array.isArray(js.traits)) {
    for (const trait of js.traits) {
      for (const key in trait) {
        js[String(key)] = mergePatch(js[String(key)], trait[String(key)]);
      }
    }

    js['x-parser-original-traits'] = js.traits;
    delete js.traits;
  }
}
```

The only real concern is `validateAndConvertMessage`. We could add another prop there like `x-validated=true` and then just check if it is there, and then just not run the validation again. I'm only not sure this is a good prop name, maybe `x-origin=components` 🤔 

But in general, it makes sense right? to process messages from components first, and then just process those unprocessed that are directly under channels?

c-pius:
Actually I had already tried this using a check if the `x-parser-message-name` property is already there but only now noticed that this is added at a later point 🤦🏻  So yeah, your suggestion makes sense and would work 👍🏻 

Regarding the name of the marker property I don't really mind any. I guess `x-validated` would be a bit more straightforward and cleaner to handle completely within `validateAndConvertMessage`. If we use `x-origin` we would actually need to either set it as part of `customMessagesOperations` func (therefore spread the logic over two functions, or pass another param to the `validateAndConvertMessage` indicating the component origin (maybe also possible using the existing `pathToPayload` param but I think not ideal as well).

 I assume you intended it to be `x-parser-validated`, right? `x-validated` would have a high probability to collide with user defined extensions on a message I guess.

derberg:
@c-pius yeap, `x-parser-` prefix is a must, sorry

derberg:
I'm just afraid about `validated` as it will kinda imply that other messages from channels are not validated 😄 
unless you remove it at the end from the document?

c-pius:
> I'm just afraid about validated as it will kinda imply that other messages from channels are not validated 😄
unless you remove it at the end from the document?

actually I would have put checking and setting the `x-parser-validated` field both into the `validateAndConvertMessage` function so that no matter if it is defined in channels or messages, if it has been validated once the flag is set. Probably something like:

```js
async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload) {
  if (xParserValidated in msg && msg[xParserValidated] === true) {
    //only validate the message if it has not been done before
    return
  }

  //... function logic

  msg[xParserValidated] = true
}
```

or do you prefer it to be only added for components/messages and removed again afterwards?

derberg:
you are right, good approach with having the logic in `validateAndConvertMessage`

I'm not 100% sure if `x-parser-validated` should stay or not because I'm not sure what it could be used for if it stays there in the document. I just prefer to leave something that might be useful. 

what about `x-parser-original-schema-format`? every custom parser adds it:
- https://github.com/asyncapi/avro-schema-parser/blob/78ea849d6b58962b9c72081d618e12755be3e2c0/index.js#L6
- https://github.com/asyncapi/openapi-schema-parser/blob/66ee0e21f62f9876a18ab8a4917d7582c9b6108d/index.js#L23
- https://github.com/asyncapi/raml-dt-schema-parser/blob/47e635dcafbc71e80a99ec7ee6e1bbfa968fd354/index.js#L21

and default parser should do it too -> https://github.com/asyncapi/parser-js/blob/master/lib/asyncapiSchemaFormatParser.js. And it would just have to be clearly described in the [readme](https://github.com/asyncapi/parser-js/blob/master/README.md#custom-message-parsers) that every custom parser must add those fields to the document

Thoughts?

c-pius:
Like the idea. Also similar to the x parser original trait approach then. Updated the PR already. Does that reflect what you meant? 

> And it would just have to be clearly described in the readme that every custom parser must add those fields to the document

Not sure how to handle this one though? From reading the existing doc I would have assumed to put it in the comment for the `getMimeTypes()` func but not sure if this would count as "clearly described" 😄

derberg:
@c-pius I think adding a comment to `parse` function in 1st point, between function signature and jsdoc should be sufficient, just a simple explanation of what extra properties must be added to each message by the parser

lib/parser.js
```diff
@@ -190,6 +195,7 @@ async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileForm
   });
 
   msg.schemaFormat = DEFAULT_SCHEMA_FORMAT;
+  msg[String(xParserOriginalSchemaFormat)] = schemaFormat;
```
derberg, line 198: code from https://github.com/asyncapi/parser-js/pull/214/files#diff-2f45270c7389933f71b380538ff52e106ee5cb66a0b573d9967467dfdbc11c02R187 runs a message parser for a given mime type, so if in your AsyncAPI file you have a message that has schema provided in avro, then this code runs -> https://github.com/asyncapi/parser-js/pull/214/files#diff-2f45270c7389933f71b380538ff52e106ee5cb66a0b573d9967467dfdbc11c02R187

- 1st the avro parser puts avro mime type under `x-parser-original-schema-format` 
- 2nd you override it with default schema format, asyncapi one, so we loose information about the original schema format

This is why instead of putting this line here, you need to have it in this parser of default asyncapi schema -> https://github.com/asyncapi/parser-js/blob/master/lib/asyncapiSchemaFormatParser.js

Add support for https://github.com/asyncapi/avro-schema-parser/blob/master/index.js#L7 too and put info about it in the readme as well please 🙏🏼

derberg:
I also wonder if this knowledge about concerns to performance, and the risk of invoking validation twice is not going to be just our secret here in PR. Not sure though if there is a possible test case that could be added to check this, like check if our default parser was invoked only once? 🤔

c-pius:
@derberg okay, understood. About to update it accordingly. I am just wondering, why is in this [line](https://github.com/asyncapi/parser-js/blob/b32899ea3dd6d8608fca32d29a42c3d459690fa3/lib/parser.js#L192) the `DEFAULT_SCHEMA` applied and not the `msg.schemaFormat || DEFAULT_SCHEMA_FORMAT` that is determined before?

> Add support for https://github.com/asyncapi/avro-schema-parser/blob/master/index.js#L7 too and put info about it in the readme as well please 🙏🏼

I tried to do that and noticed that the `anonymous-schema-<id>` parts are then added to the `x-parser-original-payload´ objects as well? I think the logic is somewhere [here](https://github.com/asyncapi/parser-js/blob/b32899ea3dd6d8608fca32d29a42c3d459690fa3/lib/models/asyncapi.js#L87). Is this okay or do we need to put a check that it should not be done for `x-parser-original-payload` objects?

c-pius:
> I also wonder if this knowledge about concerns to performance, and the risk of invoking validation twice is not going to be just our secret here in PR. Not sure though if there is a possible test case that could be added to check this, like check if our default parser was invoked only once? 🤔

I think I would have now added the `x-parser-original-schema-format` in the default/custom parser and have the check for opting out if already there in the [validateAndConvertMessage](https://github.com/asyncapi/parser-js/blob/b32899ea3dd6d8608fca32d29a42c3d459690fa3/lib/parser.js#L179). Then the custom parser wouldn't need to worry about it because we don't invoke it twice.

derberg:
> @derberg okay, understood. About to update it accordingly. I am just wondering, why is in this line the DEFAULT_SCHEMA applied and not the msg.schemaFormat || DEFAULT_SCHEMA_FORMAT that is determined before?

it is because the default parser and the custom ones (avro, raml data type or openapi) that are invoked before that line already did the conversion from the other schemas to the json schema. Original schema format is stored here `x-parser-original-schema-format` and we need to override property that holds information about schema format with the format after conversion. Does it make sense?

> I tried to do that and noticed that the anonymous-schema-<id> parts are then added to the x-parser-original-payload objects as well? I think the logic is somewhere here. Is this okay or do we need to put a check that it should not be done for x-parser-original-payload objects?

Strange, as the code that assigns anonymous id is invoked here https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L192, so before the original payload go to `x-parser-original-payload` 🤔 are you sure about it?

> I think I would have now added the x-parser-original-schema-format in the default/custom parser and have the check for opting out if already there in the validateAndConvertMessage. Then the custom parser wouldn't need to worry about it because we don't invoke it twice.

makes sense 👍🏼

c-pius:
> Strange, as the code that assigns anonymous id is invoked here https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L192, so before the original payload go to x-parser-original-payload 🤔 are you sure about it?

Did you mix up the line you wanted to provide? :) I assumed the anonymous id creation is triggered here as part of the AsyncAPI constructor ([assignIdToAnonymousSchemas()](https://github.com/asyncapi/parser-js/blob/b32899ea3dd6d8608fca32d29a42c3d459690fa3/lib/models/asyncapi.js#L87))? Therefore, after customDocumentOperations() was invoked? Maybe I am mistaken? https://github.com/asyncapi/parser-js/blob/b32899ea3dd6d8608fca32d29a42c3d459690fa3/lib/parser.js#L113

This is what I actually get:
Screenshot 2020-12-21 at 13 20 04

c-pius:
> it is because the default parser and the custom ones (avro, raml data type or openapi) that are invoked before that line already did the conversion from the other schemas to the json schema. Original schema format is stored here x-parser-original-schema-format and we need to override property that holds information about schema format with the format after conversion. Does it make sense?

I guess yes :) Didn't think about the schema being converted

derberg:
@c-pius oh man, I should not do any public activities on Mondays 😄 You are of course right, I pasted the wrong link, I meant exactly what you wrote 🤦🏼 

Anyway, good you showed your code. The problem that you are seeing there, that the original object also has `x-parser` extensions is because what you do is not a value assignment but a reference assignment. So whatever happens with `payload` will be visible in `x-parser-original-payload` because `x-parser-original-payload` keeps reference to `payload`.

Look at this code:
```
  message['x-parser-original-payload'] = payload;
  payload.test = 'my test value that will be visible also under oryginal payload';
  console.log('original: ', JSON.stringify(message['x-parser-original-payload']));
  console.log('payload: ', JSON.stringify(payload));
```

console.log is the same:
```
original:  {"type":"object","properties":{"name":{"type":"string"},"test":{"type":"object","properties":{"testing":{"type":"string"}}}},"test":"my test value that will be visible also under oryginal payload"}
payload:  {"type":"object","properties":{"name":{"type":"string"},"test":{"type":"object","properties":{"testing":{"type":"string"}}}},"test":"my test value that will be visible also under oryginal payload"}
```

c-pius:
that one is on me actually, obviously a deep / shallow copy issue 😅  thanks for the hint!

fixed it now and updated the tests. but not entirely sure about this one though:

>  // considered save here because default parser handles JSON Schema payloads
>  message[String(xParserOriginalPayload)] = JSON.parse(JSON.stringify(payload));

I assume it would be okay since payload is a JSON Schema definition in the default parser. Or would you have a preference for how to do the deep copy, at least I couldn't find a place where something similar is done already? I guess other options would be to implement in manually as helper function or use some tool like lodash.

lib/asyncapiSchemaFormatParser.js
```diff
@@ -11,10 +14,16 @@ module.exports = {
...
-async function parse({ message, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload }) {
+async function parse({ schemaFormat, message, defaultSchemaFormat, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload }) {
+  // setting original schema format required before returning because of no payload
+  message[String(xParserOriginalSchemaFormat)] = schemaFormat || defaultSchemaFormat;
+  
   const payload = message.payload;
   if (!payload) return;
 
+  // considered save here because default parser handles JSON Schema payloads
+  message[String(xParserOriginalPayload)] = JSON.parse(JSON.stringify(payload));
```
derberg, line 25: what about 👇🏼 

```suggestion
  message[String(xParserOriginalPayload)] = Object.assign({}, payload);
```

derberg:
@c-pius I added a suggestion. Lodash is always an option but for parser let us try to keep it as slim as possible

lib/asyncapiSchemaFormatParser.js
c-pius: Object.assign creates a new object but the copied properties are still shallow copies. so the payload object itself will not get a `x-parser-schema-id` applied, but for instance a "name" property within the payload will get it applied.

derberg: you are right.. 🤔 
we cannot leave `JSON.stringify` as it won't work with circular references in schemas

🤔

c-pius: Unfortunately I am not very deep into the node world (I guess you have noticed already 😄 ). So I would need to let you make the call on what to do best.

Considering that [lodash.clonedeep](https://www.npmjs.com/package/lodash.clonedeep) would be available as individual dependency and looking [behind the curtains](https://github.com/lodash/lodash/blob/2f79053d7bc7c9c9561a30dda202b3dcd2b72b90/.internal/baseClone.js#L157) on what would need to be done in an own helper function, I am not sure if manually implementing it is a good idea. On the other hand, I also cannot judge the impact of adding a new dependency to this lib.

derberg: lodash is just 3.3kb gziped which is fine, I'm more concerned about performance again. We add something that is not that much needed but will add performance issues to large files again 🤔 

I have doubts if this is the way to go. If we should not fall back to the initial idea about adding an additional field that we can use to check if the message was processed or not. Sorry, I just did not take into consideration this issue with copying objects

c-pius: no worries, got you. and yeah, since the payload is not actually transformed, the `x-parser-original-payload` would always include the same structure except the added `x-parser` attributes. not sure if there is any use for this except consistency with custom parsers regarding the `x-parser-original-payload` being there?

derberg: yeah, the use case if for those that have different formats of schemas, like avro for example, so after parsing the document they can still access the content before conversion in case they want to include it in docs or prefer to read original schema, for example in runtime validations. 

so, again back to `what could be the name of the extension that holds information if the message was already processed or not` 😄

c-pius:
> so, again back to what could be the name of the extension that holds information if the message was already processed or not 😄

the hardest question of all 😄   as far as I can tell the open options are:

- we still go for `x-parser-original-schema-format` 
  - pro: reuse of existing property
  - con: we rely on the custom parsers to actually set the property to avoid applying the parsing twice
- new property which is set and checked in `validateAndConvertMessage()`
  - pro: independent from parsers, concise within one function
  - con: additional `x-parser-...` property (e.g. `x-parser-message-validated`, `x-parser-message-parsed`, `x-parser-message-converted`)

I am still wondering actually if we could also set the `x-parser-original-schema-format` ourselves in `validateAndConvertMessage()`? Isn't the used parser actually selected by the original schema format || default here:
https://github.com/asyncapi/parser-js/blob/b32899ea3dd6d8608fca32d29a42c3d459690fa3/lib/parser.js#L182

If so, we would be independent from the custom parsers and wouldn't need an additional property

derberg:
@c-pius hey, sorry mate but I and others had a longer break here. 2020 was crazy busy and we needed a few days without asyncapi.

Let us have a new property -> `x-parser-message-parsed`

c-pius:
no worries, I also enjoyed some time off 😉 and happy new year!

I reverted the `x-parser-original-schema` and `x-parser-original-payload` content and added the check using `x-parser-message-parsed` flag in the `validateAndConvertMessage()`  👍🏻

lib/parser.js
```diff
@@ -253,4 +259,30 @@ async function customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialF
     }));
   });
   await Promise.all(promisesArray);
-}
\ No newline at end of file
+}
+
+/**
+ * Triggers additional operations on the AsyncAPI components/messages like traits application, validation and conversion
```
derberg, line 265:
```suggestion
 * Triggers additional operations on the AsyncAPI messages located in the components section of the document. It triggers operations like traits application, validation and conversion
```

lib/parser.js
```diff
@@ -253,4 +259,30 @@ async function customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+ *
+ * @private
+ * 
+ * @param  {Object} parsedJSON parsed AsyncAPI document
+ * @param  {String} asyncapiYAMLorJSON AsyncAPI document in string
+ * @param  {String} initialFormat information of the document was originally JSON or YAML
+ * @param  {Object} options Configuration options.
+ */
+async function customMessagesOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options) {
```
derberg, line 274:
```suggestion
async function customComponentsMsgOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options) {
```

lib/parser.js
```diff
@@ -169,14 +170,18 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
   validateServerSecurity(parsedJSON, asyncapiYAMLorJSON, initialFormat, SPECIAL_SECURITY_TYPES);
 
   if (!parsedJSON.channels) return;
-  
+
   validateChannels(parsedJSON, asyncapiYAMLorJSON, initialFormat);
   validateOperationId(parsedJSON, asyncapiYAMLorJSON, initialFormat, OPERATIONS);
 
+  await customMessagesOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options);
```
derberg, line 177:
```suggestion
  await customComponentsMsgOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options);
```

derberg (changes requested):
lgtm, I just made suggestions to be clear about new functionality, that it is only for messages from components.

@magicmatatjahu anything from your side? can you approve?

magicmatatjahu:
@derberg 8 new lines (without tests etc) with 42 comments? 😄 I'll check, give me some time to read whole thread.

derberg:
@magicmatatjahu no worries mate, I don't think you have to read it through, we had a long discussion here as we were just looking for the best possible solution on how to run operations on messages just once) in case the message is in components and also ref-ed from the channel. But we are back to the beginning I would say, just introduced a new extension `x-parser-message-parsed` and that is it, so I would say just look at the code to make sure I did not overlook something because of our back and forth

magicmatatjahu (dismissed):
LGTM with new label :)

lib/parser.js
```diff
@@ -169,14 +170,18 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
   await customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options);
 }
 
 async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload) {
+  //check if the message has been parsed before
+  if (xParserMessageParsed in msg && msg[String(xParserMessageParsed)] === true) return;
```
magicmatatjahu, line 183: It's only a question. `(xParserMessageParsed in msg) === true` will work?

derberg (approved):

magicmatatjahu (approved):

derberg:
Finally released. Thanks @c-pius for this PR and patience in the discussion 😄

c-pius:
awesome 🥳 

thank you for guiding me and for the great work on this project in general 😄

PR #216, opened 2020-12-11
test: improve feedback loop from browser tests

derberg:
**Description**

- console log everything that shows up in the browser console during browser tests
- cleanup of browser test must be done before and after test. If we do only after, the problem is that when the test fails, cleanup doesn't run

**Related issue(s)**
See also https://github.com/asyncapi/parser-js/pull/214

### DIFF
--- a/test/browser_test.js
+++ b/test/browser_test.js
@@ -10,6 +10,10 @@ describe('Check Parser in the browser', function() {
     //use this in case you want to troubleshoot in a real chrome window => browser = await puppeteer.launch({headless: false});
     browser = await puppeteer.launch();
     page = await browser.newPage();
+    page.on('console', msg => {
+      for (let i = 0; i < msg.args().length; ++i)
+        console.error(`Browser console error ${i}: ${JSON.stringify(msg.args()[i]._remoteObject, null, 2)}`);
+    });
     await page.goto('http://localhost:8080', { waitUntil: 'networkidle0' });
   });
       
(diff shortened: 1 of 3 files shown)


test/browser_test.js
```diff
@@ -10,6 +10,10 @@ describe('Check Parser in the browser', function() {
     //use this in case you want to troubleshoot in a real chrome window => browser = await puppeteer.launch({headless: false});
     browser = await puppeteer.launch();
     page = await browser.newPage();
+    page.on('console', msg => {
```
magicmatatjahu, line 13: As I understand that `console` event is for all `console`, am I right? If yes, then has hardcoded logic for only errors is not good. I mean someone can write normal `console.log` and then it will be treated as error. We should check if message has error type.

derberg: where do you see a hardcode here?
it worked well before I changed to `console.error` in `index.html`. I switched to `console.error` only because I found it more accurate in this case.

derberg: @magicmatatjahu ping 😉

magicmatatjahu: Sorry 😄 I meaned that `page.on('console' ...)` will be triggered for all usage `console` like `console.log`, `console.error` etc. and then you have

```js
console.error(`Browser console error ${i}: ${JSON.stringify(msg.args()[i]._remoteObject, null, 2)}`);
```

so any non-error log will be displayed as error? Should we first check that log is `error` type?

derberg: I think it should log everything, even console log, as it is helpful when debugging or extending test, no need to use headless mode and other hacks to see how test behaves.

I just made the log info more generic.

I also tested and `console.log('dupa')` puts 👇🏼 in logs
```
{
  "type": "string",
  "value": "dupa"
}
```

magicmatatjahu (approved):

PR #223, opened 2021-01-07
feat!: improve names for anonymous schemas and messages

### LINKED ISSUE #176: Anonymous schemas should have meaningful ids
#### Reason/Context
We at Eurisco are not a big fan of the anonymous schema ids. They don't give any context to what anonymous schema or properties they are apart of which makes it hard to track down where it belongs. Instead we replace any anonymous schema ids to be the path to the object, see example solution for more details.

#### Description
The function giving schemas and messages anonymous ids need to be changed to another algorithm. 

This will be breaking change in the regard anonymous schema ids are no longer present and people have to update their implementations unless they manually defined all schemas.

#### Example solution
My colleague developed a custom hook for changing the anonymous schema ids recursively by saving the path to that schema. 
```js
module.exports = {
    'generate:before': async generator => {
        preprocess(generator, generator.asyncapi)
    }
};

preprocess = (generator, obj, lastKey = "", lastSchemaID = "") => {
    // Fix anonymous x-parser-schema-id
    const schemaId = obj["x-parser-schema-id"]
    if (schemaId !== undefined) {
        if (schemaId.startsWith("<anonymous-schema-")) {
            const newSchemaId = lowerFirst(lastSchemaID + upperFirst(lastKey))
            obj["x-parser-schema-id"] = newSchemaId
            lastSchemaID = newSchemaId
        } else {
            lastSchemaID = schemaId
        }
    }
    const keys = Object.keys(obj)
    for (let i = 0; i < keys.length; i++) { 
        let key = keys[i]
        const value = obj[key];
        if (typeof value === 'object' && value != undefined) {
            let newKey = key
            if (key == "payload" && obj["name"] !== undefined && lastKey === "message") {
                lastSchemaID = obj["name"]
            } else if (key == "items" && obj["type"] === "array") {
                newKey = ""
            }
            if(!!obj['x-parser-circular-props']){
                continue;
            }
            preprocess(generator, value, newKey, lastSchemaID);
        }
    }
    return obj
}
upperFirst = (str) => {
    return str.charAt(0).toUpperCase() + str.slice(1);
};
lowerFirst = (str) => {
    return str.charAt(0).toLowerCase() + str.slice(1);
};
```

together with the hook and the  following AsyncAPI document
```yaml
asyncapi: '2.0.0'
info:
  title: Questionaire
  version: '1.0.0'
defaultContentType: application/json
channels:
  questionaire/get:
    description: "Get the newest version of the questionaires"
    publish:
      message:
        name: questionaire
        payload:
            type: object 
            x-parser-schema-id: questionaire
            properties:
                steps: 
                    type: object
                    additionalProperties: 
                        type: string
```

It would normally give the inner property `steps` of `questionaire` the id `<anonymous-schema-x>` but with the hook it will get `questionaireSteps`.

jonaslagoni:
WIP: will change as I go depending on the output it provide.

## Description
This PR gives anonymous schemas and messages meaningful ids instead of `<anonymous-message-x>` as well as giving parameters and header root schemas meaningful uid's. We switch over to use a path naming approach so one knows where the anonymous schema or name is located in the document.

### Rules of naming
These are the rules for the uid generation:

- Anonymous messages are given the name `<channel-name>Message` according to what channel it belong to.
- Anonymous parameter schemas are given the name `<channel-name><parameter-name>Parameter`
- Anonymous message header schemas are given the name `<message-name><header-name>Header`
- Anonymous message payload schemas are given the name `<message-name>Payload`

Regarding the naming of nested schemas here is the procedure, the name will cumulative based on how deep the nesting goes, here is the rules for cumulating the name:
- The parent schema's id will the be starting point `<parent-schema-name>`
- If the current schema is a property of an object the property name is added `<property-name>`
- The type of the schema is then added i.e. if of type object `Object` is appended to the name

So the full name of the anonymous schema will be`<parent-schema-name><property-name><schema-type>Schema`

#### The special cases
- Two root schemas are referenced between headers, parameters and message payloads
    - If a schema already have a uid it is never changed again, the order of which uid is used is the following: payload, header, parameter

**What it introduces**
- It changes the behavior of the naming of anonymous schemas and messages
- Fixed some functions had the wrong type for callback for JSdoc
- Changes the naming of parameters and components

**Related issue(s)**
fixes #176


jonaslagoni:
I am gonna close this PR and introduce it gradually as much as possible.

PR #225, opened 2021-01-15
chore: Refactored code location for iterators

jonaslagoni:
**Description**
This PR moves iterators and constants to move towards a model which only contains the model code and not excess functionality to reduce complexity. This is a pre-step for me to move assignment of anonymous ids to schemas in #223


jonaslagoni:
Let me know if we should move the files into another directory or if its fine where it is

jonaslagoni:
Also, do we need new unit tests for these functions? 🤔

lib/constants.js
```diff
@@ -0,0 +1,9 @@
+const xParserMessageName = 'x-parser-message-name';
+const xParserSchemaId = 'x-parser-schema-id';
+const xParserCircle = 'x-parser-circular';
+const xParserCircleProps = 'x-parser-circular-props';
+
+module.exports.xParserMessageName = xParserMessageName;
```
magicmatatjahu, line 6: Maybe we simplify this?

```js
module.exports = {
  xParserMessageName: 'x-parser-message-name',
  ...
}
```

lib/iterators.js
```diff
@@ -0,0 +1,228 @@
...
+      schema.items().forEach(arraySchema => {
+        traverseSchema(arraySchema, callback, null, schemaTypesToIterate);
+      });
+    } else {
+      traverseSchema(schema.items(), callback, null, schemaTypesToIterate);
+    }
+  }
+}
+
+module.exports.SchemaIteratorCallbackType = SchemaIteratorCallbackType;
```
magicmatatjahu, line 221: Like in above:

```ts
module.exports = {
  SchemaIteratorCallbackType ,
  SchemaTypesToIterate ,
  ...
}
```

lib/constants.js
jonaslagoni: Yea thats better 😄

magicmatatjahu (approved):

PR #228, opened 2021-01-18
fix: Unused component messages are never iterated

### LINKED ISSUE #219: anonymous schema ids for standalone messages
#### Describe the bug
Related to #214 

When parsing standalone messages of the `components.messages` object (no reference from channels), the `x-parser` properties are not added to the messages in the output schema.

See also https://github.com/asyncapi/parser-js/pull/214#issuecomment-742725894

#### How to Reproduce
Check the tests for `test/good/asyncapi-no-channels.yml` or `test/good/asyncapi-messages-channels.yml`. 

#### Expected behavior
For `test/good/asyncapi-no-channels.yml` it should return expected instead of actual

[actual.txt](https://github.com/asyncapi/parser-js/files/5684932/actual.txt)
[expected.txt](https://github.com/asyncapi/parser-js/files/5684933/expected.txt)

jonaslagoni:
**Description**
This PR fixes the iterator not iterating component messages.

**Related issue(s)**
fixes #219


lib/iterators.js
```diff
@@ -125,6 +125,11 @@ function traverseAsyncApiDocument(doc, callback, schemaTypesToIterate) {
     Object.values(doc.components().schemas()).forEach(s => {
       traverseSchema(s, callback, null, schemaTypesToIterate);
     });
+    if (doc.components().hasMessages()) {
```
magicmatatjahu, line 128: It's a strange 😄 For schemas we didn't check that schemas exist in components, but for messages we'll have this statement 😄

jonaslagoni: Yea, I was actually getting failed tests without this statement 😕

jonaslagoni: Moved some checks around so the function traverseMessage now checks for null messages instead :)

magicmatatjahu (approved):
LGTM

PR #247, opened 2021-02-04
fix: testing on mac and windows

derberg:
**Description**

- this is fix PR as another release must be triggered to release asyncapi-node bump that was blocked by failing tests
- bump puppeteer to fix browser tests for mac on ci
- more logs in browser tests
- increase timeouts to make sure it works on ci
- adjust scripts to use quotes to make Bill Gates and Microsoft happy
- disable linebreaks rule that causes headaches on windows
- remove offset where not needed in tests (otherwise windows fails)
- add try catch to browser tests


jonaslagoni (approved):
👍

PR #251, opened 2021-02-11
test: extract checkErrorWrapper and use it in other tests

aeworxet:
This PR resolves issue #140.

- moved `checkErrorWrapper` from `parse_test.js` to separate file `testsUtils.js`
- refactored `parse_test.js` to use `checkErrorWrapper` from `testsUtils.js`
- refactored `asyncapiSchemaFormatParser_test.js` to use `checkErrorWrapper` from `testsUtils.js`
- refactored `customValidators_test.js` to use `checkErrorWrapper` from `testsUtils.js`
- removed offsets where input is a string variable and not a file
- added execution of *nix commands through `shx` so they also work on Windows without any change

(`shx` wraps native Unix commands as `shx cp \"dist/bundle.js\"` to avoid errors like this on Windows platforms)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "get:name": "echo $npm_package_name",
     "lint": "eslint --max-warnings 0 --config \".eslintrc\" \".\"",
     "test:lib": "nyc --reporter=html --reporter=text mocha --exclude \"test/browser_test.js\" --recursive",
-    "test:browser": "npm run test:browser:cleanup && npm run bundle && cp \"dist/bundle.js\" \"test/sample_browser/\"  && start-server-and-test \"http-server test/sample_browser --cors -s\" 8080 \"mocha --timeout 20000 test/browser_test.js\" && npm run test:browser:cleanup",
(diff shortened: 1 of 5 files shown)


derberg (commented):
Awesome work @aeworxet! thanks a lot mate!

I only have one question. In majority of tests I see `parsedJSON: JSON.stringify(parsedInput)` and then in the wrapper we do `JSON.parse(parsedJSON)`, Just curious if you tried to not stringify, so later you do not have to parse to json again? 
I see that only `parsedJSON: invalidAsyncapi,` we do not stringify (and this is why later we need to parse to JSON. What if we turn it around, we just Parse to JSON here once and then we do not need to stringify and parse in wrapper all the others. Does that make sense?

aeworxet:
Variables, generated by `fs.readFileSync` are transferred unchanged only to `parser.parse()` and it already doesn't complain on additional `\n`s in the var.

`JSON.parse` is used only for variables which are real JSONs in string form, so they cannot be used without `JSON.parse` anyway.


The most I could do, was to:
- remove `JSON.stringify` from `customValidators_test.js`,
- add `JSON.parse` to `asyncapiSchemaFormatParser_test.js` and `parse_test.js`,
- clean out `JSON.parse` from `testsUtils.js`.

derberg (approved):
🥳

derberg:
@all-contributors please add @aeworxet  for test

Issue #252, opened 2021-02-15
Measure performance of the parser


derberg:
#### Reason/Context
The parser should be fast and lightweight. We should always make sure about it on PR level:
- measure bundle size to not make browsers cry when fetching
- measure parsing performance
- always check those to make sure a given feature doesn't affect the performance of the parser

#### Description
Please try answering few of those questions
- have a github action that measures both
- discuss and set limits on when PR should be blocked
- publish performance tests results in readme
- get  from community a super large asyncapi document that we can use for testing

magicmatatjahu:
We can use https://ajv.js.org/standalone.html to generate validate function for our schema to improve performance. At the moment we "compile" schema in each `parse` function execution ([line](https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L94)). I don't know how we can speed parsing, but I think that it should be faster :)

derberg:
still needed

PR #253, opened 2021-02-16
fix: clone message payload into x-parser-original-payload

### LINKED ISSUE #245: `x-parser-original-payload` isn't the original payload with asyncapi schema parser
Using something like `channel.subscribe().messages('myMessage').originalPayload();` will return a schema that has all of the `x-parser-schema-id` extensions baked in. This is not part of the original payload. This isn't an issue with the Avro schema parser as they remove the reference to `message.payload` for the original. 

We've managed to create a custom parser that gets around the issue:

```
import {cloneDeep} from 'lodash';

export default {
  parse: async ({ message, defaultSchemaFormat }) => {
    message['x-parser-original-schema-format'] = message.schemaFormat || defaultSchemaFormat;
    // keep copy of original payload without x-parser-schema-id extensions
    message['x-parser-original-payload'] = cloneDeep(message.payload);
    delete message.schemaFormat;
  },
  getMimeTypes: () => {
    return [
      'application/vnd.aai.asyncapi;version=2.0.0',
      'application/vnd.aai.asyncapi+json;version=2.0.0',
      'application/vnd.aai.asyncapi+yaml;version=2.0.0'
    ]
  }
};
```
cloneDeep was needed as it did set the original payload correctly without it, but then `message.payload` changed after the parser had been called so the original also changed with it.

Is this supposed to be how it works? We would like to be able to create a visualisation of the schema without all of the parser extensions included and this custom parser feels a bit sketchy.

JQrdan:
**Description**

This PR modifies the asyncApi schema parser to keep a copy of the original payload and schema in `x-parser-original-payload` and `x-parser-original-schema-format`, respectively. This allows consumers of the parser to use the original payload for 'user facing' features, such as displaying a visualisation of the payload without any of the parser-extras that get added in. This also brings the behaviour inline with the Avro schema parser as that also keeps a copy.

**Related issue(s)**
Fixes #245


derberg:
@JQrdan that was quick! 🚀  please resolve conflicts, my recommendation is to just not push changes done to the bundle.js because we anyway generate them fresh during the release

Please also have a look at the issues found by the sonarcloud

JQrdan:
> Please also have a look at the issues found by the sonarcloud

@derberg Apologies, but am I missing something with the issues found by sonarcloud? I see 4 but they seem to be from other stuff that's unrelated to this pr?

derberg:
you were too fast force pushing and I cannot answer your question as I cannot check that out 😄 
also for the future 🙏🏼 avoided force-push as it is hard to do reviews as it is impossible to easily check what was added. Do not worry about the history as we anyway squash and merge always

JQrdan:
> you were too fast force pushing and I cannot answer your question as I cannot check that out 😄
> also for the future 🙏🏼 avoided force-push as it is hard to do reviews as it is impossible to easily check what was added. Do not worry about the history as we anyway squash and merge always

Cool, sorry! It's a habit from other projects 😄  Especially when it comes to removing unwanted files from PRs. Although the force-push only included removing the bundle so I'm confused as to why the code smells are now 0 😂

lib/asyncapiSchemaFormatParser.js
```diff
@@ -11,10 +12,13 @@ module.exports = {
 /**
  * @private
  */
-async function parse({ message, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload }) {
+async function parse({ message, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload, defaultSchemaFormat }) {
   const payload = message.payload;
   if (!payload) return;
 
+  message['x-parser-original-schema-format'] = message.schemaFormat || defaultSchemaFormat;
+  message['x-parser-original-payload'] = cloneDeep(message.payload);
```
fmvilas, line 20: 🤔 I have the feeling that this deep cloning is going to cause performance issues, especially on big payload definitions. It would be great to see how much performance degrades after this PR.

test/asyncapiSchemaFormatParser_test.js
```diff
@@ -98,4 +109,38 @@ describe('asyncapiSchemaFormatParser', function() {
 
     expect(async () => await parser.parse(parsedInput)).to.not.throw();
   });
+
+  it('should deep clone schema into x-parser-original-payload', async function() {
```
fmvilas, line 113: This test says it "should deep clone schema" but we're actually not checking if it's a deep clone, or am I missing something? If you remove the `deepClone` call above and just leave it like this:

```js
message['x-parser-original-payload'] = message.payload;
```

I think this test still passes.

fmvilas (changes requested):

test/asyncapiSchemaFormatParser_test.js
JQrdan: It verifies that it's a deepclone as the parser would have modified the original payload. So if it's not a deep clone, `message['x-parser-original-payload']` fails because the payload gets modified, and so does the 'original' because they just share a reference. If I remove the deep clone, the test fails because the extra parser pieces have now appeared.

```
 AssertionError: expected { Object (type, properties, ...) } to deeply equal { Object (type, properties) }
      + expected - actual

           "id": {
             "description": "Id of the streetlight."
             "minimum": 0
             "type": "integer"
      -      "x-parser-schema-id": "<anonymous-schema-2>"
           }
           "lumens": {
             "description": "Light intensity measured in lumens."
             "minimum": 0
             "type": "integer"
      -      "x-parser-schema-id": "<anonymous-schema-3>"
           }
           "sentAt": {
             "description": "Date and time when the message was sent."
             "format": "date-time"
             "type": "string"
      -      "x-parser-schema-id": "<anonymous-schema-4>"
           }
         }
         "type": "object"
      -  "x-parser-schema-id": "<anonymous-schema-1>"
       }
```

lib/asyncapiSchemaFormatParser.js
JQrdan: Yep! This was raised as part of the issue that got us here 😅  If it does start to cause issues, it would still be nice to provide a way to optionally do a deep clone? That way users can decide if a schema is too big for them to care about keeping the original.

derberg: Yeah 😅 
We discussed it and in the end, I was like, yeah, we do not do any performance tests and first, we need to do it [here](https://github.com/asyncapi/parser-js/issues/252) I think, and was thinking this should not block this change 🤷🏼‍♂️

fmvilas: Not fully convinced but I don't want to be a blocker here. It's true that we don't have performance tests but adding things on purpose knowing it's going to heavily affect performance doesn't make me happy either 😅  Anyway, let's just move forward and we'll work on performance improvements later. I also don't want to unnecessarily add complexity to the API to optionally clone the payload. Thanks for clarifying!

test/asyncapiSchemaFormatParser_test.js
fmvilas: I see. Thanks for explaining.

What about adding an extra test saying that whatever is on `x-parser-original-payload` !== `message.payload`? If we compare an object with itself using `to.deep.equal`, it's going to pass but it's not a clone.

test/asyncapiSchemaFormatParser_test.js
```diff
@@ -98,4 +109,38 @@ describe('asyncapiSchemaFormatParser', function() {
...
+    };
+    const expectedOriginalSchemaFormat = 'application/vnd.aai.asyncapi;version=2.0.0';
+    
+    const parsedOutput = await parser.parse(asyncapi);
+    const parsedMessage = parsedOutput.channel('light/measured').publish().message().json();
+
+    expect(parsedMessage.payload).to.have.property('x-parser-schema-id');
+    expect(parsedMessage).to.have.property('x-parser-original-payload');
+    expect(parsedMessage).to.have.property('x-parser-original-schema-format');
+    expect(parsedMessage['x-parser-original-payload']).to.deep.equal(expectedOriginalPayload);
```
fmvilas, line 143: This is the extra test I meant above:

```suggestion
    expect(parsedMessage['x-parser-original-payload']).to.deep.equal(expectedOriginalPayload);
    expect(parsedMessage['x-parser-original-payload']).to.not.equal(parsedMessage.payload);
```

We just want to make sure that `x-parser-original-payload` and `message.payload` are not both the same object.

lib/asyncapiSchemaFormatParser.js
derberg: @fmvilas 
I'm not 100% sure either, I just tested it manually on local with `console.time` with current tests (yes they are small files, I know) and such deep close was not really adding too much.

```
  console.time('time to deep clone');
  message['x-parser-original-payload'] = cloneDeep(message.payload);
  console.timeEnd('time to deep clone');
```

This is what I was getting:
```
time to deep clone: 0.064ms
    ✓ should parse YAML
time to deep clone: 0.052ms
    ✓ should parse YAML correctly when no components object
time to deep clone: 0.062ms
time to deep clone: 0.043ms
    ✓ should parse 2 AsyncAPI specs in Promise.all() and not fail with resolving references (54ms)
time to deep clone: 0.036ms
    ✓ should apply traits to messages even with empty channels object
time to deep clone: 0.039ms
    ✓ should apply traits to messages used and not used in a channel
time to deep clone: 0.035ms
    ✓ Should include schemas after circular property
time to deep clone: 0.071ms
time to deep clone: 0.024ms
time to deep clone: 0.021ms
time to deep clone: 0.046ms
    ✓ should properly mark circular references
```

The schema that took the longest was https://github.com/asyncapi/parser-js/blob/master/test/good/circular-refs.yaml#L71 which is pretty complex I would say, and, still if we have 1000 schemas like that, it would only add 71ms to parsing.

This is why I was like: "yeah, not much harm will be done I guess, let's try"

derberg:
@JQrdan hey man, sorry for such a huge delay, it was completely lost among other things I had to work on. I'll have a final look and merge when all is fine

derberg (approved):

derberg:
@all-contributors please add @JQrdan for test, code

PR #258, opened 2021-03-13
docs: add small section explaining the API

jonaslagoni:
**Description**
After question in Slack, it became clear that just linking to the API is not quite enough. So I have added a small section explaining the general API approach.


smoya (approved):
LGTM 👍 🚀

PR #261, opened 2021-03-18
test: imporve test for json parser

aayushmau5:
**Description**

- Improve test for `json-parse.js` module.

**Related issue(s)**

None

### DIFF
--- a/test/json-parse_test.js
+++ b/test/json-parse_test.js
@@ -1,0 +1,76 @@
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+const jsonParseBetterErrors = require('../lib/json-parse');
+const { checkErrorWrapper } = require('./testsUtils');
+
+chai.use(chaiAsPromised);
+
+describe('jsonParseBetterErrors', function () {
+  it('should throw error if passed value is an array', async function () {
+    const expectedErrorObject = {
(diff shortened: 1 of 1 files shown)


derberg (commented):
Looks good to me for a merge.

Just want to draw your attention to the coverage report that indicated that the test covers just half of code. As far as I can see you focused on https://github.com/asyncapi/parser-js/blob/master/lib/json-parse.js#L18-L25 checking error when input is different than object. But what if the input is json but a broken one and we do evaluation like https://github.com/asyncapi/parser-js/blob/master/lib/json-parse.js#L10-L13 that "enrich" the structure of the error message?

Thoughts?

aayushmau5:
Yeah about that, I think the Invalid JSON test has already been covered at https://github.com/asyncapi/parser-js/blob/master/test/parse_test.js#L416-L440

What do you say, should I add test for those as well?

derberg:
@aayushmau5 you are right, this is then covered and from my point of view even in a better place. Please update your branch with latest master and I will in the meantime silent SonarCloud complaining about your empty arrow function

aayushmau5:
@derberg Done :)

derberg (approved):

derberg:
@aayushmau5 congrats of getting your first PR merged 🥳

PR #262, opened 2021-03-25
test: add test for parseFromUrl function

aayushmau5:
**Description**

- Add test for `parseFromUrl` function present in `parser.js` file.
- Wrap the exception thrown by `parseFromUrl` function in `parserError` object.
- Modified the `package.json` scripts to combine reports from multiple tests.
- Update README to add the Error Type.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -133,7 +133,13 @@ function parseFromUrl(url, fetchOptions, options) {
       .then(res => res.text())
       .then(doc => parse(doc, options))
       .then(result => resolve(result))
-      .catch(reject);
+      .catch(e => {
+        if (e instanceof ParserError) return reject(e);
+        return reject(new ParserError({
+          type: 'unexpected-error',
+          title: e.message,
+        }));
+      });
   });
 }
 
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -136,7 +136,7 @@ function parseFromUrl(url, fetchOptions, options) {
       .catch(e => {
         if (e instanceof ParserError) return reject(e);
         return reject(new ParserError({
-          type: 'unexpected-error',
+          type: 'fetch-url-error',
           title: e.message,
         }));
       });
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test": "test"
   },
   "scripts": {
-    "test": "npm run test:browser:cleanup && npm run test:lib && npm run test:browser",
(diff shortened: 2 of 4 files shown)


lib/parser.js
```diff
@@ -133,7 +133,13 @@ function parseFromUrl(url, fetchOptions, options) {
       .then(res => res.text())
       .then(doc => parse(doc, options))
       .then(result => resolve(result))
-      .catch(reject);
+      .catch(e => {
+        if (e instanceof ParserError) return reject(e);
+        return reject(new ParserError({
+          type: 'unexpected-error',
```
derberg, line 139: shouldn't we maybe introduce a custom type here, like maybe `url-fetching-error` or `fetching-error`?
if you agree, don't forget about updating the readme

package.json
```diff
@@ -17,7 +17,9 @@
     "get:version": "echo $npm_package_version",
     "get:name": "echo $npm_package_name",
     "lint": "eslint --max-warnings 0 --config \".eslintrc\" \".\"",
-    "test:lib": "nyc --reporter=html --reporter=text mocha --exclude \"test/browser_test.js\" --recursive",
+    "test:lib": "nyc --silent --no-clean mocha --exclude \"test/browser_test.js\" --exclude \"test/parseFromUrl_test.js\" --recursive",
+    "test:parseFromUrl":"nyc --silent --no-clean start-server-and-test \"http-server test/sample_browser --cors -s\" 8080 \"mocha test/parseFromUrl_test.js\"",
```
derberg, line 21: I like you are reusing the sample browser from another test

derberg (commented):
Well done mate. Statements coverate for parser.js jumped from 91 to 99, just one error is not yet tested 🥳 

Just one question from my side, have a look.

package.json
aayushmau5: Yup, all thanks to your for giving me pointers.

lib/parser.js
aayushmau5: Agree. `fetch-url-error` sounds good. Also, thanks for reminding to update the readme.

derberg: awesome, so go ahead with the change, in the test also validate not only message but type too please

aayushmau5:
@derberg Please check :)

derberg (approved):

derberg:
@all-contributors please add @aayushmau5 for test

PR #263, opened 2021-03-25
docs: intent API 

jonaslagoni:
**Description**
This PR suggests the new intent API for the parser.

It is implicit that all properties will have their own intents in the end but are not added for simplicity.

**Looking for feedback**

* Do you have any extra intents we should add (initially)?
    * Don't expect all intents to be included in the first iteration, you should "easily" see where other intents could be added. But remember this is just an initial version that shows the general approach. 
* Do you like the setup/dislike the setup? Why? What can be improved?
* Any intents that are completely off?
* Can you come up with better naming formats? 
* Other things?

**Next steps**

* To validate that this API can handle the [emulated breaking changes](https://github.com/asyncapi/shape-up-process/issues/93).
* Once validated, start mocking the methods to integrate the API immediately with the current tooling and validate the UX.

**Related issue(s)**
- https://github.com/asyncapi/shape-up-process/issues/90


jonaslagoni:
## Emulated breaking changes
We needed to test the intents out based on different changes from https://github.com/asyncapi/shape-up-process/issues/93

### 3.0.0-fake
Assumptions: 
- Only one operation type (publish/subscribe) per channel
Intents:
- **(GOOD)** `Channel.messagesSubscribes('<role>')` : Message[]
- **(GOOD)** `Channel.messagesPublishes('<role>')` : Message[]
- **(GOOD)** `Channel.summaryForPublishing('<role>')` : string
- **(GOOD)** `Channel.summaryForSubscribing('<role>')` : string

### 3.0.0-fake2
Because this makes it possible to define multiple schema formats per message, it is no longer a 1:1 (message:schemaFormat) but a 1:* and breaks the return values 
- **(QUESTIONABLE)** `Message.schemaFormat()` : string (this intent is not part of the initial suggestion, but added to support the breaking change example)
    - This could return multiple schema formats

### 3.0.0-fake3
Because this makes it possible to define multiple operations per channel, it is no longer 1:1 (channel:operation) but a 1:* and breaks the return values
- **(GOOD)** `Channel.messagesSubscribes('<role>')` : Message[]
- **(GOOD)** `Channel.messagesPublishes('<role>')` : Message[]
- **(QUESTIONABLE)** `Channel.summaryForPublishing('<role>')` : string
    - This could return multiple summaries.
- **(QUESTIONABLE)** `Channel.summaryForSubscribing('<role>')` : string
    - This could return multiple summaries.
- **(QUESTIONABLE)** `Channel.operationIdForPublishing('<role>')` : string 
    - This could return multiple ids.
- **(QUESTIONABLE)** `Channel.operationIdForSubscribing('<role>')` : string
    - This could return multiple ids.

### 3.0.0-fake4
Because this makes it possible to define multiple messages under a channel, and those messages can contain operations we no longer have a 1:1 (channel:operation) but a 1:* and breaks the return values
- **(GOOD)** `Channel.messagesSubscribes('<role>')` : Message[]
- **(GOOD)** `Channel.messagesPublishes('<role>')` : Message[]
- **(QUESTIONABLE)** `Channel.summaryForPublishing('<role>')` : string
    - This could return multiple summaries.
- **(QUESTIONABLE)** `Channel.summaryForSubscribing('<role>')` : string
    - This could return multiple summaries.
- **(QUESTIONABLE)** `Channel.operationIdForPublishing('<role>')` : string 
    - This could return multiple ids.
- **(QUESTIONABLE)** `Channel.operationIdForSubscribing('<role>')` : string
    - This could return multiple ids.
## Conclusion
As you can see we have a problem with the cardinality of return values in general..

The only apparent solution is to return an array of values, always, to be safe. But this is affecting the toolings using the parser making it more difficult...

smoya:
Taking this as an example of what would happen in case we return an array of strings:

> (QUESTIONABLE) Channel.summaryForPublishing('<role>') : string
This could return multiple summaries.
> (QUESTIONABLE) Channel.operationIdForPublishing('<role>') : string
This could return multiple ids.

The issue about returning an array of strings here is that the user won't be able to link the data it gets from each of those methods ever. 
For example, it won't be able to answer: "To what operation this `summary` belongs to?"

I think that, if we end up having such a change on our spec, then the `Operation` term will be stronger than what it is now and will become part of the entities (models in https://github.com/asyncapi/parser-js/pull/263/files#diff-b916c5a0628af4b4bdb634e7612d93408a30119addea9aab98fe5ac7c28adeafR13 list) we care of. 
That will mean we could introduce a new method for getting the **Operations** of a message or channel instead.

magicmatatjahu:
Awesome works guys! Sorry for late response. I read docs in PR and issue description/comments and I have problem with:

> It is implicit that all properties will have their own intents in the end but are not added for simplicity.

As I understand intents like `tagsForPublishing('<role>')` will be also available? If yes, in my opinion it can create a lot of boilerplate in code maintenance and API usage. I think that `publishFor(<role>)` and few others (mainly related to operations/channels) is great, but making API for every intents like `summaryFor...`, `tagsFor...`, `externalDocsFor...` are unnecessary. I very like the current API, because knowing what the specification looks like, I know where to look for it. I also know that other people may not like this solution and I know that in some situation you must do some `for ... loop` to collect some necessary information but these are single situations (at least for me). 

So: Intents like `publishFor(<role>)` is awesome, but then we can leave current api like `publishFor(<role>).message(), publishFor(<role>).bindings()` etc.

Currently I render spec in React component and I see that the additional API proposed in PR for the React component itself is redundant, but as I wrote to someone/for some template it may be useful, but guys :P please try to create as fewer intents as possible that are created due to the way you look at the spec rather than the data the spec has (byt this I like `publishFor(<role>)` but other not), because then you will create an extra 200 methods that can quickly become deprecated in next version of spec. I hope that my comment will be helpful for you

jonaslagoni:
> .... operations/channels) is great, but making API for every intents like `summaryFor...`, `tagsFor...`, `externalDocsFor...` are unnecessary.

In my opinion, we have to @magicmatatjahu, otherwise what is the point of an intent API if it requires the parsers a major version update when the spec updates? If we only go halfway then there is no point in doing it, it would create more confusion than good.

> I don't want to be a party pooper here but overall, it looks that to use the parser I will not only need to know the spec by also some additional vocabulary (`roles`) that is not in the spec.

@derberg correct, but IMO you would need to anyway since the parser and the spec no longer follow 1:1. So to understand the parser you would need to read the documentation anyway.

> To conclude - parser should be resilient to spec changes as much as possible + speed the work on templates, make it nicer, less complicated.
> But yeah, would be great to see your proposal used on some existing code from templates, to see what will change, and if for good or bad

@derberg I am working on actually integrating this suggested API so we can see the changes it would have.

docs/API.md
```diff
@@ -0,0 +1,50 @@
...
+- **(HELP NEEDED)** `AsyncAPIDocument.channelsSubscribes('<role>')` : Channel[] (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407), [800935961](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800935961))
+- **(HELP NEEDED)** `AsyncAPIDocument.channelsPublishes('<role>')` : Channel[] (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407), [800935961](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800935961))
+- **(HELP NEEDED)** `AsyncAPIDocument.isSubscribingToChannels('<role>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319))
+- **(HELP NEEDED)** `AsyncAPIDocument.isPublishingToChannels('<role>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319))
```
fmvilas, line 8: I don't understand the intent behind these methods. If the intent is to understand if we're subscribing to a given channel(s), where's the channel?

I'd make it like this:

```suggestion
- **(HELP NEEDED)** `AsyncAPIDocument.isSubscribingToChannel('<channel name>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319))
- **(HELP NEEDED)** `AsyncAPIDocument.canClientSubscribeToChannel('<channel name>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319))
- **(HELP NEEDED)** `AsyncAPIDocument.isPublishingToChannel('<channel name>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319))
- **(HELP NEEDED)** `AsyncAPIDocument.canClientPublishToChannel('<channel name>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319))
```

Again, as before, the role is implicit in the method name.

docs/API.md
```diff
@@ -0,0 +1,50 @@
...
+- `Server.protocol()` : string (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
+
+### SecurityScheme
+Empty for now.
+
+### Channel
+- `Channel.path()` : string
+- `Channel.messages()` : Message[]
+- **(HELP NEEDED)** `Channel.messagesPublishes('<role>')` : Message[]
+- **(HELP NEEDED)** `Channel.messagesSubscribes('<role>')` : Message[]
```
fmvilas, line 32: What if we just avoid mentioning `publish` and `subscribe` here? Something like `messages() : Message[]`. To make it possible to filter by `publish` and `subscribe` operations, we can probably create another "entity" at the same level as Channel, Server, etc. but called Operation. That Operation object can have a `messages(): Message[]` too, where you only get the messages of this specific operation.

docs/API.md
```diff
@@ -0,0 +1,50 @@
...
+- `Channel.isSubscribing('<role>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319), [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
+- `Channel.isPublishing('<role>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319), [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
```
fmvilas, line 34: I'd port here what I described above.

docs/API.md
```diff
@@ -0,0 +1,50 @@
...
+- `Channel.description()` : string (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319), [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
+- `Channel.extension('<extension property>')` : any (outcome of comment(s): [799598596](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799598596))
+-  **(HELP NEEDED)** `Channel.extensionForPublishing('<role>', '<extension property>')` : any (outcome of comment(s): [799598596](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799598596))
+-  **(HELP NEEDED)** `Channel.extensionForSubscribing('<role>', '<extension property>')` : any (outcome of comment(s): [799598596](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799598596))
```
fmvilas, line 38: If we introduce a new entity called Operation, these things become less weird.

docs/API.md
```diff
@@ -0,0 +1,50 @@
...
+- **(HELP NEEDED)** `Channel.summaryForPublishing('<role>')` : string (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
+- **(HELP NEEDED)** `Channel.summaryForSubscribing('<role>')` : string (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
+- `Channel.parameters()` : Schema[] (outcome of comment(s): [800935961](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800935961))
+- **(HELP NEEDED)** `Channel.operationIdForPublishing('<role>')` : string (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
+- **(HELP NEEDED)** `Channel.operationIdForSubscribing('<role>')` : string (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407))
```
fmvilas, line 46: If we introduce a new entity called Operation, these things become less weird.

fmvilas (commented):
I think it's looking great and you're making great progress here. It's never easy to design an API and you're scratching your head a lot 😄 That's really good 👏 

My advice is that you forget about AsyncAPI for a moment and design the API with generic concepts that may or may not map directly to the AsyncAPI spec. For instance, think of someone creating an EDA. The main concepts that come to the top of my head are:

* My application
* Other applications
* The broker
* The topics on the broker
* The messages on those topics
* The messages my app publish to these topics
* The messages my app subscribe to on these topics
* The messages other apps can publish to these topics
* The messages other apps can subscribe to on these topics

Now try to map these to AsyncAPI:

* My application => AsyncAPI file
* Other applications => Other apps consuming my AsyncAPI file
* The broker => Server
* The topics on the broker => Channels (without publish/subscribe info)
* The messages on those topics => Channel.*.message (where * is either publish/subscribe but we hide it at this level)
* The messages my app publish to these topics => Channel.subscribe.message
* The messages my app subscribe to on these topics => Channel.publish.message
* The messages other apps can publish to these topics => Channel.publish.message
* The messages other apps can subscribe to on these topics => Channel.subscribe.message

Hope it helps! Keep up the great work! 💪

docs/API.md
```diff
@@ -0,0 +1,50 @@
+
+### Root - AsyncAPIDocument
+- `AsyncAPIDocument.hasContentType('<content type>')` : boolean (outcome of comment(s): [799481319](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-799481319))
+- `AsyncAPIDocument.channels()` : Channel[] (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407), [800935961](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800935961))
+- **(HELP NEEDED)** `AsyncAPIDocument.channelsSubscribes('<role>')` : Channel[] (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407), [800935961](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800935961))
```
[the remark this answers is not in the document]
jonaslagoni: The role is described in the [description of the API](https://github.com/asyncapi/parser-js/blob/5ac333dc96c4100a6e7f538581ce4c9d6bb889fc/docs/Intents.md#decisions) and might not be apparent to the user from the start. However, we took the decision to switch and add the role parameter to easy tool developer's interaction with the API. Especially for those tools which use the perspective of both.

It is one of those things that is not an easy decision to make 😅 I can understand if we want to keep it in the name of the intent.

docs/API.md
```diff
@@ -0,0 +1,50 @@
...
+- **(HELP NEEDED)** `AsyncAPIDocument.channelsPublishes('<role>')` : Channel[] (outcome of comment(s): [800282407](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800282407), [800935961](https://github.com/asyncapi/shape-up-process/issues/84#issuecomment-800935961))
```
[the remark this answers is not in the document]
jonaslagoni: Same as above comment.

docs/API.md
jonaslagoni: The API is not seen from the outside, but from a user who uses the API with different use-cases.

docs/API.md
jonaslagoni: Good point, we also discussed whether or not to add the `operation` as an entity. However, decided against it since we felt like it was more a property on the `channel` rather than an entity. We highlighted this in the description:
> So why are operation for example not an entity? Because we don't see the operation as needing an identity, it merely contains values we need associated with the Channel.

docs/API.md
jonaslagoni: We didn't need to check for a specific channel, but in general 😄

jonaslagoni:
@fmvilas thanks for the feedback!

> My advice is that you forget about AsyncAPI for a moment and design the API with generic concepts that may or may not map directly to the AsyncAPI spec. For instance, think of someone creating an EDA. The main concepts that come to the top of my head are:

That is basically also what we did, but I guess the [description](https://github.com/asyncapi/parser-js/blob/5ac333dc96c4100a6e7f538581ce4c9d6bb889fc/docs/Intents.md) does not clarify this well enough 🤔 The idea was to separate the entities completely from the spec and simply think about what concepts make sense and which relationships they have with each other.

I'll see if I can change the description to state this separation better then it is.

docs/API.md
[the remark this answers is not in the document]
fmvilas: The problem with the param is that it introduces a new concept to learn so it increases friction. A developer using both "roles" can simply use ternary operators:

```
const value = x ? clientSubscribableChannels() : suscribingChannels()
```

Still easy for them.

docs/API.md
fmvilas: I don't get it. What was the meaning of `isSubscribingToChannels('<role>')` then? Can you provide a description as if it was for documentation?

docs/API.md
fmvilas: But it's not. An operation is something tightly bound to the application. In the sentence:

```
My app publishes to channel X
```

I'd recognize the operation (verb) as a trait of the application (subject) and the channel is a mere addition. I believe all the weird things happening here and below are because of the lack of an Operation entity.

docs/API.md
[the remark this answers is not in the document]
jonaslagoni: Yea it definitely will create some friction there, however, that friction will resolve itself in terms of code complexity and user interaction (This is just my opinion of course, and slightly biased). 

For example, how would the end user interact with a library that provides both perspectives? (If we think further ahead).

java-spring-template - uses `inverseOperations` as a boolean template parameter.
java-spring-cloud - uses `view` template parameter with the values `client` or `provider`.

At the moment it is up to developers to decide how to handle both views and by deciding on a concept for all tooling such as `role` it would create a common term that all toolings (that use this parser of course). Whether `role` is the correct wording for it that can be discussed 😄.

We might not want to do that, but it would be a huge benefit in my opinion.

docs/API.md
jonaslagoni: Now that I revise it the use-case is wrong or rather not existing since the use-case uses `Channel.isSubscribing('<role>')` instead 🤔 

Gonna remove it.

docs/API.md
smoya: I see the operation as the real intent of the user (Publish? Subscribe? Whatever?) linked optionally to some message(s).

Your suggestion is to create a new `Operation` entity. However, an Operation won't ever exist without, at least, a channel and a message (please correct me if I'm wrong).
Why don't we, instead, give channels and messages to the users, and the ability to filter those by operation (as a concept rather than an entity?).

docs/API.md
jonaslagoni: Removing it since we are only following use-cases at the moment 🙂

docs/API.md
jonaslagoni: Yea, think you might be right here.

cc. @smoya

docs/API.md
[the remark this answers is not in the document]
smoya: > The problem with the param is that it introduces a new concept to learn so it increases friction. A developer using both "roles" can simply use ternary operators:

We had mixed feelings about the `role` concept as well in this place. However, we moved forward because of the reasons you can find in [here](https://github.com/asyncapi/parser-js/pull/263/files#diff-b916c5a0628af4b4bdb634e7612d93408a30119addea9aab98fe5ac7c28adeafR37-R45).
I agree those are not written in stone (@jonaslagoni  wdyt?) so in this case, I agree that **explicit** is better than **implicit** here, and removing the concept of `role` as an argument in favor of generating all the different methods (one per message operation type) would give a self-explanatory API.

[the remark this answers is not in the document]
fmvilas: > For example, how would the end user interact with a library that provides both perspectives? (If we think further ahead).

Looking at the future, this confusion should disappear. Most likely with something like [this](https://github.com/asyncapi/spec/issues/390#issuecomment-662032185).

docs/API.md
fmvilas: > our suggestion is to create a new Operation entity. However, an Operation won't ever exist without, at least, a channel and a message (please correct me if I'm wrong).

An operation would not exist without an Application, a Channel, and a Message. That's the problem. It's linked to all the other entities (except Server, at least not directly). And to make it even more complex, an "Application" here can mean different things: the app described in the AsyncAPI file or the client application. It's a directed graph. That's why you need the application part, to understand the direction of the graph.

```
Application --> publish --> Message --> Channel
```

But who's the Application here? If it's the "provider", then the graph above talks about behavior. If it's the "client", the graph above talks about capability (e.g., a client **can** publish).

We can solve this problem by creating more "verbs" to make it clearer:

```
Application --> performs/allows --> Operation{type, subject, message, channel}
```

Where `type` can be `publish`, `subscribe`, or maybe even `request`. And `subject` can be `provider`/`application` or `client`. I don't foresee more "subjects" appearing but I do foresee more "types" appearing over time, like `query` and `mutation`.

> Why don't we, instead, give channels and messages to the users, and the ability to filter those by operation (as a concept rather than an entity?)

Who are "the users" here? What's going to be the point of view of the operation? From the client or the provider?

smoya: > But who's the Application here? 

That question is repeated several times during our research (but also formulated by other community members within issues) made us introduce the concept of the `role`. We talk about the role here.

> We can solve this problem by creating more "verbs" to make it clearer:

Then you suggest to rather convert Operation as a first-class Entity behaving like the unit of work an application or a  client can perform. 
As a note, you are talking about possible future changes within the spec, which I see are not that far from the first change I emulated in https://github.com/asyncapi/shape-up-process/issues/93#issuecomment-803973100

In conclusion, this will mean our spec would look more Operation-oriented, which in fact it aligns with the average use-case (to describe async operations?).

I have few technical questions regarding the Parser API:

- Would you create a method per operation verb and role (or whatever we wanna call it)? You suggested `AsyncAPIDocument.clientPublishableChannels()` and `AsyncAPIDocument.publishableChannels()`. Would that mean in the future we would need to introduce methods such as `AsyncAPIDocument.clientRequestableChannels()` ?
- Would you introduce similar intents but for Messages and Operations? For example:
  - `AsyncAPIDocument.publishableMessages()`
  -  `AsyncAPIDocument.publishableOperations()`
  -  `AsyncAPIDocument.clientSubscribableOperations()`
  -  `AsyncAPIDocument.subscribableOperations()`
  -  `AsyncAPIDocument.requestableOperations()` (future)
- Similar to the above, but for intents on messages, channels and operations in an intent-fashion shape? For example:
    - `message.clientPublishableChannels()`
    - `channel.PublishableMessages()`
    - `operation.messages()`

fmvilas: Yes, it will —most probably— become more operation-oriented.

And regarding your questions, my answers are:

- Yes
- Yes
- Don't have an opinion :)

docs/API.md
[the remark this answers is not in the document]
smoya: We @jonaslagoni and I decided to remove the role arg in favor of having methods that represent all the intents. Changes were added here https://github.com/asyncapi/parser-js/pull/263/commits/6da212318e582f0927685f25936d882ff6c529a5

docs/API.md
smoya: @fmvilas Changes applied here https://github.com/asyncapi/parser-js/pull/263/commits/6da212318e582f0927685f25936d882ff6c529a5

jonaslagoni: Only the API changed not the description*

jonaslagoni:
After playing around a bit with the API these are some of the thoughts I had.

## Improvements to the API
- `Server` is built around the notion of being global for all channels. I think we should change this to be associated with `Operations`, `Channels`, and `Messages` instead of being connected only to the root of the document.
- `Message.contentType()` should be added and automatically use the global default content type if none is specified for the message.
- A way to check what an operation is, such as `Operation.isClientSubscribing`.

## Breaking changes
It seems like we solved the problems of the breaking changes we had in the last suggestion 👏 

### 3.0.0-fake
Since we already return an array of operations, this will not break the parser API.

### 3.0.0-fake2
Does not break the parser API since we can add support function such as `getAllSchemaFormats` to the standard `getSchemaFormat` so we deprecate the old function as time goes on.

### 3.0.0-fake3
Still works since we have a bidirectional flow of channels, messages, and operations already.

### 3.0.0-fake4
Still works since we have a bidirectional flow of channels, messages, and operations already.

jonaslagoni:
>     * `Server` is built around the notion of being global for all channels. I think we should change this to be associated with `Operations`, `Channels`, and `Messages` instead of being connected only to the root of the document.

@smoya and I discussed it and would like to suggest making the association between servers and operations such as:
- `Server.operations()`
- `Operation.servers()`

Thoughts: Will there be any problems with the generator in terms of the template parameter `server`? 

>     * `Message.contentType()` should be added and automatically use the global default content type if none is specified for the message.

@smoya and I agreed we should add this.

> * A way to check what an operation is, such as `Operation.isClientSubscribing`.

@smoya and I discussed this and came up with a suggestion for adding the following:
```
Operation.isClientSubscribing : boolean
Operation.isClientPublishing : boolean
Operation.isApplicationSubscribing : boolean
Operation.isApplicationPublishing : boolean
Operation.type : ClientSubscribing, ClientPublishing, ApplicationSubscribing, ApplicationPublishing
```
Why add both `type` and the other functions? Because they solve different use cases, adding something specific based on a specific operation such as `if(operation.isClientSubscribing()) return generateClientSubscribe();`. The second use case is for example wanting to get the type of operation for something like documentation.

Thoughts?

jonaslagoni:
Bindings are a bit weird to use, especially if you want to get nested properties such as `channel.binding('nats', 'requestReply').is` if you do want the is a property you have no way of checking if the binding (`is`) exist.

We have two choices at the moment, either `channel.binding('nats').requestReply.is` or `channel.binding('nats', 'requestReply', 'is')`. Me and @smoya suggest using `channel.binding('nats').requestReply.is` since bindings are the wild west.

Would it make sense to have a binding intent such as `channel.binding('nats').isClientRequester()`?

smoya:
@jonaslagoni and I are gonna move with the mocking of the API based on the previous suggestions.
Then, validating them through the following templates:
- https://github.com/asyncapi/nodejs-ws-template
- https://github.com/asyncapi/markdown-template

jonaslagoni:
I am gonna wait with refactoring `Intents.md` until we are sure it does not change again.

docs/Intents.md
```diff
@@ -0,0 +1,56 @@
+# NOTICE - This document is not yet up to date
```
jonaslagoni, line 1: I am gonna wait with refactoring Intents.md until we are sure it does not change again.

PR #264, opened 2021-03-26
fix: duplicate type problem and unresolved Schema type

KhudaDad414:
**Description**

- Fixes types.d.ts duplicate type generation and.
- Fixes types.d.ts unwanted type generation.

**Related issue(s)**

### DIFF
--- a/lib/models/asyncapi.js
+++ b/lib/models/asyncapi.js
@@ -211,13 +211,11 @@ class AsyncAPIDocument extends Base {
    * @param {String} propName if the schema is from a property get the name of such
    * @param {SchemaIteratorCallbackType} callbackType is the schema a new one or is the crawler finishing one.
    * @returns {boolean} should the crawler continue crawling the schema?
-   *
    */
 
   /**
    * Traverse schemas in the document and select which types of schemas to include.
    * By default all schemas are iterated
-   * 
    * @param {TraverseSchemas} callback
    * @param {SchemaTypesToIterate[]} schemaTypesToIterate
    */
--- a/lib/models/asyncapi.js
+++ b/lib/models/asyncapi.js
@@ -209,6 +209,7 @@ class AsyncAPIDocument extends Base {
   /**
    * Callback used when crawling a schema.
    * @callback TraverseSchemas
+   * @alias module:@asyncapi/parser#TraverseSchemas
    * @param {Schema} schema which is being crawled
    * @param {String} propName if the schema is from a property get the name of such
    * @param {SchemaIteratorCallbackType} callbackType is the schema a new one or is the crawler finishing one.
--- a/lib/models/asyncapi.js
+++ b/lib/models/asyncapi.js
@@ -5,7 +5,8 @@ const Info = require('./info');
 const Server = require('./server');
 const Channel = require('./channel');
 const Components = require('./components');
-
+// eslint-disable-next-line no-unused-vars
+const Schema = require('./schema');
 const MixinExternalDocs = require('../mixins/external-docs');
 const MixinTags = require('../mixins/tags');
 const MixinSpecificationExtensions = require('../mixins/specification-extensions');
@@ -206,10 +207,20 @@ class AsyncAPIDocument extends Base {
   }
 
   /**
+   * Callback used when crawling a schema.
+   * @callback TraverseSchemas
+   * @param {Schema} schema which is being crawled
+   * @param {String} propName if the schema is from a property get the name of such
+   * @param {SchemaIteratorCallbackType} callbackType is the schema a new one or is the crawler finishing one.
+   * @returns {boolean} should the crawler continue crawling the schema?
+   *
+   */
+
+  /**
    * Traverse schemas in the document and select which types of schemas to include.
    * By default all schemas are iterated
    * 
-   * @param {Function} callback
+   * @param {TraverseSchemas} callback
    * @param {SchemaTypesToIterate[]} schemaTypesToIterate
    */
   traverseSchemas(callback, schemaTypesToIterate) {
--- a/lib/models/asyncapi.js
+++ b/lib/models/asyncapi.js
@@ -5,8 +5,6 @@ const Info = require('./info');
 const Server = require('./server');
 const Channel = require('./channel');
 const Components = require('./components');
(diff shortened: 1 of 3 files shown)


lib/models/asyncapi.js
```diff
@@ -205,21 +205,11 @@ class AsyncAPIDocument extends Base {
...
-  /**
-   * Callback used when crawling a schema.
-   * @callback TraverseSchemas
-   * @param {Schema} schema which is being crawled 
-   * @param {String} propName if the schema is from a property get the name of such
-   * @param {SchemaIteratorCallbackType} callbackType is the schema a new one or is the crawler finishing one.
-   * @returns {boolean} should the crawler continue crawling the schema?
-   * 
-   */
-
```
jonaslagoni, line 217: I feel like we should require `Schema` instead of removing the entire JSDoc for the callback. What do you think? 🙂

KhudaDad414: @jonaslagoni I agree. How would you suggest I approach it. (Sorry, fairly new to JSdoc)

jonaslagoni: All you have to do is require the Schema model in the top of the model. The linter might not like a dependency that is never "used" per say, if it does just disable that check for that line.

Did that clarify how? :)

KhudaDad414: @jonaslagoni I have tried to do it but no luck.
1) imported the schema.
`// eslint-disable-next-line no-unused-vars
const Schema = require('./schema');`
2) tried using `@requires Schema` tag.
any advice?

jonaslagoni: 1. step is correct 👌 
2. step is to just use the old JSDoc statements as is without modifying it. You dont have to use any tags 🙂

KhudaDad414: But importing the schema makes no difference. Still getting the same error.

jonaslagoni: @KhudaDad414 can you try committing the changes you have, better to spot the problem that way

jonaslagoni: Just in case you did not run it, remember to re-generate the types with `npm run types` before trying to use the changes

KhudaDad414: > Just in case you did not run it, remember to re-generate the types with npm run types before trying to use the changes

I run it.

KhudaDad414: @jonaslagoni 
committed the changes.

jonaslagoni: Ahh yea I see the problem 😅 2 sec.

lib/models/asyncapi.js
```diff
@@ -208,18 +209,18 @@ class AsyncAPIDocument extends Base {
   /**
    * Callback used when crawling a schema.
    * @callback TraverseSchemas
```
jonaslagoni, line 211: I am not sure if this will work. Try the following:
```suggestion
   * @callback TraverseSchemas
   * @alias module:@asyncapi/parser#TraverseSchemas
```

jonaslagoni: Forgot that we wraps everything into the module alias... Which is why just requiring the Schema does not make sense🙂

KhudaDad414: @jonaslagoni Same error. :(

jonaslagoni: Try re-generate the types as well 🙂

jonaslagoni: Or where there no changes? 🤔

KhudaDad414: > Try re-generate the types as well

already done. Schema is still undefined in types.d.ts.

jonaslagoni: Hmm...

lib/models/asyncapi.js
```diff
@@ -208,18 +209,19 @@ class AsyncAPIDocument extends Base {
...
-   * @param {Schema} schema which is being crawled 
+   * @alias module:@asyncapi/parser#TraverseSchemas
```
jonaslagoni, line 212: This took a lot of time to figure out... 

Problem is the type definition HAVE to be declared inside the global module `"@asyncapi/parser"` for it to know where to find the Schema declaration. I tried to change the type of the Schema variable but was unable to do something along the lines of `* @param {module:@asyncapi/parser.Schema} schema which is being crawled`  but with no luck. 

So I ended up with the following definition:

```suggestion
   * @callback module:@asyncapi/parser.TraverseSchemas
   * @param {Schema} schema which is being crawled
```

lib/models/asyncapi.js
```diff
@@ -5,7 +5,8 @@ const Info = require('./info');
 const Server = require('./server');
 const Channel = require('./channel');
 const Components = require('./components');
-
+// eslint-disable-next-line no-unused-vars
+const Schema = require('./schema');
```
jonaslagoni, line 9: This can be removed now

types.d.ts
```diff
@@ -8,40 +8,6 @@ declare type SchemaIteratorCallbackType = {
...
-    payloads: string;
-    headers: string;
-    components: string;
-    objects: string;
-    arrays: string;
-    oneOfs: string;
-    allOfs: string;
-    anyOfs: string;
-};
-
```
jonaslagoni, line 44: This needs to be re-added as well

jonaslagoni (changes requested):
So I ended up figuring out why the JSDoc didn't want to play ball.

Reference: 
- https://github.com/jsdoc/jsdoc/issues/101 - Why `@module` did not work, see related issues.
- https://stackoverflow.com/a/42843182/6803886 - solution

lib/models/asyncapi.js
KhudaDad414: @jonaslagoni made the changes. thanks! couldn't have done it without you.

jonaslagoni (approved):
@KhudaDad414 thanks for the contribution 🙇 

@magicmatatjahu @derberg wanna review it as well?

derberg:
nah, all good from my side, let's have this bug fixed. Thanks @KhudaDad414 🍷

jonaslagoni:
@all-contributors please add @KhudaDad414 for code, bug

Issue #229, opened 2021-01-18
Some schemas are never iterated


jonaslagoni:
### Describe the bug
The following schemas are never iterated and therefore are never assigned ids.
- Standalone component parameters 
- Standalone component traits with headers

### How to Reproduce
The following yaml file should give parameters and headers uids.
```yaml
asyncapi: 2.0.0
info:
  title: My API
  version: '1.0.0'
channels: {}
components:
  parameters:
    testParam:
      schema: 
        type: object
        properties:
          name:
            type: string
  messageTraits:
    extension:
      headers:
        type: object
        properties:
          some-common-header:
            type: string
```

### Expected behavior
Expected schemas to iterated

jonaslagoni:
After some research here are the schemas keywords we need to add to the iterator:

- `patternProperties`
- `definitions`
- `dependencies`

Furthermore this check https://github.com/asyncapi/parser-js/blob/9efaeee7f8dbad2e8b3a2e1a793f178780009599/lib/iterators.js#L65 should be removed completely IMO. Since if properties have been defined (lets say they omit the `type` keyword) we still need to be able to iterate those properties.


Also this check https://github.com/asyncapi/parser-js/blob/9efaeee7f8dbad2e8b3a2e1a793f178780009599/lib/iterators.js#L60 needs to be done at each of the property/array keywords instead, so it should be moved to https://github.com/asyncapi/parser-js/blob/9efaeee7f8dbad2e8b3a2e1a793f178780009599/lib/iterators.js#L186 https://github.com/asyncapi/parser-js/blob/9efaeee7f8dbad2e8b3a2e1a793f178780009599/lib/iterators.js#L209 and wrap the code.

jonaslagoni:
If anyone is up for the task we are always ready to help you out in the implementation of the issue, if it needs more explaining, just let us know 👋

arjungarg07:
I am getting familiar with the codebase of parser-js and will pick up this issue real soon.

arjungarg07:
Hey @jonaslagoni, By `standalone components` you mean which do not have `$ref` (reference) to another type of schemas?
Can you please clear this doubt on `standalone components` Thanks

jonaslagoni:
@arjungarg07 Yes either that or directly referenced by the document such as for `testParam` is never referenced by a channel and therefore never given an id. Because it is a standalone parameter no one uses 😄 
```
asyncapi: 2.0.0
info:
  title: My API
  version: '1.0.0'
[...]

PR #265, opened 2021-03-26
fix: standalone component schemas are never iterated

arjungarg07:
**Description**

This PR fixes the iterator not iterating:
- Standalone component parameters
- Standalone component traits with headers

**Related issue(s)**
#229


derberg:
@jonaslagoni as you are the original reporter of the bug, I wanted to check with you as this PR addresses only the main description and not what was later mentioned in [the comments](https://github.com/asyncapi/parser-js/issues/229#issuecomment-787040871). Is it fine for you? just don't want to jump in and mess around if you maybe discussed it already somewhere

jonaslagoni:
> Is it fine for you? just don't want to jump in and mess around if you maybe discussed it already somewhere

I would prefer if the comment was addressed as well since it is related to the issue. I don't mind if it is done in multiple PR's though.

derberg:
@arjungarg07 up to you then, you want to update this PR or split work?

arjungarg07:
@derberg I will update this PR itself for more convenience.

lib/iterators.js
```diff
@@ -61,19 +67,35 @@ function traverseSchema(schema, callback, prop, schemaTypesToIterate) {
...
-  
-  if (schema.type() !== undefined) {
-    switch (schema.type()) {
-    case 'object':
-      recursiveSchemaObject(schema, callback, schemaTypesToIterate);
-      break;
-    case 'array':
-      recursiveSchemaArray(schema, callback, schemaTypesToIterate);
-      break;
+  if (schema.patternProperties() !== undefined && schemaTypesToIterate.includes(SchemaTypesToIterate.patternProperties)) {
```
arjungarg07, line 70: Shall I make functions like `traversePatternProperties()`, `traverseDependencies()`, `traversePatterndefinitions()`, and wrap the code there because linter tests are not passing due to Cognitive Complexity?

arjungarg07:
Hey @jonaslagoni, In case this slipped 😅 please review it. Thanks😄

jonaslagoni:
> Hey @jonaslagoni, In case this slipped 😅 please review it. Thanks😄

Sorry @arjungarg07 had a small break 😄 Gonna review it now.

lib/iterators.js
```diff
@@ -61,19 +67,35 @@ function traverseSchema(schema, callback, prop, schemaTypesToIterate) {
...
+    for (const [prop, patternPropertiesSchema] of Object.entries(props)) {
+      const circularProps = schema.circularProps();
+      if (circularProps !== undefined && circularProps.includes(prop)) continue;
+      traverseSchema(patternPropertiesSchema, callback, prop, schemaTypesToIterate);
+    }
+  }
+  if (schema.definitions() !== undefined && schemaTypesToIterate.includes(SchemaTypesToIterate.definitions)) {
+    const props = schema.definitions();
+    for (const [prop, definitionsSchema] of Object.entries(props)) {
+      console.log(JSON.stringify(props));
```
jonaslagoni, line 81:
```suggestion
```

lib/iterators.js
jonaslagoni: 🤔 I think we could reduce complexity by providing some helper functions such as https://github.com/asyncapi/parser-js/blob/7bb3aaf2452f336775e1cdd56c290976c25ffcde/lib/iterators.js#L112 but to iterate JS objects. As you can see it is the same code that is being run just with different objects.

Wanna try create such a function? 🙂

jonaslagoni (changes requested):
Overall the implementation looks fine to me! 👏 

we are missing some tests though to ensure that it keeps working. We have some tests for the root function [traverseSchemas](https://github.com/asyncapi/parser-js/blob/ac8def6ba38f887d63c83e1fbfa95fa92c4caf69/test/models/asyncapi_test.js#L390) which can be used as inspiration.

What do you think is best? Adding other tests to that grouping as linked or create another test file just for the iterator itself? 🤔

arjungarg07:
> What do you think is best? Adding other tests to that grouping as linked or create another test file just for the iterator itself? 
🤔

I think it would be good to add another test file just for the iterator.

lib/iterators.js
arjungarg07: Yeah, On it!

arjungarg07:
> What do you think is best? Adding other tests to that grouping as linked or create another test file just for the iterator itself?

@jonaslagoni Shall I make a separate PR for this?

jonaslagoni:
> @jonaslagoni Shall I make a separate PR for this?

hmm 🤔 Lets keep the changes in this PR to a minimum, so I would suggest the refactor of the tests to happen in another one. What do you think @arjungarg07 ?

magicmatatjahu:
@arjungarg07 Is PR still valid?

Issue #201, opened 2020-11-19
Handle in proper way empty object/array in has* like functions


magicmatatjahu:
#### Reason/Context

After discussion in this PR https://github.com/asyncapi/parser-js/pull/192 (comment https://github.com/asyncapi/parser-js/pull/192#issuecomment-729031318) we should consider in which way we should go with handling empty object/array in `has*` like functions. At this moment, `parser-js` returns `true` if given field has object/array with values, and also true if object/array is empty - in those functions is used `!!` statement to convert value to boolean. Unfortunately JS treats empty object/array as `true` value. In opposite, empty object/array should treat as `false` value. 

#### Description

If we decide to change default behavior, then changing logic will be huge breaking change to handle schema in templates.

Also we should discuss point of view from this comment: https://github.com/asyncapi/parser-js/issues/173#issuecomment-699637697

jonaslagoni:
I still have the point of view that `has*` implies something has been defined i.e. it has values or has been defined with values in which empty arrays/objects does not :smile:

derberg:
I like `has*` functions (didn't think if they can have better name) as they imply clearly what function is for. We should have those functions all around  and use them only to determine if given part of spec is present or not. This way, in the future, it will be easy to change the logic inside those, depending how we understand that something `is there` or `isn't`. Function like `properties` should be only to extract properties, but I'd like to easily check if they are there with `hasProperties`.

For me, if object or array are empty, `has*` function should return `false`

derberg:
I'm wondering what is the point of keeping it open 😄

magicmatatjahu:
@derberg 

> At this moment, parser-js returns true if given field has object/array with values, and also true if object/array is empty - in those functions is used !! statement to convert value to boolean. Unfortunately JS treats empty object/array as true value. In opposite, empty object/array should treat as false value

for that :) Probably when we will release `3.0.0` version of AsyncAPI ParserJS will have breaking changes and we can think about handling such a problem.

smoya:
I think you might be interested in joining https://github.com/asyncapi/parser-js/pull/453

smoya:
This is already being handled in the future `v2.0.0` release, which follows the [Intent-Driven API](https://github.com/asyncapi/parser-api).

We could close this ticket once the `v2.0.0` version of the parser is fixed.

magicmatatjahu:
Issue is resolved in v2 ParserJS - now in https://github.com/asyncapi/parser-js/tree/next-major branch

PR #270, opened 2021-03-29
feat: mock intent implementation

jonaslagoni:
**Description**

- ...
- ...
- ...

**Related issue(s)**


lib/parser.js
```diff
@@ -26,20 +26,22 @@ module.exports = {
...
  * @param {Object} [options] Configuration options.
  * @param {String} [options.path] Path to the AsyncAPI document. It will be used to resolve relative references. Defaults to current working dir.
  * @param {Object} [options.parse] Options object to pass to {@link https://apidevtools.org/json-schema-ref-parser/docs/options.html|json-schema-ref-parser}.
  * @param {Object} [options.resolve] Options object to pass to {@link https://apidevtools.org/json-schema-ref-parser/docs/options.html|json-schema-ref-parser}.
  * @param {Object} [options.applyTraits=true] Whether to resolve and apply traits or not.
- * @returns {Promise<AsyncAPIDocument>} The parsed AsyncAPI document.
+ * @returns {Promise<any>} The parsed AsyncAPI document.
  */
-async function parse(asyncapiYAMLorJSON, options = {}) {
+async function parseToJson(asyncapiYAMLorJSON, options) {
```
smoya, line 44: Is this a BC within the current implementation? Is this method the one users are calling right now?

jonaslagoni: This is instead of having duplicate functionality so `parseIntents` and `parse` shares the same functionality.

smoya: Oh I see https://github.com/asyncapi/parser-js/pull/270/files#diff-2f45270c7389933f71b380538ff52e106ee5cb66a0b573d9967467dfdbc11c02R144. nvm then 👍

lib/intent-models/server-security.js
```diff
@@ -0,0 +1,23 @@
...
+ class IntentServerSecurity extends IntentBase {
+    /**
+     * @returns {IntentSecurityScheme}
+     */
+    securityScheme() { return new IntentSecurityScheme();}
+    
+    /**
+     * @returns {string[]}
+     */
+    values() { return []; }
```
jonaslagoni, line 20: Why should `values` not be part of `IntentSecurityScheme` and then we can remove `IntentServerSecurity` 🤔 

cc @smoya

smoya: Because they represent different things. Please correct me if I'm wrong with the following:

A SecurityScheme can be reused by other servers. On the other hand, ServerSecurity holds the config values for a specific SecuritySchema that apply to a particular server.

jonaslagoni: The way I see it is that with the spec is that you define server with a security schema, you define it as:

Server security specified as:
```
security: [
  {
     "api_key": []
  }
]
```

With a security component named `api_key`: 

```
{
  "type": "httpApiKey",
  "name": "api_key",
  "in": "header"
}
```

My intent is to get the security object (located in the security components). Where the so-called values (`[]`) are only used if the security scheme is of type "oauth2" or "openIdConnect", then the value is a list of scope names. 

So to me, it would make sense for `IntentSecurityScheme` to contain a scope names intent which is the previous `values` variable, and then remove `IntentServerSecurity` completely.

This generally makes more sense conceptually to me in terms of a general understanding of what the security model should contain. This current approach is more tuned towards the spec directly.

What do you think?

lib/intent-models/message.js
```diff
@@ -0,0 +1,123 @@
...
+    * @param {string} bindingProtocol 
+    * @returns {any|undefined} 
+    */
+  binding(bindingProtocol) {return {};}
+
+  /**
+   * 
+   * @returns {string} 
+   */
+  contentType() { return 'application/json';}
```
jonaslagoni, line 89: Should this be an enum instead? 🤔 It would make it easier from a tooling perspective, even though the spec says it just a string.

lib/intent-models/message.js
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * @returns {Object<string, any>[]}
+   */
+  examples() { return {}; }
```
jonaslagoni, line 93: I feel like the return type can be better explained, especially since there is quite the confusion about what this example property may contain. What do you think?

lib/intent-models/asyncapi.js
```diff
@@ -0,0 +1,187 @@
...
+ * @alias module:@asyncapi/parser#IntentAsyncAPIDocument
+ * @extends IntentBase
+ */
+class IntentAsyncAPIDocument extends IntentBase {
+  /**
+   * @returns {string}
+   */
+  version() {
+    return '2.0.0';
+  }
```
jonaslagoni, line 25: Does it make sense to call it `version()` rather than what the spec uses `asyncapi()`?

lib/intent-models/channel.js
```diff
@@ -0,0 +1,67 @@
...
+   * @param {string} bindingProtocol 
+   * @returns {boolean}
+   */
+  hasBinding(bindingProtocol) { return true; }
+  /**
+    * 
+    * @param {string} bindingProtocol 
+    * @returns {any|undefined} 
+    */
+  binding(bindingProtocol) {return {};}
```
jonaslagoni, line 43: I would suggest we use enums here for `bindingProtocol` rather than `string`, what do you think?

lib/intent-models/channel.js
```diff
@@ -0,0 +1,67 @@
...
+ * 
+ * @class
+ * @alias module:@asyncapi/parser#IntentChannel
+ * @extends IntentBase
+ */
+class IntentChannel extends IntentBase {
+  /**
+   * @returns {IntentOperation[]}
+   */
+  operations() { return [new IntentOperation()];}
```
jonaslagoni, line 20: Should we include all the intents from the root in terms of the point of view for operations, such as `applicationPublishOperations` and `applicationSubscribeOperations`?

lib/intent-models/server.js
```diff
@@ -0,0 +1,88 @@
...
+  hasUrl() { return true; }
+  /**
+   * @returns {string|undefined}
+   */
+  url() {return 'localhost';}
+
+  /**
+   * @returns {IntentOperation[]}
+   */
+  operations() {return [new IntentOperation()];}
```
jonaslagoni, line 45: Should we include all the intents from the root in terms of the point of view for operations, such as `applicationPublishOperations` and `applicationSubscribeOperations`?

jonaslagoni (commented):
@smoya and I have added all intents as we see they should be and tweaked a few things. However, we need your help in finalizing it.

I have included a review with questions I had while going over the API, feel free to add your own review as well, we want to hear your opinion 🙏 

We suggest that we slightly change how we view the parser API or at least suggest how it should be seen. Instead of trying to make the API fit perfectly to the spec (so you can access all properties) what if we just focus on what makes sense from the event-driven point of view? We can still give access to the underlying model in the future, so tool developers always have the choice to go around the intent API in edge cases where this is the only choice. For example: why would we want to give access to traits? They are something for the spec to allow the inheritance of traits between messages etc.

Adopting this view these are the properties from the spec which are purposefully not added to the API:
- Traits
    - Since whether a property comes from a trait or not should not matter from tooling perspective.
- Message schema format
    - It is a property for parsers to know how to parse schemas, but is it relevant to include it in the API?
- Components
    - What is the intent to only access messages, etc from components. I know some define messages long before channels are defined. But shouldn't those be accessed with `AsyncAPI.messages()`?

Questions:
- Should we include `isEmpty()` support functions for those properties that return an array? Or should we still use `hasX()`? Related parser issue: https://github.com/asyncapi/parser-js/issues/201
- What about external docs. Would it not make sense to wrap all descriptions in a new model called `Description` which then contains all these different aspects of a description?
- What functions should always return a value (i.e. not being optional)? Should we follow all required properties from the spec or? We need to remember that changing a value from required to optional is a breaking change, so the fewer the better.

lib/intent-models/asyncapi.js
jonaslagoni: Also I think we can make this enum right?

lib/intent-models/asyncapi.js
```diff
@@ -0,0 +1,177 @@
...
+   */
+  hasId() {
+    return true;
+  }
+  /**
+   * @returns {string|undefined}
+   */
+  id() {
+    return 'test-id';
+  }
```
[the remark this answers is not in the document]
jonaslagoni: If we do this `.json()` makes little sense for the `Info` object since the property wont be there.

lib/intent-models/asyncapi.js
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * 
+   * @returns {boolean}
+   */
+  hasExternalDocs() { return true; }
+  /**
+    * 
+    * @returns {IntentExternalDocument|undefined}
+    */
+  externalDocs() { return new IntentExternalDocument();}
```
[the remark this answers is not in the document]
jonaslagoni: same thing with https://github.com/asyncapi/parser-js/pull/270/files#r617758980

lib/intent-models/asyncapi.js
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+  * @returns {boolean}
+  */
+  hasTags() { return true; }
+  /**
+   * @returns {IntentTag[]|undefined}
+   */
+  tags() { return [new IntentTag()]; }
```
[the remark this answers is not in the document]
jonaslagoni: same thing with https://github.com/asyncapi/parser-js/pull/270/files#r617758980

lib/intent-models/license.js
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * 
+   * @returns {boolean}
+   */
+  hasUrl() { return true; }
+  /**
+   * @returns {string|undefined}
+   */
+  url() {return 'test.com'; }
+}
```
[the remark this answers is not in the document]
smoya: I don't see the benefit.

lib/intent-models/asyncapi.js
smoya: What's the benefit on doing enum? In fact it penalizes readability: `Version200` vs `'2.0.0` ?

lib/intent-models/channel.js
smoya: Question: are users able to create bindings without modifying the spec? If so, we should not use enum here since users may want to create their own bindings and then, will need to change the parser when in fact won't be needed.

lib/intent-models/server.js
smoya: I am not pretty sure if there is a clear use case. But perhaps makes sense. Could be v2.1.0 anyway so we don't have to add those now.

lib/intent-models/message.js
jonaslagoni: The correct return type is `[Map[string, any]]` where `string` is either `headers` or `payload`. 

But we also have to be careful about this, cause there is an incoming breaking change here: https://github.com/asyncapi/spec/issues/329 
Not sure if we can reduce the impact in any way 🤔

lib/intent-models/asyncapi.js
jonaslagoni: While you may read it that way, from tooling perspective it just becomes easier to check a statement such as `if(asyncapi.version() === AsyncAPI.versions.Version200)`
vs
`if(asyncapi.version() === "2.0.0")` where the developer need to look up what values this might be.

lib/intent-models/channel.js
jonaslagoni: They cannot, bindings are part of spec, what you are thinking of is extensions. 

Binding fields are referred to as fixed fields: https://www.asyncapi.com/docs/specifications/2.0.0#channelBindingsObject

lib/intent-models/server.js
jonaslagoni: Isn't the use case the same as the one for the root object? Just depend on which object you currently have access to 🤔

lib/intent-models/license.js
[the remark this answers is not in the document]
jonaslagoni: hmm, yea the reason for suggesting it was to reduce the number of models as much as possible, where it makes sense. 

Not sure this one does either 🤔

lib/intent-models/channel.js
smoya: Then yes, we can do it. I will update the doc introducing this change asap. Would you mind updating this mock if you have time? Otherwise I'll do.

lib/intent-models/asyncapi.js
jonaslagoni: @smoya do you disagree or? 😆

lib/intent-models/channel.js
jonaslagoni: Hmmm, now that I think about it, what if we did:
```js
  /**
   * 
   * @param {string | ProtocolEnum} bindingProtocol 
   * @returns {boolean}
   */
  hasBinding(bindingProtocol) { return true; }
```
This way we don't restrict any use cases while still providing the same help for tooling if they want it? 🤔

jonaslagoni: While enums, in general, make sense, I am not sure there does not exist a use case where strings are easier to use..

lib/intent-models/asyncapi.js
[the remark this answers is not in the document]
smoya: IIRC we decided to not include `.json()` at least in the first version.

lib/intent-models/asyncapi.js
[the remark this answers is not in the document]
smoya: I think makes sense. I'm moving it on the API description. Would you mind doing the same at parser mock level?

lib/intent-models/asyncapi.js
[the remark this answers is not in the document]
smoya: I'm moving this to `Info` model at API doc level. Would you mind doing the same at Parser mock side?

lib/intent-models/asyncapi.js
smoya: LOL yeah I disagree. I'm leaning towards readability here.

lib/intent-models/channel.js
smoya: I think I will leave it for the next version ?

lib/intent-models/channel.js
smoya: This is just the detail of implementation. Each programming language implement (or not) enums in a different fashion. At the end an enum here means an `string` that can be **only** one value of a fixed list.

lib/intent-models/info.js
```diff
@@ -0,0 +1,72 @@
...
+   */
+  hasContact() {
+    return true;
+  }
+  /**
+   * @returns {IntentContact|undefined}
+   */
+  contact() {
+    return new IntentContact();
+  }
```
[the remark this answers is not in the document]
smoya: Modified the API doc. Would you mind changing it in the Parser mock side?

lib/intent-models/license.js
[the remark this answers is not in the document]
smoya: I still don't see the benefit. License object could grow in the future.

lib/intent-models/message.js
smoya: No, they should not be a enum. [MIME Types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) are **not only** an official list of them but rather users can create their custom ones.

lib/intent-models/message.js
smoya: As far as I see, https://github.com/asyncapi/spec/issues/329 will introduce a breaking change on the spec. 
As the return type will change on the Parser API, we could consider introducing right now a new model called `Example`. However, I don't see how we could make them handle the raw version (`[Map[string, any]]`) without adding a method that returns it explicitly. For me this sounds like a patch and not very intuitive.
We could also return either a `[Map[string, any]]` or an `Example` Object depending on the content of the examples json, however I wouldn't advocate for that as it's completely a bad practice, also not pretty suitable and common in strong typed languages.

I would just carry on with a possible breaking change here...

smoya: Btw, isn't the type nomenclature `Map[string, any][]` rather than `[Map[string, any]]`?

lib/intent-models/asyncapi.js
jonaslagoni: Can you explain why you think a string is more readable from the tooling perspective? 🤔

lib/intent-models/asyncapi.js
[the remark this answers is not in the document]
jonaslagoni: 👍

lib/intent-models/asyncapi.js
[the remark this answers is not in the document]
jonaslagoni: 👍 Done

lib/intent-models/channel.js
jonaslagoni: Alright 👍

lib/intent-models/info.js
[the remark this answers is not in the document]
jonaslagoni: 👍 Done

lib/intent-models/license.js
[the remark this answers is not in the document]
jonaslagoni: Fair point.

lib/intent-models/message.js
jonaslagoni: Fair point 👍

lib/intent-models/asyncapi.js
smoya: I mean, tooling is written by people. People will prefer to read the version as plain string rather than a constant?
I mean, `AsyncAPI.versions.Version200` could mean version `200`? It is so hard to read.

lib/intent-models/asyncapi.js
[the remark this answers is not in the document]
smoya: Totally. I modified the docs at https://github.com/asyncapi/parser-api/pull/1/commits/7c07ea5be2723919ee92493d8997035b2a1a1987. Would you mind updating the parser mock side?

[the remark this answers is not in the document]
jonaslagoni: Done 👍

lib/intent-models/message.js
jonaslagoni: 🤔 Yea I think it is 🧐

lib/intent-models/asyncapi.js
jonaslagoni: True, fair, lets leave it for now.

lib/intent-models/channel.js
jonaslagoni: alright, well I adapted the mock.

lib/intent-models/message.js
jonaslagoni: I have changed the return type to `Map[string, any][]` in the mock

lib/intent-models/server-security.js
jonaslagoni: Changed the mocks.

lib/intent-models/channel.js
```diff
@@ -0,0 +1,67 @@
...
+  binding(bindingProtocol) {return {};}
+
+  /**
+   * 
+   * @returns {boolean}
+   */
+  hasParameters() { return true; }
+  /**
+   * 
+   * @returns {Schema[]|undefined}
```
smoya, line 52: Wrong return type. Should be `IntentChannelParameter[]`

smoya: Nevermind. With latest changes on the API is not needed.

PR #280, opened 2021-04-07
fix: wrong position info about errors in the root of the doc

### LINKED ISSUE #240: Wrong position information about errors in the root of the document
#### Describe the bug
If I add some additional property to the root of the document (which is not allowed) I get an error like below (if I combine it from `title` and `startLine` and `startColumn`)

```
[
  "/ should NOT have additional properties on line undefined, column undefined"
]
```

#### How to Reproduce
Try to parse AsyncAPI document with property `test: true` on the root of the document.

#### Expected behavior
I should not only get information `/ should NOT have additional properties` but also an indication on where the error happened in the document so the user can easily find it

KhudaDad414:
**Related issue(s)**
Fixes #240

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -243,7 +243,7 @@ utils.improveAjvErrors = (errors, asyncapiYAMLorJSON, initialFormat) => {
   return errors.map(error => {
     const defaultLocation = { jsonPointer: error.dataPath || '/' };
     const additionalProperty = error.params.additionalProperty;
-    const jsonPointer = additionalProperty ? error.dataPath || `/${additionalProperty}`: error.dataPath;
+    const jsonPointer = additionalProperty ? `${error.dataPath}/${additionalProperty}`: error.dataPath;
     return {
       title: `${error.dataPath || '/'} ${error.message}`,
       location: ast ? findLocationOf(jsonPointerToArray(jsonPointer), ast, initialFormat) : defaultLocation,
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -245,7 +245,7 @@ utils.improveAjvErrors = (errors, asyncapiYAMLorJSON, initialFormat) => {
 
     return {
       title: `${error.dataPath || '/'} ${error.message}`,
-      location: ast ? findLocationOf(jsonPointerToArray(error.dataPath), ast, initialFormat) : defaultLocation,
+      location: ast ? findLocationOf(jsonPointerToArray(error.dataPath || `/${error.params.additionalProperty}`), ast, initialFormat) : defaultLocation,
     };
   });
 };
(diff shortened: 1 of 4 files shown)


KhudaDad414:
@derberg I know this is a nasty approach, but I couldn't figure out a better way. any comments?

derberg (commented):
@KhudaDad414 looks good but could you please add a test that will protect us in future. Simple test where you provide some new prop on root in the document and expect an error on a given position in the document, in here https://github.com/asyncapi/parser-js/blob/master/test/parse_test.js

lib/utils.js
```diff
@@ -242,10 +242,11 @@ utils.improveAjvErrors = (errors, asyncapiYAMLorJSON, initialFormat) => {
   const ast = getAST(asyncapiYAMLorJSON, initialFormat);
   return errors.map(error => {
     const defaultLocation = { jsonPointer: error.dataPath || '/' };
-
+    const additionalProperty = error.params.additionalProperty;
+    const jsonPointer = additionalProperty ? error.dataPath || `/${additionalProperty}`: error.dataPath;
```
derberg, line 246: so you mean `error.dataPath` is not there when we talk about additional properties on the root?

KhudaDad414: Yes. That seems to be the problem.

KhudaDad414: > @KhudaDad414 looks good but could you please add a test that will protect us in future. Simple test where you provide some new prop on root in the document and expect an error on a given position in the document, in here https://github.com/asyncapi/parser-js/blob/master/test/parse_test.js

@derberg tests added for both scenarios. (root and non-root additional properties)

test/parse_test.js
```diff
@@ -103,19 +103,44 @@ describe('parse()', function() {
...
+            startLine: 15,
+            startColumn: 1,
+            startOffset: offset(305,15),
+            endLine: 15,
+            endColumn: 11,
+            endOffset: offset(315,15)
+          }
+        },
+        {
+          title: '/info should NOT have additional properties',
```
derberg, line 119: just to make sure. To get information here, in the info object, about the additional property we would need to start reading `params: { additionalProperty: 'test' }` from ajv?

derberg: no wait, sorry, you already read it here https://github.com/asyncapi/parser-js/pull/280/files#diff-5dfe38baf287dcf756a17c2dd63483781b53bf4b669e10efdd01e74bcd8e780aR245

but it is not used in this error because `error.dataPath` is probably there as it is a nested object. Do I get it right?

KhudaDad414: @derberg If the additional property is not in the root, then `error.dataPath` contains the path to its parent. But if it is in the root, it has no parent and `error.dataPath` will be an empty string. I solved the problem by passing the additional property to `findLocationOf` (instead of `error.dataPath`). then `findLocationOf` locates the additional property and returns its location.

derberg: then in my opinion we should concat `error.dataPath` with `additionalProperty` if both are available, so later you get a helpful error that point to `/info/test`. Makes sense?

KhudaDad414: Yes. pushed the changes. works perfectly.

derberg (approved):
well done 🥳

PR #282, opened 2021-04-08
feat: tck integration

### LINKED ISSUE #137: Make sure new parser releases are not introducing any regression
#### Reason/Context
Parser has many tests but not all cases are covered. https://github.com/asyncapi/tck/ was created to hold all the possible valid and invalid cases. We should make sure that with new releases the percentage of the parser compatibility doesn't decrease. 

#### Description
Maybe we can add an action that runs tck on a PR and validates the numbers and block merge? tck outputs JSON with results so this should be relatively easy

KhudaDad414:
**Related issue(s)**
Resolves #137

### DIFF
--- a/asyncapi-parser.json
+++ b/asyncapi-parser.json
@@ -1,0 +1,1774 @@
+{
+  "parser": {
+    "name": "asyncapi-parser",
+    "language": "js",
+    "url": "https://github.com/asyncapi/parser-js",
+    "version": "^1.4.4"
+  },
+  "results": [
+    {
+      "file": "/tests/asyncapi-2.0/AsyncAPI Object/Fields Types/invalid-asyncapi-type.yaml",
+      "success": false,
+      "error": "Error: parsedJSON.asyncapi.startsWith is not a function"
+    },
+    {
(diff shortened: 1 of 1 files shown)


KhudaDad414:
@derberg I was playing around with tck and I found out that 102  invalid files passed from current parser-js plus 10 valid files did not pass. I have added the result in this PR. is this normal?

asyncapi-parser.json
```diff
@@ -0,0 +1,1774 @@
...
+      "success": true
+    },
+    {
+      "file": "/tests/asyncapi-2.0/Channel Bindings Object/amqp/Fields Types/invalid-exchange-type.yaml",
+      "success": true
+    },
+    {
+      "file": "/tests/asyncapi-2.0/Channel Bindings Object/amqp/Fields Types/invalid-exchange-vhost-type.yaml",
+      "success": true
+    },
```
KhudaDad414, line 117: some invalid files that passed.

derberg: yes, it is possible that some invalid files will pass because parser doesn't have 100% coverage yet 😞 like for example we do not validate payload examples yet. Have a look at the report https://asyncapi.github.io/tck/

maybe before we actually can use TCK we should first make sure the parser has 100% coverage for both, valid and invalid files 🤔

Issue #119, opened 2020-07-14
Improve error handling in parser


magicmatatjahu:
#### Reason/Context

On the Slack, there was a discussion in which several proposals were proposed to improve the handling of errors in parser.

#### Description

Proposals:

1. Add `custom-validation-error` new error type for custom validation like: validate variables of server name etc...

2. Create the predefinied classes for error types like in models:

```js
class ValidationError extends ParserError {
  super(...args) {
     args.type = "validation-errors";
     super(args);
  }

  // additional logic
}
```

3. Think about subtypes for `custom-validation-errors`, for `channels`, `server-variables` etc. It would be great if it will be implemented with HTTP Problems RFC.

> **NOTE:** this is improvement not for 1.0.0 but for the next version.

derberg:
I'm with the idea of introducing more error types!

I also think we need to make it super clear what is our approach to validation vs throwing errors. Should we always throw errors and have no warnings? or should we always throw errors and leave it up to the parser users to decide if a particular error type they want to silent or not.

Why?
This cases:
- https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Parameter%20Object/valid-parameter-not-defined.yaml
- https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Server%20Variable%20Object/valid-variable-not-defined.yaml

now we are throwing errors that some Parameter or Variable object is not provided. The thing is that it is not mandatory to have those objects. Tck shows parser doesn't work as expected.

We might need to maybe introduce some kind of `strict` flag and throw such special error only if `strict: true`

derberg:
We discussed it during an open meeting. We had an agreement that having a kind of `silent` flag is good as long as it is `false` by default. We just did not discuss in detail what can be `silent`. For now, my suggestion is to use the spec as a "decision maker" here, so if something is described by the spec and is not done in the document should be a normal error, others (even if important for code gen, but not for docs) should be available for `silent`. But now when I think about it, `silent` instead of boolean could be an array of error types, that you can pass and silent 🤔  We will have to make a good proposal here when we start coding

BOLT04:
Hi @magicmatatjahu and @derberg 🙂, this looks like an interesting improvement, but I have some questions.

> It would be great if it will be implemented with HTTP Problems RFC

[...]

PR #284, opened 2021-04-11
feat: tck 100% coverage

### LINKED ISSUE #283: 100% coverage for tck
#### Reason/Context
before we can integrate tck into this repository(for closing #137 ), we have to make sure that parser-js has  100% tck coverage.

KhudaDad414:
**Description**

- I am planning to integrate tck into parser-js. before doing that we have to make sure that tck has 100% coverage in this repository.
**Related issue(s)**
Resolves #283


KhudaDad414:
@derberg I know you have a lot on your plate right now, but can you confirm if this is a good approach to add 100% coverage for tck? after your confirmation, I can add validation for 100% of tck files (I think).

derberg:
@KhudaDad414 adding 100% coverage for tck would be awesome, but it should not happen in one run but one test case after another to not overcomplicate review process and have it smooth.

big challenges ahead:
- sometimes we validate and throw error on something that actually should just be a warning and not error, but problem is we do not have a concept of `warning` in the parser, that warns but doesn't fail parsing -> https://github.com/asyncapi/parser-js/issues/119#issuecomment-668506614. Fixing this first will block you from having a 100% coverage
- majority of failures you can see here https://asyncapi.github.io/tck/asyncapi-parser_js_detailed_report.html is that test cases have not only simple documents but also documents with bindings. Problem is that the current spec doesn't require providing a json schema for binding, and therefore there is no easy way to do validation. In my opinion, handling types definitions in our code is not the best approach, hard to maintain, especially that we expect bindings and their releases are managed by different community members that not necessarily care about JS parser. I n my opinion we should rather make sure we have JSON Schemas for bindings even if it is not required by the spec

derberg:
@KhudaDad414 we clarified a bit, have a look at https://github.com/asyncapi/spec/issues/507. Best approach is to create json schema files for bindings

KhudaDad414:
@derberg thanks! I will start working on it soon.

PR #285, opened 2021-04-12
fix: support tree-shakable models/utils

magicmatatjahu:
**Description**

Working on https://github.com/asyncapi/shape-up-process/issues/88 I have to remove unnecessary dependencies as much as possible while bundling. I wanna create two bundles for browser:
- with parser onboard
- and without (for hydration)

In the second case I have to remove all parser unnecessary stuff, but unfornatelly bundler cannot properly tree-shakable unnecessary utils from models - in component we use directly two models `AsyncAPIDocument` and `Schemas` to support parser-js API. Problem is with [this file](https://github.com/asyncapi/parser-js/blob/master/lib/utils.js) - bundler treats `cjs` module as JS object, not like `esm` module so includes everything from "module". This PR moves every util related to models to separate file.

Examples with weight:
- component with parser - about 800kb
- component without parser but without this PR - about 445kb
- component without parser and with this PR (tested locally) - about 170kb

Component without parser after my changes has also included `Buffer` due to this line https://github.com/asyncapi/parser-js/blob/master/lib/models/message.js#L16 Buffer works only on NodeJS and bundler tried to resolve it by adding about 64kb additional code related only to `Buffer` polyfill... I will try to handle the Buffer later in this cycle, but I prefer to leave it and go with completing more important tasks. Removing almost 650kb is a big deal anyway.

**Related issue(s)**
Part of https://github.com/asyncapi/shape-up-process/issues/88
Part of https://github.com/asyncapi/html-template/pull/166


derberg:
shouldn't model utils go to `models` dir then?

were there any change in the logic of utils once you moved them, hard to spot when there is a change in code location. Hilfe!

magicmatatjahu:
@derberg I thought about that, but `models-utils` are used also in `mixins`, but this is not a problem to move logic to `models/utils.js` file.

> were there any change in the logic of utils once you moved them, hard to spot when there is a change in code location. Hilfe!

Nope, only moved utils to `models-utils` and fixing imports in files to appropriate.

Thanks for review!

derberg:
then do `models/utils.js` and `mixings/utils.js` ?

magicmatatjahu:
@derberg Ok, I moved to `models/utils.js` file.

derberg (approved):
lgtm

magicmatatjahu:
Thanks!

Issue #266, opened 2021-03-29
Stringify parsed schema


magicmatatjahu:
Currently if you want stringify parsed schema you must do it manually by `JSON.stringify` function. However, here we have a problem with circular schemas and we must add as third parameter custom function to handle these circular schemas.

I suggest to have static `stringify` function inside `AsyncAPIDocument` to handle described problem:

```js
class AsyncAPIDocument {
  static stringify(spec) {
    // operation on spec._json to handle circular schemas
    // all internal extensions `x-parser` will be included inside stringified schema
  }
}
```

Some use cases for this feature:
- handle validation and parsing in server-side and then only send parsed schema to front-end part - I have PR to show my idea in action https://github.com/asyncapi/studio/pull/17 Generally, we parse schema in server-side, then we send parsed schema as string and pass schema as `new AsyncAPIDocument({parsedSchema})` to React component,
- for hydration in unification html-template and React component - very similar approach as 1. In templates we have parsed schema saved as `asyncapi` parameter to template. We can stringify it and save it as string and and pass schema as `new AsyncAPIDocument({parsedSchema})` to React component - example: https://github.com/asyncapi/html-template/blob/41fd1a68fcfdbe5c2a52265e270cc5afbc0eec95/template/index.js#L78

I can make this logic in mentioned repos but we can add it as feature to parser.

jonaslagoni:
I like the idea, might want to add it to `Base` since one might want to stringify different kind of objects and not just the root.

derberg:
do I get it right, you want to parse it to get a string and then pass it to the component that parses it again? or am I missing something? I case I'm right, isn't it better to make react component accept `AsyncAPIDocument` and skip parsing in the component?

magicmatatjahu:
@derberg You're right about React component, it should accepts the `AsyncAPIDocument` instance, but problem is with circular schemas inside spec. As I wrote if you parse spec in server-side, you must stringify it before send, but if you have circulars schemas then you will have error with stackoverflow, because `JSON.stringify` cannot handle deep circular references and our `stringify` function will be only wrapper for `JSON.stringify` to handle these circular refs. If something is incomprehensible, let me know :)

Some code how it will look in sever-side and front-end apps:

```ts
// server-side
const spec = ...
const parsedSpec = new Parser(...).parse(spec);
const stringifiedSpec = parsedSpec.stringify();
response.send(stringifiedSpec, ...)

// front-end app
[...]

PR #288, opened 2021-04-22
feat: add x-parser-spec-parsed extension

magicmatatjahu:
**Description**

Add `x-parser-spec-parsed: true` extension to root of parsed spec, similar like `x-parser-message-parsed` to avoid unnecessary calculations. I have use case when I stringify parsed spec (see [issue](https://github.com/asyncapi/parser-js/issues/266)) and after stringification I must pass this spec again to `AsyncAPIDocument` constructor. Also I fixed bug, because when we are assigning anonymous id of anonymous message we should first check that id is defined - for schemas we do it. 

**Related issue(s)**
See https://github.com/asyncapi/html-template/pull/166#pullrequestreview-642090407 especially part with:

> also this anonymous-message-1 should not be there 🤔 maybe my messages do not have name prop or title but they are in components and the key name should be applied right?


derberg (commented):
lgtm, can you just please add some comment [here](https://github.com/asyncapi/parser-js/pull/288/files#diff-a76417acabb91ce292a6285966f525f516f90be36426823ad0e410ce8c81e901R43) so we do not have to dig in the history later on to figure out why we assign this prop. 

Tbh, we already have so many parser specific extensions that I think it is about time to think about documenting them 😅 what do you think?

derberg:
also, tests are failing and sonarcloud complains, have a look

magicmatatjahu:
@derberg Thanks! I will fix tests and sonar clouds errors (I wrote PR yesterday about 23pm 😅).

> Tbh, we already have so many parser specific extensions that I think it is about time to think about documenting them 😅 what do you think?

Something like that?

```md
## Custom extensions

Parser during validation and parsing adds custom extensions (started by `x-parser`) that add additional information about the specification:

| Name | Value | Location | Description |
|---|---|---|
| `x-parser-spec-parsed` | `true` | Root of parsed specification | Indicates that specifixation is parsed |
```

derberg:
> I wrote PR yesterday about 23pm 

then maybe I should actually do a deeper code review 🤔 😆 

for the docs, you went to far, just add a comment in code, readme is something for a followup, don't spend to much on it now, just open an issue if you want

derberg (approved):

PR #292, opened 2021-04-24
test: add schema tests for the cases where properties are missing

### LINKED ISSUE #278: Improve test coverage for the schema module
This issue was created as a result of suggestions made in issue #233.

The schema module currently has missing tests for some conditions as highlighted in #233.

As my first contribution to the parser project I will be adding these missing tests.

fr3shw3b:
**Description**

- Adds tests to Schema model to handle the cases when properties aren't provided in the json document that the Schema is initialised with.

**Related issue(s)**
Resolves #278


magicmatatjahu (approved):
LGTM!

PR #301, opened 2021-05-19
chore: enable releases for spec-related release branches

derberg:
see also https://github.com/asyncapi/spec/issues/536

it is supposed to work the same as here https://github.com/asyncapi/spec/pull/539 + npm release

relevant workflow configuration is in master already as it came from a global workflow https://github.com/asyncapi/parser-js/commit/c9d751ce3e5591f0b506db2a84b5dc6dff080f11

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "homepage": "https://github.com/asyncapi/parser-js",
   "devDependencies": {
     "@semantic-release/commit-analyzer": "^8.0.1",
-    "@semantic-release/github": "^7.0.4",
+    "@semantic-release/github": "7.2.3",
     "@semantic-release/npm": "^7.0.3",
     "@semantic-release/release-notes-generator": "^9.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
   "homepage": "https://github.com/asyncapi/parser-js",
   "devDependencies": {
     "@semantic-release/commit-analyzer": "^8.0.1",
-    "@semantic-release/github": "^7.0.4",
+    "@semantic-release/github": "7.2.3",
     "@semantic-release/npm": "^7.0.3",
     "@semantic-release/release-notes-generator": "^9.0.1",
```
magicmatatjahu, line 46: You can update also this package to `9.0.2`, but it's optional.

magicmatatjahu (approved):

package.json
derberg: yeah I don't want to 😄

PR #304, opened 2021-05-21
test: added iteration schema keyword tests

arjungarg07:
**Description**
This PR adds tests for schemas keywords such as patternProperties, definitions , dependencies added to the iterator.
 fixes: https://github.com/asyncapi/parser-js/issues/229


derberg:
@arjungarg07 hey, please have a look at sonar cloud feedback, and also remove changes to `bundle.js` from the PR as it is updated later after merge by automated flows

arjungarg07:
> @arjungarg07 hey, please have a look at sonar cloud feedback, and also remove changes to `bundle.js` from the PR as it is updated later after merge by automated flows

On it.

lib/iterators.js
```diff
@@ -55,25 +61,38 @@ const SchemaTypesToIterate = Object.freeze({
  * @param {TraverseSchemas} callback 
  * @param {SchemaTypesToIterate[]} schemaTypesToIterate
  */
+/* eslint-disable sonarjs/cognitive-complexity */
```
arjungarg07, line 64: disabled cognitive-complexity for this one, then too sonar cloud code smell issue is coming.

test/models/iterator_test.js
```diff
@@ -0,0 +1,35 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const AsyncAPIDocument = require('../../lib/models/asyncapi');
+
+describe('AsyncAPIDocument', function () {
+  describe('constructor', function() {
+    it('Should traverse patternProperties, definitions, dependencies ', function () {
+      const doc = JSON.parse(fs.readFileSync(path.resolve(__dirname, '../good/iterator.json'),'utf8'));
+      const d = new AsyncAPIDocument(doc);
```
arjungarg07, line 11: Why sonar cloud is giving me a bug here stating `Replace AsyncAPIDocument with a constructor function.`

jonaslagoni: I actually have no idea, to me, this sounds like a false positive.

@derberg any idea 🤔

arjungarg07: Yeah! the same bug is reported 91 times in every file by sonar in this PR too: [integrate eslint jsdoc validation](https://github.com/asyncapi/parser-js/pull/305)

For Reference:

Screenshot 2021-05-25 at 9 16 53 PM

derberg: did you click `Why is this an issue?` link? the answer is there 😛 
although confusing is that actually `AsyncAPIDocument` has a constructor that has `@constructor` jsdoc. Maybe the problem here is that `AsyncAPIDocument` extends `Base` that doesn't have a constructor with `@constructor` jsdoc. Try this path.

lib/iterators.js
jonaslagoni: It is fine, the reviewer will check whether it makes sense to disable it. Right of first glance, I don't see how you can reduce the size here, another reviewer might be able to 🙂 

Just a general rule of thumb think about whether it makes sense to refactor the complexity or not. Would it be more complex to split any of the functionality up? If the answer is yes, then disable the eslint, otherwise, it should of course be refactored.

Did that clarify it?

arjungarg07: Yeah, that makes sense. Also, the PR is ready from my side.

test/models/iterator_test.js
arjungarg07: was figuring a way to fix that but couldn't, thanks for pointing the workaround and now it's done 🙂

lib/iterators.js
```diff
@@ -206,6 +236,7 @@ function recursiveSchemaObject(schema, callback, schemaTypesToIterate) {
  * @param {SchemaTypesToIterate[]} schemaTypesToIterate
  */
 function recursiveSchemaArray(schema, callback, schemaTypesToIterate) {
+  if (schemaTypesToIterate.includes(SchemaTypesToIterate.arrays) && schema.type() !== 'array') return;
```
jonaslagoni, line 239: Same as with object.

jonaslagoni (changes requested):
Generally, it looks really good! Added some small comments, also regarding the test. You are testing `AsyncAPIDocument.traverseSchemas` function and not the actual `iterator` functionality.

If you want to keep the current test then there is no reason to have two test locations, you should move the test to here: https://github.com/asyncapi/parser-js/blob/bc7980fb4596ee344834dd35b7ac958326514053/test/models/asyncapi_test.js#L437

Unless you want to change the tests to be a unit test for the iterator functionality alone. 

You can adapt your PR description from `RelatedIssue: 229` to `fixes https://github.com/asyncapi/parser-js/issues/229` then once this PR is merged the issue are closed 🙂

lib/iterators.js
```diff
@@ -183,6 +212,7 @@ function traverseMessage(message, callback, schemaTypesToIterate) {
  * @param {SchemaTypesToIterate[]} schemaTypesToIterate
  */
 function recursiveSchemaObject(schema, callback, schemaTypesToIterate) {
+  if (schemaTypesToIterate.includes(SchemaTypesToIterate.objects) && schema.type() !== 'object') return;
```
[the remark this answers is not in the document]
arjungarg07: Yeah, you're right!

lib/iterators.js
```diff
@@ -55,25 +61,38 @@ const SchemaTypesToIterate = Object.freeze({
...
+    checkSchemas(schema.definitions());
+  }
+  if (schema.dependencies() !== undefined && schemaTypesToIterate.includes(SchemaTypesToIterate.dependencies)) {
+    const props = schema.dependencies();
+    if (props !== null)
+      for (const [prop_, dependenciesSchema] of Object.entries(props)) {
+        const circularProps = schema.circularProps();
+        if (circularProps !== undefined && circularProps.includes(prop_)) continue;
+        traverseSchema(dependenciesSchema, callback, prop_, schemaTypesToIterate);
+      }
```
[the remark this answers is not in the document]
arjungarg07: aah! my bad, forgot to refactor this one 😞

lib/iterators.js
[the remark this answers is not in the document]
jonaslagoni: If you get stuck, feel free to commit the code and ping me 🙂

magicmatatjahu:
@arjungarg07 Hi! Is this PR still valid?

arjungarg07:
@magicmatatjahu I think not, I need to probably reiterate it again since a lot of changes have gone into codebase. Closing it.

Issue #164, opened 2020-08-24
Integrate eslint plugin for jsdoc validation and fix potential issues


derberg:
#### Reason/Context
We rely on JSDoc not only to generate API ref docs but now also to generate TypeScript types. We should take case for the quality of those.

#### Description
Please try answering few of those questions
- integrate https://www.npmjs.com/package/eslint-plugin-jsdoc with eslint
- fix potential issues

derberg:
video about the issue https://youtu.be/Iqs_2BiNEEo?t=1996

arjungarg07:
@derberg Can I pick up this one?

derberg:
@arjungarg07 looks like another good candidate for October, you have work started, some comments recently from the community in the PR. Might be good to open fresh Hacktoberfest PR 😄

pragya-20:
Hi @derberg, I want to pick this up. Can I start working on it?

derberg:
@pragya-20 thanks for pinging

I will close this issue. From the past I know one contributor spent a lot of time trying to fix it. Better just accept what we have and wait for 2.0 that is in active development and will be rewrite this parser to TS, so no more meed of testing jsdoc

PR #305, opened 2021-05-22
refactor: integrate eslint jsdoc validation and fix warnings

arjungarg07:
**Description**
Remaining warnings: 50

**Related issue(s)**
#164


derberg:
@arjungarg07 looks good so far, can you resolve conflicts, I guess that because of it the test workflows were not triggered

arjungarg07:
This must be hard to review. Eye strain warning😬

arjungarg07:
@derberg Do we really need js-doc comments in `test/mixins` files also?

derberg:
@arjungarg07 I don't think we have jsdoc there in tests? what file you mean exactly?

derberg:
@arjungarg07 have a look at sonarcube, some bugs are reported 😅

derberg:
@arjungarg07 and last but not least, I recommend running `npm run lint` locally to trigger these new rules locally and get errors before pushing

arjungarg07:
> @arjungarg07 I don't think we have jsdoc there in tests? what file you mean exactly?

https://github.com/asyncapi/parser-js/blob/974be738947a1752142d77db5c737571163b7ad3/test/mixins/bindings_test.js#L66

https://github.com/asyncapi/parser-js/blob/974be738947a1752142d77db5c737571163b7ad3/test/mixins/description_test.js#L36

Eslint giving warning for such functions: Missing JSDoc comments.

arjungarg07:
> @arjungarg07 and last but not least, I recommend running `npm run lint` locally to trigger these new rules locally and get errors before pushing

Yup, I did and no errors which sonar cloud is giving like `Replace Tag with a constructor function` on pushing are not being reported locally apart from some remaining jsdoc warnings.

derberg:
> Eslint giving warning for such functions: Missing JSDoc comments.

ah you mean jsdoc is missing. As this is not a function in test utils I do not have a strong opinion here, up to you, add jsdoc or just add comment to ignore the rule there.

> Yup, I always run npm run lint but errors which sonar cloud is reporting here like Replace Tag with a constructor function on pushing are not being reported locally apart from some remaining jsdoc warnings.

but have a look at the failing PR testing check, there are many warnings there that cause test fail, I suspect they are the same as the ones from sonar

arjungarg07:
Continuing the discussion here: https://github.com/asyncapi/parser-js/pull/304#discussion_r639491873

@derberg I did according to what this link said and added `@constructor` in the jsdoc but then `jsdoc` lint asks to replace `@constructor` with `@class` and sonar cloud is reporting the same bug. I believe the jsdoc and sonar linting rules are being conflicted.

Screenshot 2021-05-26 at 5 23 07 PM

Also, sorry but I can't understand why these bugs are being reported in the test files as I didn't even make any changes to those files.

Screenshot 2021-05-26 at 5 27 02 PM

arjungarg07:
@derberg I need to ask one more thing that the `eslint/sonarjs` plugin we use is same as `SonarCloud`?

coz I found that the `eslint/sonarjs` plugin doesn't have the same rules as `SonarJs` Cloud and maybe that's the reason when I run `npm run lint` it does not report any bug like `Sonar Cloud` is reporting here. 

Screenshot 2021-05-26 at 5 41 45 PM

derberg:
> I need to ask one more thing that the eslint/sonarjs plugin we use is same as SonarCloud app?

https://github.com/SonarSource/eslint-plugin-sonarjs is from SonarSource, and SonarSource is a company behind SonarCloud. SonarJS is from the same group of people. We use rules that are [recommended](https://github.com/asyncapi/parser-js/blob/master/.eslintrc#L12). Maybe this part is not synchronized between plugin - sonarjs - sonarcloud. I'm not really sure, never had time to go that far to understand it. We never got so many errors all suddenly 😄 

maybe just tune up the jsdoc plugin as looking at https://jsdoc.app/#block-tags looks like `class` is preferred but `constructor` is a synonym, but what they write here https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-check-tag-names I think means that you need to explicitly configure support for synonyms `The following synonyms are also recognized if you set them in tagNamePreference as a key (or replacement):`

lib/models/tag.js
```diff
@@ -8,17 +8,25 @@ const MixinSpecificationExtensions = require('../mixins/specification-extensions
...
  */
 class Tag extends Base {
   /**
-   * @returns {string}
+   * @param {...any} args arguments
+   * @constructor
+   */
+  constructor(...args) {
+    super(...args);
+  }
```
arjungarg07, line 27: @derberg I did the necessary change here to make sure the sonar cloud bug doesn't come up, but I think it's not the possible workaround for this issue. 
Also, I did check that the same bug which is being reported [here](https://sonarcloud.io/project/issues?id=asyncapi_parser-js&pullRequest=305&resolved=false&types=BUG) is also mentioned in the [master branch](https://sonarcloud.io/project/issues?id=asyncapi_parser-js&resolved=false&types=BUG) but we have ignored it. 
perhaps this is a false positive and we need to ignore this check for this repo.

derberg:
@arjungarg07 ok, I think best would be if you forget about SonarCloud atm and just focus on eslint and make sure that there is no warnings when running `npm run lint`

magicmatatjahu:
@arjungarg07 Is PR still valid?

lib/parser.js
```diff
@@ -31,12 +31,12 @@ module.exports = {
...
- * @param {Object} [options.parse] Options object to pass to {@link https://apidevtools.org/json-schema-ref-parser/docs/options.html|json-schema-ref-parser}.
- * @param {Object} [options.resolve] Options object to pass to {@link https://apidevtools.org/json-schema-ref-parser/docs/options.html|json-schema-ref-parser}.
- * @param {Object} [options.applyTraits=true] Whether to resolve and apply traits or not.
+ * @param {string} asyncapiYAMLorJSON An AsyncAPI document in JSON or YAML format.
+ * @param {object} [options] Configuration options.
+ * @param {string} [options.path] Path to the AsyncAPI document. It will be used to resolve relative references. Defaults to current working dir.
+ * @param {object} [options.parse] Options object to pass to {@link https://apidevtools.org/json-schema-ref-parser/docs/options.html|json-schema-ref-parser}.
+ * @param {object} [options.resolve] Options object to pass to {@link https://apidevtools.org/json-schema-ref-parser/docs/options.html|json-schema-ref-parser}.
+ * @param {object} [options.applyTraits=true] Whether to resolve and apply traits or not.
```
BOLT04, line 39: This section seems to be in conflict too. It appears to be the same fields and documentation of the type `ParserOptions` defined here:
https://github.com/asyncapi/parser-js/blob/83e9d4ea774255d71d8aa4594c4a29b02eb96062/lib/parser.js#L32-L37

What do you think about keeping the initial documentation and use the defined `ParserOptions`? Also, this type could be used to document the `options` parameter of the function `dereference`, it seems to have the same format.

lib/parser.js
```diff
@@ -143,6 +143,15 @@ function parseFromUrl(url, fetchOptions, options) {
   });
 }
 
+/**
+ * 
+ * @param {object} refParser refParser 
+ * @param {object} parsedJSON parsed AsyncAPI document
+ * @param {string} initialFormat information of the document was originally JSON or YAML
+ * @param {string} asyncapiYAMLorJSON An AsyncAPI document in JSON or YAML format.
```
BOLT04, line 151: Same comment in [here](https://github.com/asyncapi/parser-js/pull/305/files#r701450076), what do you think @arjungarg07 ?

PR #310, opened 2021-05-31
fix: throw error on unknown schema format 

### LINKED ISSUE #157: Throw meaningfull error when parser of specific schemaFormat is not available
#### Describe the bug
We do not check if there is a parser available for a given custom `schemaFormat` and throw not meaningful error that `PARSERS[String(...)] is not a function`

Hint: this is where we trigger parsers and we need to validate if the function is really there -> https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L179

#### How to Reproduce
Just try to parse the spec with payload defined with some openapi/raml/avro schema. Without registering schema parsers for those schema formats.

#### Expected behavior
In my opinion we need to add new error type and document it in the readme too https://github.com/asyncapi/parser-js#error-types

How about: `schema-parser-not-registered`? with `detail` property that indicates exactly what message has this problem? and of course exact location of the error in the file.

Bonus would be to have an idea how to tell the user that for some schemas there are parsers, and where they can find them

NektariosFifes:
**Description**
Added Error Type in case an unknown/unsupported schema format is provided. Added related test.

**Related issue(s)**

Fixes #157


lib/parser.js
```diff
@@ -190,13 +190,23 @@ async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileForm
...
   const schemaFormat = msg.schemaFormat || DEFAULT_SCHEMA_FORMAT;
 
-  await PARSERS[String(schemaFormat)]({
+  const parser = PARSERS[String(schemaFormat)];
+
+  if (parser === undefined) {
+    throw new ParserError({
+      type: 'schema-parser-not-registered',
+      title: 'The file does not correspond to a supported schema format',
+      detail: 'The asyncApi document cannot be handled by provided parsers.\nIt occurred while trying to parse the following' +
```
magicmatatjahu, line 199:
```suggestion
      detail: `The following message payload "${pathToPayload}" has "${schemaFormat}" schema format which isn't supported by parser instance. Add a missed format parser through the "registerSchemaParser" function.`
```

magicmatatjahu: I'm not sure if we should also give the last note regarding the use of the function @derberg

README.md
```diff
@@ -182,6 +182,7 @@ This package throws a bunch of different error types. All errors contain a `type
 |`impossible-to-register-parser`| None | Registration of custom message parser failed.
 |`schema-validation-errors`| `parsedJSON`, `validationErrors` | Schema of the payload provided in the AsyncAPI document is not valid with AsyncAPI schema format.
 |`fetch-url-error`| None | The URL provided for fetching AsynAPI document is invalid.
+|`schema-parser-not-registered`| detail | The asyncApi document cannot be handled by provided parsers.
```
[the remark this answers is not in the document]
magicmatatjahu: I'm not sure if we should also give the last note regarding the use of the function @derberg

test/parse_test.js
```diff
@@ -604,6 +604,19 @@ it('should apply `x-parser-spec-parsed` extension', async function() {
...
+  const notRegisteredFormatFile = fs.readFileSync(path.resolve(__dirname, './wrong/invalid-schema-format.yaml'), 'utf8');
+  await parser.parse(notRegisteredFormatFile, {path: __filename}).catch((err) => {
+    errorPresent = err;
+  });
+
+  expect(errorPresent.title).equal('The file does not correspond to a supported schema format');
+  expect(errorPresent.type).equal('https://github.com/asyncapi/parser-js/schema-parser-not-registered');
+  expect(errorPresent.detail).equal('The asyncApi document cannot be handled by provided parsers.\n' +
+    'It occurred while trying to parse the following message payload /channels/mychannel/publish/message/payload');
+});
```
magicmatatjahu, line 618: Please, update these lines after updating the detail of error :)

magicmatatjahu (commented):
Great! :) I add some comments, only regarding to description of the new error.

test/wrong/invalid-schema-format.yaml
```diff
@@ -0,0 +1,10 @@
+asyncapi: 2.0.0
```
derberg, line 1: please rename file to `unknown-schema-format` as the file itself is valid

lib/parser.js
```diff
@@ -190,13 +190,27 @@ async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileForm
 
   const schemaFormat = msg.schemaFormat || DEFAULT_SCHEMA_FORMAT;
 
-  await PARSERS[String(schemaFormat)]({
+  //set for the test
+  global.tempPathToPayload = pathToPayload;
```
derberg, line 194: we should not mix code related to test with main library logic. Why do you have it, most of the error message is anyway hardcoded in the test, then this part can be too

test/parse_test.js
```diff
@@ -604,6 +604,19 @@ it('should apply `x-parser-spec-parsed` extension', async function() {
   await expect(parsedSpec.json()[String(xParserSpecParsed)]).to.equal(true);
 });
 
+it('should throw schema-parser-not-registered error on unrecognized schema', async function() {
```
derberg, line 607: please use `checkErrorWrapper` as in other tests in case you need to asses different fields from the error object

lib/parser.js
```diff
@@ -190,13 +190,27 @@ async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileForm
...
+  global.tempSchemaFormat = schemaFormat;
+
+  const parser = PARSERS[String(schemaFormat)];
+
+  if (parser === undefined) {
+    throw new ParserError({
```
derberg, line 200: we are missing `validationErrors` object that among other things contains exact jsonPointer that points to specific part of the document that is broken (containing schema format that is not supported by the parser) an also `startLine` and others

derberg: Have a look how this is done in custom validators https://github.com/asyncapi/parser-js/blob/master/lib/customValidators.js#L58

lib/parser.js
```diff
@@ -190,13 +190,27 @@ async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileForm
...
   const schemaFormat = msg.schemaFormat || DEFAULT_SCHEMA_FORMAT;
 
-  await PARSERS[String(schemaFormat)]({
+  //set for the test
+  global.tempPathToPayload = pathToPayload;
+  global.tempSchemaFormat = schemaFormat;
+
+  const parser = PARSERS[String(schemaFormat)];
+
+  if (parser === undefined) {
```
derberg, line 199: can't we just do `!parser`?

derberg (changes requested):
A good issue for the first contribution! I left some comments as we need more detailed errors and special wrapper to use in tests.

Also, did you think about what I wrote in the linked issue, about nice developer experience and kind of super good error handling here, that if we notice special schema types supported by our already existing schema parsers (for raml dt, avro and openapi) then we drop an error with information where custom schema parser is located?

lib/parser.js
derberg: I like such detailed errors, also would be cool to refer specific implementations if known

lib/parser.js
magicmatatjahu: Also as derberg wrote in my previous comment, please add info which parser should be used if is missed, something like `For avro format, please use @asyncapi/avro-schema-parser` etc. Please make it for officially supported parsers. For custom schema format please don't add info about it :)

derberg: just to clarify and make it easier for you, an example:

1. you parse AsyncAPI document and one of messages has `schemaFormat: 'application/vnd.apache.avro;version=1.9.0'`
2. You detect no custom schema parser is there, and that it is [one of avro](https://github.com/asyncapi/avro-schema-parser/blob/master/index.js#L14-L19)
3. You throw errors like `you shall not pass, no schema parser available that can parse this schema. The good news is that we have one for Avro, just install @asyncapi/avro-schema-parser and use registerSchemaParser to register it with the current parser`

Now, this may be tricky to achieve. Avro mime types supported by avro schema parser are defined in avro-schema-parser, not in parser-js, so it can be tricky to come up with some good idea for it. Let us know if ya needs some help figuring that out

NektariosFifes: just to inform you that during this period due to the gsoc and the university exams i dont have time and i will not be able to add this feature. Is this change urgent right now?

magicmatatjahu: No. We know about university and gsoc, don't worry, we are very happy that you find free time and wanna help us :) You can end it in the next weeks (or maybe also months :P), or if we decide that it's urgent, someone handle that for you, ok? :)

jonaslagoni: @magicmatatjahu as the PR actually solves the issue (maybe not with the extended features you want) does it make sense to split up the issue? So this PR can be merged as is now? Or do you want to have this PR stay open until someone picks it up?

magicmatatjahu: @jonaslagoni Łukasz will be back on Monday, so we will make review  again and then decide, merge or not :)

lib/parser.js
```diff
@@ -190,13 +190,23 @@ async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileForm
...
   const schemaFormat = msg.schemaFormat || DEFAULT_SCHEMA_FORMAT;
 
-  await PARSERS[String(schemaFormat)]({
+  const parser = PARSERS[String(schemaFormat)];
+
+  if (!parser) {
+    throw new ParserError({
+      type: 'schema-parser-not-registered',
+      title: 'The file does not correspond to a supported schema format',
+      detail: `The following message payload "${pathToPayload}" has "${schemaFormat}" schema
```
magicmatatjahu, line 199: As @derberg suggest in first comment in another thread:

> we are missing validationErrors object that among other things contains exact jsonPointer that points to specific part of the document that is broken (containing schema format that is not supported by the parser) an also startLine and others
> 
> Have a look how this is done in custom validators https://github.com/asyncapi/parser-js/blob/master/lib/customValidators.js#L58

We should also add location object where error is occured. Please add it :)

Or we should add it in the followup PR? As I see, we probably don't have necessary data in the function to create this object. @derberg Any thoughts?

magicmatatjahu (commented):
Additional comments. Please refer to them :) Please wait for @derberg review :) And as Jonas suggested, we can merge this PR and add missed features like `location` and information what schema parser we should install in the next PR.

derberg:
The fancy error that points to existing custom parsers implementations can be done in a follow-up @NektariosFifes @jonaslagoni @magicmatatjahu imho.

but the `validationErrors` is default stuff we have in all the errors and in my opinion, this should be done with this PR. Do you need some help?

NektariosFifes:
@derberg  i will try to finish it now that i have more time

magicmatatjahu:
@NektariosFifes Is it still active PR? Will you find time to finish it?

NektariosFifes:
@derberg hey i will start working on it again. is it still needed?

NektariosFifes:
> @NektariosFifes Is it still active PR? Will you find time to finish it?

@magicmatatjahu i will start today

magicmatatjahu:
@NektariosFifes As I see we have another contributor which want to fix that (please check PR https://github.com/asyncapi/parser-js/pull/449) and it is at the end of the implementation, so it's probably better that you don't waste your time on it and just do a review in that PR, what do you think? :)

NektariosFifes:
> @NektariosFifes As I see we have another contributor which want to fix that (please check PR #449) and it is at the end of the implementation, so it's probably better that you don't waste your time on it and just do a review in that PR, what do you think? :)

i will check and probably start another issue. so should i close this?

magicmatatjahu:
Yes, you should. But I would ask for feedback in that other PR, ok? :)

magicmatatjahu:
@NektariosFifes If you run on the master, then you have problem with your system config. If on the mentioned PR, then we have problems in PR.

magicmatatjahu:
v16.13.2

NektariosFifes:
@magicmatatjahu  you have linux?

magicmatatjahu:
MacOS 12

Issue #232, opened 2021-01-18
Handling schema true and false


jonaslagoni:
#### Describe the bug
You can define a schema as `true` or `false` to accept or reject all such as 
```
...
"properties":
   "age": true,
...
```

Now the problem is that our anonymous uid assignment tries to assign a uid to such schemas when its not possible.

This problem triggered the tests to fail for https://github.com/asyncapi/raml-dt-schema-parser/pull/25.
#### Expected behavior
There are a couple of ways we can handle this. 
1. We dont assign schemas which are `true` or `false` uids.
1. We convert `true` or `false` to its sugarcoated values `{}` and `{"not": {}}`

jonaslagoni:
After a lengthy discussion on Slack with the json-schema community my perspective of the `true` and `false` schema's has changed. I saw it more as a `type` then the actual representation of absence and presence of types which is why it is defined as is.

jonaslagoni:
@magicmatatjahu @derberg thoughts? 🤔

derberg:
@jonaslagoni I saw your discussion in slack. If this doesn't represent the data but rather indicates if data can exist, then I guess we should not assign id to it. On the other hand, you would generate a data type for it right, even if it is just false/true?

jonaslagoni:
> On the other hand, you would generate a data type for it right, even if it is just false/true?

Yes, with for example in typescript I would want to generate `any` type for that property. 

I think we shouldn't assign uid to those schemas and just ignore them as you said 😄

magicmatatjahu:
I also ran into this problem 😅 However ids are not the biggest problem at the moment, but the fact that the Base model is not adapted to use the `true/false` schema (see [code](https://github.com/asyncapi/parser-js/blob/master/lib/models/base.js#L11) - we cannot assign to schema `false` value) and also that we [use draft-04](https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L92) in the parser and the [specification itself states](https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#L362) that it supports draft 07 - `true/false` schemas are supported from draft-07.

If you write `true/false` schemas you should have error like:

```txt
[
  "/components/schemas/lightMeasuredPayload/properties/lol should be object at line 156, column 9"
]
```

You can check it in https://playground.asyncapi.io/

derberg:
[...]

PR #311, opened 2021-06-02
fix: support true/false JSON Schemas

magicmatatjahu:
**Description**

As in title. As we support JSON Schema draft-07 as schema, then we should also support `true/false` schemas in JSON Schema of our spec and of course in parser. More info https://github.com/asyncapi/parser-js/issues/232 

**Related issue(s)**
Fixes https://github.com/asyncapi/parser-js/issues/232
Blocked by https://github.com/asyncapi/asyncapi-node/pull/63

### DIFF
--- a/lib/models/schema.js
+++ b/lib/models/schema.js
@@ -17,6 +17,12 @@ const MixinSpecificationExtensions = require('../mixins/specification-extensions
  * @returns {Schema}
  */
 class Schema extends Base {
+  constructor (json) {
+    const isBoolSchema = typeof json === 'boolean';
+    super(isBoolSchema ? {} : json);
+    this.isBoolSchema = isBoolSchema;
+  }
+
   /**
    * @returns {string}
    */
@@ -360,6 +366,13 @@ class Schema extends Base {
    */
   examples() {
     return this._json.examples;
+  }
+
+  /**
+   * @returns {boolean}
+   */
+  isBooleanSchema() {
+    return this.isBoolSchema;
   }
 
   /**
--- a/test/asyncapiSchemaFormatParser_test.js
+++ b/test/asyncapiSchemaFormatParser_test.js
@@ -48,18 +48,6 @@ describe('asyncapiSchemaFormatParser', function() {
         }
       },
       {
-        title: '/channels/mychannel/publish/message/payload/additionalProperties should be object',
-        location: {
-          jsonPointer: '/channels/mychannel/publish/message/payload/additionalProperties',
-          startLine: 13,
-          startColumn: 38,
-          startOffset: offset(252, 13),
-          endLine: 15,
-          endColumn: 15,
-          endOffset: offset(297, 15)
-        }
-      },
-      {
         title: '/channels/mychannel/publish/message/payload/additionalProperties should be boolean',
         location: {
           jsonPointer: '/channels/mychannel/publish/message/payload/additionalProperties',
@@ -108,6 +96,50 @@ describe('asyncapiSchemaFormatParser', function() {
     const parsedInput = JSON.parse(inputString);
 
     expect(async () => await parser.parse(parsedInput)).to.not.throw();
+  });
+
+  it('should handle true/false JSON Schemas', async function() {
+    const inputSpec = {
+      asyncapi: '2.0.0',
+      info: {
+        title: 'Example Spec',
+        version: '1.0.0',
+      },
+      channels: {
+        testChannel: {
+          publish: {
+            message: {
+              payload: {
+                type: 'object',
+                properties: {
+                  trueSchema: true,
+                  falseSchema: false,
+                  normalSchema: {
+                    type: 'string',
+                  }
+                },
+              }
+            }
+          }
+        }
+      },
+      components: {
+        schemas: {
+          testSchema: {
+            type: 'object',
+            properties: {
+              trueSchema: true,
+              falseSchema: false,
+              normalSchema: {
+                type: 'string',
+              }
+            },
+          }
+        }
+      }
+    };
+  
+    expect(async () => await parser.parse(inputSpec)).to.not.throw();
   });
 
   it('should deep clone schema into x-parser-original-payload', async function() {
--- a/test/models/schema_test.js
+++ b/test/models/schema_test.js
@@ -573,6 +573,18 @@ describe('Schema', function() {
     });
   });
 
+  describe('#isBooleanSchema()', function() {
+    it('should return a true when schema is true', function() {
+      const d = new Schema(true);
+      expect(d.isBooleanSchema()).to.be.equal(true);
+    });
+
+    it('should return a true when schema is false', function() {
+      const d = new Schema(false);
+      expect(d.isBooleanSchema()).to.be.equal(true);
+    });
+  });
+
   describe('#isCircular()', function() {
     it('should return a boolean', function() {
       const doc = { 'x-parser-circular': true};


lib/models/schema.js
```diff
@@ -17,6 +17,12 @@ const MixinSpecificationExtensions = require('../mixins/specification-extensions
  * @returns {Schema}
  */
 class Schema extends Base {
+  constructor (json) {
+    const isBoolSchema = typeof json === 'boolean';
+    super(isBoolSchema ? {} : json);
+    this.isBoolSchema = isBoolSchema;
+  }
```
magicmatatjahu, line 24: I have the biggest problem here, how to handle it, because we don't support `true/false` schema from the beginning. This solution both adds boolean schema support, and is also backward compatible. Another solution is to return the schema as boolean, but then we must adjust a lot of functions in another models, which returns array of schemas or single schema. Any thoughts?

test/asyncapiSchemaFormatParser_test.js
```diff
@@ -47,18 +47,6 @@ describe('asyncapiSchemaFormatParser', function() {
...
-        location: {
-          jsonPointer: '/channels/mychannel/publish/message/payload/additionalProperties',
-          startLine: 13,
-          startColumn: 38,
-          startOffset: offset(252, 13),
-          endLine: 15,
-          endColumn: 15,
-          endOffset: offset(297, 15)
-        }
-      },
```
magicmatatjahu, line 61: When we will merge it https://github.com/asyncapi/asyncapi-node/pull/63 then this error won't be throw.

lib/models/schema.js
smoya: I think it makes sense to keep **all** schemas as `Schema` object types. Also, if you return pure boolean types here, checks would need always to be performed as a strict comparison (`===` / `!===`), which sounds pretty error-prone. `schema.isBoolSchema` sounds like a more readable version and also less error-prone since we can make a statement on "all schemas are represented as `Schema` objects.

magicmatatjahu: Yes, you're right :)

jonaslagoni: Yep true, I can see it as benefit as well 👍

lib/models/schema.js
```diff
@@ -17,6 +17,24 @@ const MixinSpecificationExtensions = require('../mixins/specification-extensions
...
+    const isBoolSchema = typeof json === 'boolean';
+
+    let jsonValue = json;
+    if (isBoolSchema) {
+      if (json === true) {
+        // {} is a sugar syntax for `true` Boolean Schema
+        jsonValue = {};
+      } else {
+        // { not: {} } is a sugar syntax for `false` Boolean Schema
+        jsonValue = { not: {} };
```
jonaslagoni, line 30: I would not expect the parser to handle this tbh... It would be confusing when my boolean all of the sudden disappeared from `_json`.

test/asyncapiSchemaFormatParser_test.js
```diff
@@ -110,6 +98,50 @@ describe('asyncapiSchemaFormatParser', function() {
...
+      asyncapi: '2.0.0',
+      info: {
+        title: 'Example Spec',
+        version: '1.0.0',
+      },
+      channels: {
+        testChannel: {
+          publish: {
+            message: {
+              payload: {
```
jonaslagoni, line 112: The payload has to be `true` and `false` to test this properly.

jonaslagoni (changes requested):
Added a few things to discuss.

lib/models/schema.js
magicmatatjahu: I wrote about it in the comment in discussion

> Also in current changes I passed for `true` schema the `{}` model body and for `false` the `{ not: {} }` and by this we have two ways for handling Boolean Schemas.

In some situation it's good, in another not 😄

jonaslagoni: In general, we should never restructure/change the original JSON by changing the core representation of it. It seems wrong from a parser's perspective.

magicmatatjahu: Done, I also removed `booleanValue`, because we can use for it the normal `json` function.

lib/models/schema.js
```diff
@@ -17,6 +17,11 @@ const MixinSpecificationExtensions = require('../mixins/specification-extensions
  * @returns {Schema}
  */
 class Schema extends Base {
+  constructor (json) {
+    super(json === false ? {} : json);
+    this._json = json;
```
jonaslagoni, line 22: Would it not be possible to change the `base` check to check for undefined instead of booleans as well?🤔 Or would that create problems? 🤔 

Setting `_json` twice seems like the wrong approach 😅

jonaslagoni (changes requested):
Much cleaner 👍 Still got a small comment.

lib/models/schema.js
magicmatatjahu: The base ctor is has:

```js
    if (!json) throw new ParserError(`Invalid JSON to instantiate the ${this.constructor.name} object.`);
    this._json = json;
```

I don't want to change it, because handling `false` schemas in base class isn't a good idea. At the moment it's only needed for `Schema` so I pass empty object as workaround

> Setting _json twice seems like the wrong approach 😅

Yep, but it's still "fine" if we make this assignment in child ctor, not in other place :)

magicmatatjahu: Also checking undefined is not good idea - JSON Schema hasn't `undefined` type/value

jonaslagoni: To my understanding, the [base class constructor](https://github.com/asyncapi/parser-js/blob/269206c564cb1271a56771676b8f71601d80ba6c/lib/models/base.js#L11) check, in theory, should only make sure that the received value is not undefined/null 🤔 

Therefore you could change that referenced line from a lazy check to strict:

```js
    if (json === undefined || json === null) throw new ParserError(`Invalid JSON to instantiate the ${this.constructor.name} object.`);
```

That way you can remove `this._json = json;` in this constructor.

jonaslagoni: I might be wrong though 😄

magicmatatjahu: Perhaps I'm sitting in work too long today 😅 Thanks! Changed.

jonaslagoni: haha 😆 That or I have, lets see if others find a problem with this approach 😆

jonaslagoni (dismissed):
Is there a reason why you have updated `dist/bundle.js`? 

Other than that it looks good from my perspective 🙂

lib/models/schema.js
magicmatatjahu: but even if you put `""` or `0` what it's a value in spec? Can we pass something like this in spec itself? Only `false` has some "value", but only related to Schema object.

magicmatatjahu:
@jonaslagoni Probably when I run the integration tests I didn't check the changes for dist... but it has change on every release :) Could you approve again?

lib/models/schema.js
jonaslagoni: Not that I know of no, hence my suggestion. 

The only dilemma I can see could be if the JSON Schema for the specification allowed such values where objects are expected, then it would no longer throw a parser error here. I just don't see the use-case 🤷

magicmatatjahu: I also 😅

types.d.ts
```diff
@@ -1093,4 +1093,3 @@ declare module "@asyncapi/parser" {
      */
     function registerSchemaParser(parserModule: any): void;
 }
-
```
jonaslagoni, line 1096: Not sure if the auto generator will have a problem with this new line. Think you need to remove it.

jonaslagoni (changes requested):
It still detects bundle changes 🙂

types.d.ts
magicmatatjahu: dist and types is generated on every release, so bot will update it :P

magicmatatjahu: fixed :P

test/asyncapiSchemaFormatParser_test.js
```diff
@@ -110,6 +98,71 @@ describe('asyncapiSchemaFormatParser', function() {
...
+      }
+    };
+    
+    let err;
+    try {
+      await parser.parse(inputSpec);
+    } catch (e) {
+      err = e;
+    }
+    expect(err).to.equal(undefined);
```
jonaslagoni, line 163:
```suggestion
    expect(async () => await parser.parse(inputSpec)).to.not.throw();
```

jonaslagoni (changes requested):
Last change I think 😆

jonaslagoni (approved):
I think we can accept that code smell as we already have a similar one and I dont see any other way of shorting that check 👍

PR #312, opened 2021-06-02
fix: specify that parse function input can be also an object

### LINKED ISSUE #308: Confused input argument for parser
We currently define the input document as type [string](https://github.com/asyncapi/parser-js/blob/974be738947a1752142d77db5c737571163b7ad3/lib/parser.js#L34). 

However, both string and object are supported because the input parses through the underlying [toJs](https://github.com/asyncapi/parser-js/blob/974be738947a1752142d77db5c737571163b7ad3/lib/utils.js#L108) function.

What should be considered "correct"? Both "string" and "object" or just "string"? 🤔

vishesh13byte:
**Description**

- ...
- ...
- ...

**Related issue(s)**
  Fixes #308


test/parse_test.js
```diff
@@ -33,6 +33,12 @@ describe('parse()', function() {
     expect(JSON.stringify(result.json())).to.equal(outputJSON);
   });
   
+  it('should parse AsyncAPI document passed as JS object', async function() {
+    const object = JSON.parse(inputYAML);
```
magicmatatjahu, line 37: `inputYAML` as variable name tells you has yaml content, so you cannot use JSON.parse here, because then you have an error:

```
SyntaxError: Unexpected token a in JSON at position 0
```

You should use `inputJSON` which shiuld be parsed before passing to `parser.parse` function.

magicmatatjahu (changes requested):

test/parse_test.js
vishesh13byte: Sorry, I didn't know about that error and thanks for pointing it out. I will fix this👍.

derberg:
please adjust title, fix is good because we will have to publish on npm updated ts types, but describe it maybe something as `fix: specify that parse function input can be also an object` ?

vishesh13byte:
I have made the requested changes, please have a look :). @derberg @magicmatatjahu @jonaslagoni 

Thank you.

test/parse_test.js
```diff
@@ -33,6 +33,12 @@ describe('parse()', function() {
     expect(JSON.stringify(result.json())).to.equal(outputJSON);
   });
   
+  it('should parse AsyncAPI document passed as JS object', async function() {
+    const object = JSON.parse(inputJSON);
+    const result = await parser.parse(object, { path: __filename });
+    expect(JSON.stringify(result.json())).to.equal(outputJSON);
```
magicmatatjahu, line 39: Most likely parsed `inputJSON` is not the same as `outputJSON`. Probably you should to add a new `outputJSON` related to `inputJSON`. And try to test it locally too after pushing new changes :)

vishesh13byte: I have created a  new `outputJSONForObjectInput` for expected output related to `inputJSON` and all the tests have passed 😄. Thank you for your help, please have a look. 
parse_test

vishesh13byte: Also let me know if I should remove the changes in package.json and package-lock.json. They were added because I installed eslint latest version as test and lint scripts were not running on my system for some reason maybe a versin mismatch.

magicmatatjahu: @vishesh13byte Eslint is a dev dependency so it's ok to have latest version. It can stay.

magicmatatjahu (approved):
For me everything is good :) @jonaslagoni @derberg Do you want make another review?

derberg:
@vishesh13byte please revert changes to `dist/bundle.js`. This file should be updated by our pipelines. Other than that is ready to go 🎉

vishesh13byte:
@derberg @magicmatatjahu made those changes , you can have a look. Thank you :).

magicmatatjahu:
@vishesh13byte You have still changes on `dist/bundle`. You can copy code from this file and commit :) https://raw.githubusercontent.com/asyncapi/parser-js/master/dist/bundle.js

vishesh13byte:
@magicmatatjahu I did exactly that 😟, I am trying again.

magicmatatjahu:
@vishesh13byte It's probably a new line or something else :) After publishing it should be changed anyway.

magicmatatjahu (approved):

magicmatatjahu:
@vishesh13byte Thanks for contribution! :)

magicmatatjahu:
@all-contributors please add @vishesh13byte for test

PR #322, opened 2021-06-24
Update @asyncapi/specs (asyncapi-node)

fmvilas:
Since we're missing automation here, I'm bumping the version manually myself.


fmvilas:
Nvm, it was already done.

PR #324, opened 2021-06-24
feat: assigning schema id to parameter schemas in components

jonaslagoni:
**Description**
This PR fixes an issue where we assign parameters wrong ids.

**Related issue(s)**
Fixes https://github.com/asyncapi/parser-js/issues/325

### DIFF
--- a/lib/anonymousNaming.js
+++ b/lib/anonymousNaming.js
@@ -20,7 +20,7 @@ function assignNameToComponentMessages(doc) {
 /**
  * Assign ids based on parameter keys.
  * 
- * @param {{[key:string]: Schema}} parameterObject 
+ * @param {Record<string,Schema>} parameterObject 
  */
 function assignIdToParameters(parameterObject) {
   for (const [parameterKey, parameter] of Object.entries(parameterObject)) {
--- a/lib/anonymousNaming.js
+++ b/lib/anonymousNaming.js
@@ -49,7 +49,8 @@ function assignUidToComponentSchemas(doc) {
 }
 
 /**
- * Assign uid to component parameters
+ * Assign uid to component parameters schemas
+ * 
  * @private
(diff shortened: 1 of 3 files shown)


KhudaDad414:
Well, that was fast. :)

lib/anonymousNaming.js
```diff
@@ -45,6 +47,22 @@ function assignUidToComponentSchemas(doc) {
...
+ * Assign uid to component parameters schemas
+ * 
+ * @private
+ * @param {AsyncAPIDocument} doc 
+ */
+function assignUidToComponentParameterSchemas(doc) {
+  if (doc.hasComponents()) {
+    for (const [parameterKey, parameter] of Object.entries(doc.components().parameters())) {
+      if (parameter.schema()) {
+        parameter.schema().json()[String(xParserSchemaId)] = parameterKey;
```
derberg, line 61: feels like this entire loop is the same as in line 29, just the place where param is taken from is different. Can we do something about it?

derberg (changes requested):
Probably title should be something like `feat: assigning schema id to parameter schemas in components` and the other fix could be in description

lib/anonymousNaming.js
jonaslagoni: Added reusability 😄

lib/anonymousNaming.js
```diff
@@ -17,6 +17,19 @@ function assignNameToComponentMessages(doc) {
   }
 }
 
+/**
+ * Assign ids based on parameter keys.
+ * 
+ * @param {Record<string,Schema>} parameterObject
```
derberg, line 23:
```suggestion
 * @private
 * @param {Record<string,Schema>} parameterObject 
```

derberg (approved):

PR #327, opened 2021-06-24
fix: use proper json schema and tests to point to 2.1

derberg:
I had to make fixes to PR from Laurent as like in case of `asyncapi-node` I did not notice changes were against 2.1

anyway, I'm confused about one thing, should these be changed too?:
- https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L12
- https://github.com/asyncapi/parser-js/blob/master/lib/asyncapiSchemaFormatParser.js#L44-L51

or should now the default schema depend on the version of the spec user uses, so if 2.0 then 2.0 and if 2.1 then 2.1?

Thoughts?

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -284,3 +284,19 @@ utils.setNotProvidedParams = (variables, val, key, notProvidedChannelParams, not
         : missingChannelParams);
   }
 };
+
+/**
+ * returns default schema format for a given asyncapi version
+ *
+ * @function getDefaultSchemaFormat
+ * @private
+ * @param  {String} asyncapiVersion AsyncAPI spec version. 
+ */
+utils.getDefaultSchemaFormat = (asyncapiVersion) => {
+  const schemaMatrix = {
+    '2.0.0': 'application/vnd.aai.asyncapi;version=2.0.0',
+    '2.1.0': 'application/vnd.aai.asyncapi;version=2.1.0',
+  };
+
+  return schemaMatrix[String(asyncapiVersion)];
+};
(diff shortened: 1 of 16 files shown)


fmvilas:
I think it should depend on the schema version the user is using. It should have always been like this but since we only had one version (1.x were not supported) we didn't care.

derberg:
@fmvilas https://github.com/asyncapi/parser-js/blob/master/lib/models/message-traitable.js#L55 is a bug right? shouldn't it be `return this._json.schemaFormat;` ?

fmvilas:
🤔 You caught me off-guard here. I'd say yes, it's a bug. I don't recall why it's like this.

derberg:
@fmvilas ok, fixed with the last commit. What do you think about such an approach for default schema?

lib/utils.js
```diff
@@ -284,3 +284,19 @@ utils.setNotProvidedParams = (variables, val, key, notProvidedChannelParams, not
...
+ * @private
+ * @param  {String} asyncapiVersion AsyncAPI spec version. 
+ */
+utils.getDefaultSchemaFormat = (asyncapiVersion) => {
+  const schemaMatrix = {
+    '2.0.0': 'application/vnd.aai.asyncapi;version=2.0.0',
+    '2.1.0': 'application/vnd.aai.asyncapi;version=2.1.0',
+  };
+
+  return schemaMatrix[String(asyncapiVersion)];
```
fmvilas, line 301: Why not just this?

```suggestion
  return `application/vnd.aai.asyncapi;version=${asyncapiVersion}`;
```

derberg: I somehow thought matrix looks cleaner 🤷🏼‍♂️ but also fine to do it this way

fmvilas (approved):

PR #351, opened 2021-09-02
fix: specify arguments with ParserOptions type as optionals

### LINKED ISSUE #350: Parser options should be optional in TypeScript
#### Describe the bug
Currently, the parser argument is required, even though it is optional.

https://github.com/asyncapi/parser-js/blob/251f46d6cc988e88d863b31320f6e1d31b6196ee/lib/parser.js#L43 should be defined as optional.

magicmatatjahu:
**Description**

Specify (in TS types) arguments with ParserOptions type as optionals.

image

**Related issue(s)**
Fix #350


jonaslagoni (approved):

magicmatatjahu:
@jonaslagoni Thanks and sorry! 😅

PR #364, opened 2021-09-16
feat: added servers field to channel object

GeraldLoeffler:
**Description**

JS parser changes to support a new `servers` field on the `channel` object.

Relates to PR https://github.com/asyncapi/spec/pull/531

**Related issue(s)**

https://github.com/asyncapi/spec/issues/244
https://github.com/asyncapi/spec/issues/255
https://github.com/asyncapi/spec/issues/475

### DIFF
--- a/lib/customValidators.js
+++ b/lib/customValidators.js
@@ -5,7 +5,8 @@ const {
   groupValidationErrors,
   tilde,
   parseUrlQueryParameters,
-  setNotProvidedParams
+  setNotProvidedParams,
+  getUnknownServers
 } = require('./utils');
 const validationError = 'validation-errors';
 
@@ -298,7 +299,8 @@ function isSrvrSecProperArray(schemaType, specialSecTypes, secObj, secName) {
 }
 
 /**
- * Validates if parameters specified in the channel have corresponding parameters object defined and if name does not contain url parameters
+ * Validates if parameters specified in the channel have corresponding parameters object defined and if name does not contain url parameters.
+ * Also validates that all servers listed for this channel are declared in the top-level servers object.
  *
  * @private
  * @param  {Object} parsedJSON parsed AsyncAPI document
@@ -313,11 +315,13 @@ function validateChannels(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
   const chnlsMap = new Map(Object.entries(chnls));
   const notProvidedParams = new Map(); //return object for missing parameters
   const invalidChannelName = new Map(); //return object for invalid channel names with query parameters
+  const unknownServers = new Map(); //return object for server names not declared in top-level servers object
 
   chnlsMap.forEach((val, key) => {
     const variables = parseUrlVariables(key);
     const notProvidedChannelParams = notProvidedParams.get(tilde(key));
     const queryParameters = parseUrlQueryParameters(key);
+    const unknownServerNames = getUnknownServers(parsedJSON, val);
 
     //channel variable validation: fill return obeject with missing parameters
     if (variables) {
@@ -333,6 +337,11 @@ function validateChannels(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
     //channel name validation: fill return object with channels containing query parameters
     if (queryParameters) {
       invalidChannelName.set(tilde(key), queryParameters);
+    }
+
+    //server validatoin: fill return object with unknown server names
+    if (unknownServerNames.length > 0) {
+      unknownServers.set(tilde(key), unknownServerNames);
     }
   });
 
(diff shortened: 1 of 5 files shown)


test/models/channel_test.js
```diff
@@ -27,15 +30,64 @@ describe('Channel', function() {
...
+    it('should return a boolean indicating if the channel has a servers list', function() {
+      const d1 = new Channel(jsWithServers);
+      const d2 = new Channel(jsWithoutServers);
+      expect(d1.hasServers()).to.equal(true);
+      expect(d2.hasServers()).to.equal(false);
+    });
+  });
+
+  describe('#servers()', function() {
+    it('should return an array of String server names or an empty array', function() {
```
magicmatatjahu, line 52: I would prefer to split this tests to two separates one, first for `jsWithServers` and second for `jsWithoutServers`, but only if you want.

magicmatatjahu (approved):
Very fast 😄  Awesome 👏🏼  Only one comment, but only if you want split tests, but you don't have to.

GeraldLoeffler:
@magicmatatjahu : i've now split the servers() test as requestes

magicmatatjahu:
@GeraldLoeffler We only need the PR for `servers` in the `json-schemas` repo :)

derberg (changes requested):
@magicmatatjahu @GeraldLoeffler basic functionality is delivered but we are missing the validation part. We need to validate that list of provided servers is actually declared in the Servers Object

btw, the change in JSON schema is merged and RC released so please use it in this PR -> `2.9.0-2021-09-release.2`

magicmatatjahu:
> basic functionality is delivered but we are missing the validation part. We need to validate that list of provided servers is actually declared in the Servers Object

Yeah... I forgot about that 🤦🏼  

@GeraldLoeffler We have to add a function [in this file](https://github.com/asyncapi/parser-js/blob/master/lib/customValidators.js), which would check if the server names match the server names in the AsyncAPI document, and then execute this function in this function -> https://github.com/asyncapi/parser-js/blob/master/lib/parser.js#L178 If you have any problems I'm here to help, even if you don't have time this week or it's too much for you let me know, then we'll work something out together.

lib/models/channel.js
```diff
@@ -42,6 +42,31 @@ class Channel extends Base {
...
+  /**
+   * @returns {String[]}
+   */
+  servers() {
+    if (!this._json.servers) return [];
+    return this._json.servers;
+  }
+
+  /**
+   * @returns {String}
```
magicmatatjahu, line 61: Please also add parameter annotation:

```suggestion
   * @param {number} index - Index of the server.
   * @returns {String}
```

magicmatatjahu (changes requested):

lib/models/channel.js
GeraldLoeffler: @magicmatatjahu : done

GeraldLoeffler:
@magicmatatjahu : i habe implemented the requested custom validator (`that list of provided servers is actually declared in the Servers Object`). I think it's fairly complete, also in terms of tests (although more test cases could be defined).

lib/customValidators.js
```diff
@@ -351,12 +360,17 @@ function validateChannels(parsedJSON, asyncapiYAMLorJSON, initialFormat) {
     asyncapiYAMLorJSON,
     initialFormat
   );
-  const allValidationErrors = parameterValidationErrors.concat(
-    nameValidationErrors
+  const serverValidationErrors = groupValidationErrors(
+    'channels',
+    'channel does not have a corresponding server object for',
```
magicmatatjahu, line 365:
```suggestion
    'channel is defined for servers that are not defined in the document',
```

test/customValidators_test.js
```diff
@@ -831,7 +831,77 @@ describe('validateChannel()', function () {
...
+    const parsedInput = JSON.parse(inputString);
+
+    const expectedErrorObject = {
+      type: 'https://github.com/asyncapi/parser-js/validation-errors',
+      title: 'Channel validation failed',
+      parsedJSON: parsedInput,
+      validationErrors: [
+        {
+          title:
+            'jmsQueue channel does not have a corresponding server object for: unknownServer',
```
magicmatatjahu, line 885:
```suggestion
            'jmsQueue channel is defined for servers that are not defined in the document: unknownServer',
```

magicmatatjahu (changes requested):
Overall good, only title of error isn't correct, so I added suggestion :)

magicmatatjahu (approved):

magicmatatjahu:
@GeraldLoeffler Thanks for pushing suggestions!
@derberg Could you check again?

test/customValidators_test.js
```diff
@@ -831,7 +831,77 @@ describe('validateChannel()', function () {
...
+    const parsedInput = JSON.parse(inputString);
+
+    const expectedErrorObject = {
+      type: 'https://github.com/asyncapi/parser-js/validation-errors',
+      title: 'Channel validation failed',
+      parsedJSON: parsedInput,
+      validationErrors: [
+        {
+          title:
+            'jmsQueue channel is defined for servers that are not defined in the document: unknownServer',
```
derberg, line 885: the final message is confusing, it should  say something like `jmsQueue channel contains servers that are not on the main servers list in the root of the document: unknownServer`

derberg (commented):
one comment for error message
also please add test case that shows errors for 2 different channels, and one of them has 2 invalid servers

the rest is perfect

test/customValidators_test.js
magicmatatjahu: I would prefer without the word `main`:

```
jmsQueue channel contains servers that are not on the servers list in the root of the document: unknownServer
```

magicmatatjahu: @derberg I added this test locally but I don't have permission to push changes. Can we merge the current PR and I will quickly add a follwup PR? Gerald wrote that he would not have much time until the end of the week.

derberg (approved):

PR #386, opened 2021-10-07
feat: add validation for duplicate tags

### LINKED ISSUE #143: Add validation of duplicated tags
#### Reason/Context
We need to validate if users did not provide 2 tags with the same name on the same level. Tags can be provided in many objects, we should avoid duplicates only in a given object not across the whole document.

#### Description
- I don't think we can do it with JSON Schema. I found something like https://groups.google.com/forum/#!topic/json-schema/z34YqedG-1s but it won't work for us as we use ` "additionalProperties": false`
> https://json-schema.org/understanding-json-schema/reference/combining.html
Unfortunately, now the schema will reject everything. This is because the Properties refers to the entire schema. And that entire schema includes no properties, and knows nothing about the properties in the subschemas inside of the allOf array.
This shortcoming is perhaps one of the biggest surprises of the combining operations in JSON schema: it does not behave like inheritance in an object-oriented language. There are some proposals to address this in the next version of the JSON schema specification.

- we need to add a custom validation of tags wherever tags are possible
- example files that have duplicates but the parser doesn't throw an error:
  - https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Operation%20Object/invalid-duplicate-tags.yaml
  - https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/AsyncAPI%20Object/invalid-duplicate-tags.yaml
  - https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Message%20Object/invalid-duplicate-tags.yaml
  - https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Message%20Trait%20Object/invalid-duplicate-tags.yaml
  - https://github.com/asyncapi/tck/blob/master/tests/asyncapi-2.0/Operation%20Trait%20Object/invalid-duplicate-tags.yaml

BOLT04:
**Description**

- Added a custom validator called `validateTags` in `parser.js` to make sure there are no duplicate tags in the AsyncAPI doc.
- Added 5 tests, one for each object the `tags` could be on: root, operations, operation traits, channels, messages and message traits.

**Questions**
I have some questions about the changes I made, so I opened a draft PR to check the bullet points with the community first 🙂: 
- I noticed there are some typos, for example on the `customValidators.js` file:
https://github.com/asyncapi/parser-js/blob/9ada9bf196bd6112fc272a9aaf4f06db0b398510/lib/customValidators.js#L326

@derberg @magicmatatjahu would it be okay to make a commit fixing these in this PR, or should it be separate? 

- I also noticed the eslint configured options was targeting ES2018, so to test locally I changed to `ecmaVersion: 2020`. Is this acceptable or should I change the implementation to not use the [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator?
- Is this considered a feature or bug fix? I wasn't sure, but I used `feat` since the issue had the label `enhancement`. But if the commits are wrong we could simply squash and change the commit message 🙂 
- The documentation for the other custom validator functions don't link to the `ParserError` type. It's a small touch, but should we update the JSdoc of all functions to have `{@link ParserError}`?
- The other validate functions had the `@private` mark, for example, `validateChannels`. From my understanding, if the function is exported then it's publicly available in the library, right? I didn't include the private mark because I thought the function is public, if this was a wrong assumption I can make a commit with the fix 👍 

Any help on these points is much appreciated 😃. 

**Related issue(s)**
Closes #143


BOLT04:
Thanks for all the feedback @magicmatatjahu, I'll make the requested changes 👍

BOLT04:
Hi @magicmatatjahu, could you review the PR again 🙂?

magicmatatjahu (dismissed):
LGTM! @BOLT04 Thanks for contribution and patience! :) I don't see any problems, but if someone reports it, I will fix it myself :)

magicmatatjahu:
@BOLT04 There are some problems in your tests, could you check it out?

BOLT04:
Hi @magicmatatjahu, I ran the tests in a windows maching with the updated branch and they are all passing: https://recordit.co/ElSMW3KcIK
Could you help me figure out why it's failing in CI? It seems to be all the offsets so is it possible in ubuntu those values change? If that is the case, how could we fix it?

magicmatatjahu:
@BOLT04 Sorry, I missed your update in notifications 😅  Hmm. the problem is with `offset` function which you used in tests - https://github.com/asyncapi/parser-js/blob/master/test/testsUtils.js#L21 - and I think that `EOL.length` have in your machine different value than default one in CI, and by this tests on your computer passed and on CI failed. In other tests we also use this function and on both mac, windows and linux the tests pass. Could you add permission for me to your branch to change it myself so you don't have to - I don't want take your time? :)

BOLT04:
@magicmatatjahu thank you so much for your help 🙂, I'll give you permissions 👍

BOLT04:
@magicmatatjahu did you receive the invite for the repository?

magicmatatjahu (approved):

magicmatatjahu:
@BOLT04 Hi! Sorry for late response but I took a few days off last week because of a national holiday. I have fixed the errors and will be merging. Thank you again for your contributions and your patience :)

PR #406, opened 2021-11-10
fix: memory leak of AJV

magicmatatjahu:
**Description**

More info https://github.com/asyncapi/studio/issues/171

I checked and it's true. The longer you use the studio, the more memory it needs and the stackoverflow occurs after a while. The problem is not only on the side of Parser, but memory leak mainly occurs in this package. Every `parse` function call creates new AJV instance and creates new validating function for given specification version. Additionally every parser call for AsyncAPI Schema Object also creates new instance of AJV and validating function. This PR improves speed and memory usage of the parser.

**Related issue(s)**
Part of https://github.com/asyncapi/studio/issues/171


lib/parser.js
```diff
@@ -175,6 +176,22 @@ async function handleCircularRefs(refParser, parsedJSON, initialFormat, asyncapi
...
+ */
+function getValidator(version) {
+  let validate = ajv.getSchema(version);
+  if (!validate) {
+    ajv.addSchema(asyncapi[String(version)], version);
+    validate = ajv.getSchema(version);
+  }
+  return validate;
+}
+
```
jonaslagoni, line 194: Is there a reason you dont use the same function, instead of duplicating it?

jonaslagoni (commented):
Just encountered the same bug with Studio 😆 

Would really like to get a test for this 🧐 Not sure how to achieve it though 🤔 Any thoughts?

lib/parser.js
magicmatatjahu: > Would really like to get a test for this 🧐 Not sure how to achieve it though 🤔 Any thoughts?

I only see the way to do the integration tests and test the memory consumption for e.g. 1000 iteration of parsing.

> Is there a reason you dont use the same function, instead of duplicating it?

Eeee 😅  Do you have in mind separate function for spec and separate for schema? I thought about that, but I would like to have separate AJV instance for that, because validation of schemas is done as custom operation.

jonaslagoni:
@magicmatatjahu do you see a way to test this? I.e. test that the instance IS the same between calls 🤔

magicmatatjahu:
@jonaslagoni I think about it, but I don't want to export that function from module... there is a an option to create separate module, only for AJV logic, but I have problem with function for schema instance, because there is a custom logic to retrieve JSON Schema for that.

jonaslagoni (dismissed):
👍

magicmatatjahu:
@jonaslagoni Ok I added tests for memory usage. Could you look again?

test/parse_test.js
```diff
@@ -646,6 +646,18 @@ it('should fail on invalid examples', async function() {
...
 });
 
+describe('memory usage', function () {
+  it('should use this same instance of validation function in each call', async function() {
+    const asyncapi = fs.readFileSync(path.resolve(__dirname, './good/asyncapi-complex-schema.yml'), 'utf8');
+
+    for (let i = 0, l = 1000; i < l; i++) {
+      await parser.parse(asyncapi);
+      const used = process.memoryUsage().heapUsed / 1024 / 1024;
+      expect(used < 60).to.equal(true); // less than 60 MB
```
jonaslagoni, line 656: Should this not be tested outside of the for loop? 🤔 

Also, do you think this is reliable, in the long run? 🧐 What if other processes spikes during the test, etc? It would be rather vulnerable to false positives right?

magicmatatjahu: > Should this not be tested outside of the for loop? 🤔

No, because each call (without changes in PR) should increase memory and this test checks if memory usage is within a certain limit.

> Also, do you think this is reliable, in the long run? 🧐 What if other processes spikes during the test, etc? It would be rather vulnerable to false positives right?

`process.memoryUsage()` checks memory usage per process, so yes we can have the possibility that the memory usage will increase, that's why I gave this 60mb as +-10mb reserve - maximum usage is within 50mb +-1mb

jonaslagoni: I don't really like this approach with checking memory, but any other solution would be too complex... Lets hope we will never see false positives 🤞

jonaslagoni (approved):
👍

jonaslagoni (approved):
LGTM, third time is the charm 🤣

magicmatatjahu:
@jonaslagoni Should be fine now :) Could you check again? 😆

jonaslagoni (approved):
👍

PR #418, opened 2021-12-06
docs: update stringify docs

magicmatatjahu:
**Description**

Update stringify docs:
- add example of stringified spec
- add info about prefix in JSON Pointers

**Related issue(s)**
See comment https://github.com/asyncapi/html-template/pull/286#issuecomment-986660092

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,9 +223,15 @@ Parser dereferences all circular references by default. In addition, to simplify
 
 Converting a parsed document to a string may be necessary when saving the parsed document to a database, or similar situations where you need to parse the document just once and then reuse it.
 
-For that, the Parser supports the ability to stringify a parsed AsyncAPI document through the static `AsyncAPIDocument.stringify(...parsedDoc)` method. This method differs from the native `JSON.stringify(...json)` implementation, in that every reference that occurs (at least twice throughout the document) is converted into a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path.
+For that, the Parser supports the ability to stringify a parsed AsyncAPI document through the static `AsyncAPIDocument.stringify(...parsedDoc)` method. This method differs from the native `JSON.stringify(...json)` implementation, in that every reference that occurs (at least twice throughout the document) is converted into a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path with `$ref:` prefix, like:
+
+```json
+{
+  "foo": "$ref:$.some.path.to.the.bar"
+}
+```
 		
-To parse a stringified document into an AsyncAPIDocument instance, you must use the static `AsyncAPIDocument.parse(...stringifiedDoc)` method. It isn't compatible with the native `JSON.parse()` method.
+To parse a stringified document into an AsyncAPIDocument instance, you must use the static `AsyncAPIDocument.parse(...stringifiedDoc)` method. It isn't compatible with the native `JSON.parse()` method. It replaces the given references pointed by [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path with `$ref:` prefix to the original objects.
 
 A few advantages of this solution:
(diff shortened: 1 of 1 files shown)


derberg (approved):
lgtm but better if @alequetzalli have a look, especially that I'm not sure of this part: `prefix to the original objects`

README.md
```diff
@@ -223,9 +223,15 @@ Parser dereferences all circular references by default. In addition, to simplify
 
 Converting a parsed document to a string may be necessary when saving the parsed document to a database, or similar situations where you need to parse the document just once and then reuse it.
 
-For that, the Parser supports the ability to stringify a parsed AsyncAPI document through the static `AsyncAPIDocument.stringify(...parsedDoc)` method. This method differs from the native `JSON.stringify(...json)` implementation, in that every reference that occurs (at least twice throughout the document) is converted into a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path.
+For that, the Parser supports the ability to stringify a parsed AsyncAPI document through the static `AsyncAPIDocument.stringify(...parsedDoc)` method. This method differs from the native `JSON.stringify(...json)` implementation, in that every reference that occurs (at least twice throughout the document) is converted into a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path with `$ref:` prefix, like:
```
alequetzalli, line 226:
```suggestion
For that, the Parser supports the ability to stringify a parsed AsyncAPI document through the static `AsyncAPIDocument.stringify(...parsedDoc)` method. This method differs from the native `JSON.stringify(...json)` implementation, in that every reference that occurs (at least twice throughout the document) is converted into a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path with a `$ref:` prefix:
```

README.md
```diff
@@ -223,9 +223,15 @@ Parser dereferences all circular references by default. In addition, to simplify
...
+
+```json
+{
+  "foo": "$ref:$.some.path.to.the.bar"
+}
+```
 		
-To parse a stringified document into an AsyncAPIDocument instance, you must use the static `AsyncAPIDocument.parse(...stringifiedDoc)` method. It isn't compatible with the native `JSON.parse()` method.
+To parse a stringified document into an AsyncAPIDocument instance, you must use the static `AsyncAPIDocument.parse(...stringifiedDoc)` method. It isn't compatible with the native `JSON.parse()` method. It replaces the given references pointed by [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path with `$ref:` prefix to the original objects.
```
alequetzalli, line 234:
```suggestion
It replaces the given references pointed by the [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) path, with an `$ref:` prefix on the original objects. 
```

I **hope** this is correct? I am trying to make it sound better but unsure if I made it worse

alequetzalli:
@magicmatatjahu  left 2 tiny suggestions 😀👍🏽👍🏽

README.md
magicmatatjahu: @alequetzalli  Hmm, I don't think so that `on` is better that `to`, because it should means that every occurrence of the JSON Pointer is changed to the object where JSON Pointer points, e.g.:

```json
{
  foo: '$ref:$.bar.foo',
  bar: {
    foo: {
      someKey: 'someValue'
    }
  }
}
``` 

so then `foo` key will be changed to the `bar.foo` value, so at the end we will have this:

```json
{
  foo: {
    someKey: 'someValue'
  },
  bar: {
    foo: {
      someKey: 'someValue'
    }
  }
}
```

I hope that it helps :)

magicmatatjahu: @alequetzalli :)

alequetzalli: sorry, I did not see this notification yesterday 🤦‍♀️ 

Got it! I went and edited my previous commit suggestion, and left the original `to` like you had it before. 😀👍🏽👍🏽

alequetzalli (approved):
LGTM 🚀 ❤️✨✨

derberg (approved):

PR #423, opened 2021-12-09
fix: upgrade @asyncapi/specs

jonaslagoni:
**Description**
This PR upgrades the `@asyncapi/specs` library to the new v3.

One important note is, that with the new `@asyncapi/specs` package, the schema for validating JSON Schema's has to be removed in validation libraries. Why?

Long story short, because the AsyncAPI JSON Schema documents are written with a [specific meta schema](https://github.com/asyncapi/spec-json-schemas/blob/5d6ea0361a5b30707afa67a2df28e2805095c10f/schemas/2.3.0.json#L3). That meta schema, is already loaded by Ajv, and when you try to load the bundled AsyncAPI JSON Schema, it now contains that same meta schema, because we use it for payload validation. And Ajv does NOT like to load duplicate schemas and simply throw an error and cannot validate documents.

**Related issue(s)**
Blocked by https://github.com/asyncapi/spec-json-schemas/pull/128


lib/asyncapiSchemaFormatParser.js
```diff
@@ -7,8 +7,9 @@ const cloneDeep = require('lodash.clonedeep');
 const ajv = new Ajv({
   jsonPointers: true,
   allErrors: true,
-  schemaId: 'id',
+  schemaId: 'auto',
   logger: false,
+  validateSchema: true,
```
derberg, line 12: according to docs it is true by default

lib/asyncapiSchemaFormatParser.js
```diff
@@ -80,12 +82,18 @@ function getValidator(version) {
  * 
  * @private
  * @param {Object} asyncapiSchema AsyncAPI specification JSON Schema
+ * @param {Object} version AsyncAPI version.
  * @returns {Object} valid JSON Schema document describing format of AsyncAPI-valid schema for message payload
  */
-function preparePayloadSchema(asyncapiSchema) {
+function preparePayloadSchema(asyncapiSchema, version) {
+  const payloadSchema = `http://asyncapi.com/definitions/${version}/schema.json`;
```
derberg, line 89: why can't we grab it from `@asyncapi/specs` instead of remote?

sorry if that was answered somewhere already. Also, I do not remember really much from times when I created `asyncapiSchemaFormatParser.js`

just dereferencing something in the parser from remote is going to slow down parsing, or?

jonaslagoni: We are using the local one from `@asyncapi/specs` 🙂 

It is only what the reference is called. Ajv automatically matches it with the schema inside `definitions` 🙂

jonaslagoni:
PR is ready

lib/asyncapiSchemaFormatParser.js
derberg: do we need http://asyncapi.com/definitions/ published before we merge?
tbh I'm not 100% sure how these $id work...to complicated for me 😄

lib/asyncapiSchemaFormatParser.js
derberg: @jonaslagoni

lib/asyncapiSchemaFormatParser.js
jonaslagoni: Not at all, the `$id`s when bundled together like so, do not need external access to the schemas 🙂 Because validation tools must look for the reference within before reaching externally. 

We don't need to expose them at all if everyone uses the bundled schemas, but say you want to use this one: https://github.com/asyncapi/spec-json-schemas/blob/master/definitions/2.4.0/asyncapi.json, it would need the schemas exposed on http://asyncapi.com/definitions/ as the schemas are not located within.

Does that make sense?

lib/asyncapiSchemaFormatParser.js
jonaslagoni:
```suggestion
```

derberg:
@jonaslagoni sorry for being so late 😓 please solve conflicts and I can accept

derberg (approved):

PR #430, opened 2021-12-16
feat: allow servers and channels to be fetched from components

smoya:
> Note that this PR is targeting `asyncapi:2022-01-release` branch, so merging it won't release a new version.

**Description**
This PR is part of https://github.com/asyncapi/spec/issues/660.
It allows both `servers` and `channels` to be retrieved from components, meaning the following will be now valid:

```yaml
servers:
  production:
    $ref: '#/components/servers/myserver'
channels:
  some/events:
    $ref: '#/components/channels/myChannel'
components:
  servers:
    myserver:
      url: "http://localhost:5000/ws"
      protocol: ws
  channels:
    myChannel:
      description: "mychannel"
```

**Related issue(s)**
https://github.com/asyncapi/spec/issues/660


lib/models/asyncapi.js
```diff
@@ -164,6 +164,29 @@ class AsyncAPIDocument extends Base {
...
+
+    if (this.hasChannels()) {
+      this.channelNames().forEach(channelName => {
+        channels.set(channelName, this.channel(channelName));
+      });
+    }
+
+    if (this.hasComponents()) {
+      Object.entries(this.components().channels()).forEach(([key, value]) => {
+        // TODO This means that if two channels have the same name (but located in different place), the first one will be overwritten.
```
smoya, line 181: WDYT about this?

magicmatatjahu: Messages as saved with `uid` based on https://github.com/asyncapi/parser-js/blob/209403599a2542258049bb556fd88673a87004da/lib/models/message.js#L17 so you can end with overrides. Hmm 🤔  In the case of channels you should name it with url path, so you can end with duplication but I think that your spec will be invalid - I mean two url for that same event - so here we can accept your solution.

In the case of servers, we should collect them by urls, not by names, because in components and servers they may have different names but they will be the same servers. It would be best to add servers from the `#/servers` section first and then check the url of servers from `#/components/servers` to see if such a server already exists, if it does then don't add the new item to the Map :)

smoya: Having the following example doc:

```yaml
asyncapi: 2.2.0
info:
  title: test
  version: 1.0.0
servers:
  test:
    url: 'test.asyncapi.com'
    protocol: ws
channels:
  /:
    subscribe:
      message:
        $ref: '#/components/messages/testMessage'
components:
  messages:
    testMessage: {}
  channels:
    /: 
      publish:
        message:
          $ref: '#/components/messages/testMessage'
```

Don't you think we should keep both `/` channels instead of overwriting one of them? They can be totally different channels. As channels don't have any extra property such `name` or `id` (related https://github.com/asyncapi/spec/issues/614) I don't think we should overwrite them.

And I would say the same issue happens with servers if we consider a same URL can be used for let's say different protocols, URL field is not enough for making a server unique.

derberg: maybe it should be delegated to the spec, and spec should require unique names for server and channel across document, as it is with correlationId?

@fmvilas thoughts? it is related to https://github.com/asyncapi/spec/issues/660

smoya: Regarding implementation here... From what I see, two servers will be equal **only** if all their properties are equal.  Using only URL field is not enough because you can have two servers with the same URL but different protocols.

Does it make sense to compare the whole object to determine if they should be merged? What should we do if they are not equal but share the same name? Shall we use some kinda prefix? cc @derberg @magicmatatjahu

smoya: > maybe it should be delegated to the spec, and spec should require unique names for server and channel across document, as it is with correlationId?

The point to consider is that this will introduce a breaking change on the spec. I wouldn't discard your idea, but maybe worth to look for a intermediate step that fits into `2.3.0`.

smoya: I've updated the code to handle collisions in both cases (`allChannels`, `allServers`) by:

- servers declared in `servers` take precedence over the ones in components. Same for channels in `channels`.
- In case there is a same server or channel (same base64 hash), we add it to the list by:
  - If there is a name collision, the name of the one declared in components becomes the base64 hash
  - If there is no name collision, it is added as it is.

Added tests as well.

WDYT?

cc @derberg @magicmatatjahu

magicmatatjahu: Hmmm 🤔  

> If there is a name collision, the name of the one declared in components becomes the base64 hash

Maybe we should use only simple `component-` prefix. `base64` hash says little about the name of the component. Take into account that these names may be somehow needed for rendering in templates as e.g. class references or something like that and a hash is not a good solution then.

derberg: If `allChannels` and `allServers` are causing us headaches now + you add them just to be consistent, and not that we see a need for such methods...maybe it is worth just dropping these for now, and solve it properly when we will work on 3.0 ?

derberg:
2 more comments from my side, before you merge, do not forget:
- we do not update bundle on PR, but it will be anyway overwritten during release so maybe we can ignore
- remember to bump spec schema dependency to the release candidate before merging

lib/models/asyncapi.js
fmvilas: Agree with Lukasz. I'd not add them for now. Let's make sure this is considered for v3 and then we can work back a solution for v2.x documents.

smoya: I think that since there is code already that uses `asyncapi.allMessages()` this is needed. See https://github.com/search?q=org%3Aasyncapi+asyncapi.allMessages%28%29&type=code. Maybe those templates don't need that method and they should change their call to `asyncapi.Messages()`? Not sure, maybe you @magicmatatjahu or @derberg can answer that.

In fact, if we should move forward, I realize now maybe I would need to create a PR on those repositories changing all `asyncapi.Channels()` to `asyncapi.AllChannels()` and the same for servers then.

derberg: Can you rephrase your question? I did not get it. 

`allMessages` is indeed needed, so this is why it was added and is used in templates. This should stay as it is. There is a use case for it, as some folks simply do not use channels, and have messages only under `components`.

What we mean here is that Servers and Channels are new. We know we need to access them using regular helpers. So technically speaking, we need to have Servers and Channels part of Components model as you did it in this PR, and we can drop `allChannels` and `allServers` implementation for now, as it is nice to have, and can be added later when we really need it. 

I tried to answer your questions without clarification from your side, but I'm not 100% sure I did it? Thoughts?

smoya: Ok, I didn't think we could access those channels and servers without those new methods but indeed you can always get them via `asyncapi.components()`. Then I agree it is fine to drop those new methods.

> I tried to answer your questions without clarification from your side, but I'm not 100% sure I did it? Thoughts?

Yes you did 👍

smoya: Ok @magicmatatjahu @fmvilas @derberg `allChannels` and `allServers` functions has been removed from my PR now.

smoya:
> * remember to bump spec schema dependency to the release candidate before merging

Done: 
```
"@asyncapi/specs": "asyncapi/spec-json-schemas#2022-01-release",
```

> we do not update bundle on PR, but it will be anyway overwritten during release so maybe we can ignore


We can leave it here right? Or should I remove it?

fmvilas:
> We can leave it here right? Or should I remove it?

I'd remove it as it could be a source of git conflicts.

smoya:
> 

It's done!

smoya:
> > We can leave it here right? Or should I remove it?
> 
> I'd remove it as it could be a source of git conflicts.

It's done!

fmvilas (approved):
🚀

package.json
```diff
@@ -67,7 +67,7 @@
   },
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.6",
-    "@asyncapi/specs": "^2.11.0",
+    "@asyncapi/specs": "asyncapi/spec-json-schemas#2022-01-release",
```
derberg, line 70: @fmvilas are you sure we should grab sources from github?
imho we should either pull a specific version from npm `2.13.0-2022-01-release.1` or just fix to the release candidate tag, so `2022-01-release`

smoya: Question: If we use `2022-01-release` as version, it will point to the same code as latest commit in that branch? So it is the same as I added but pointing to the npm package tag instead of github branch, right? If so, I can change to it as IMHO it also makes sense as you said.

derberg: technically if you point to `2022-01-release` and you do `npm install` then `@asyncapi/specs` version `2.13.0-2022-01-release.1` will be installed, if `2` release candidate is available, then `2.13.0-2022-01-release.2` will be installed. 

so yeah, might be that the best solution is to point to the `2022-01-release` npm tag instead of a specific version each time

smoya: It's done!

derberg (approved):
/dnm

derberg:
@smoya yo, https://github.com/asyncapi/spec-json-schemas/releases/tag/v2.13.0-2022-01-release.3 is out, so you either need to just regenerate package-json or in package.json also point directly to `v2.13.0-2022-01-release.3`, more intuitive and straightforward indicator that this needs to be updated with every release candidate

PR #433, opened 2021-12-27
feat: release for version 2.3.0 of the spec

derberg:

Also, the reason to create this PR a long time before the release is to enable automation (bot keeps upstream branches always up to date with master) that we have in place to regularly update the release branch with whatever is changed in the master branch. So nobody has to do it manually.


package.json
```diff
@@ -67,7 +67,7 @@
   },
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.6",
-    "@asyncapi/specs": "^2.11.0",
+    "@asyncapi/specs": "v2.13.0",
```
magicmatatjahu, line 70: `v` prefix?

magicmatatjahu (approved):

Issue #401, opened 2021-11-03
Implement Parser-API (Intent-Driven)


smoya:
## Reason/Context
Some months ago we created https://github.com/asyncapi/parser-api, an API definition all AsyncAPI parsers should eventually implement.

An introduction can be found as well in this blog post: https://www.asyncapi.com/blog/intent-driven-api

Ideally, we should implement it for the next major **AsyncAPI Spec v3.0.0**. 

@jonaslagoni and I scheduled a Contributor-first meeting on the 6th December 2021 for having a debate around pros/cons of implementing such a thing.

The video is available [here](https://www.youtube.com/watch?v=5Ht2mtRlJc0).
The slides are available [here](https://docs.google.com/presentation/d/1JCNqQKqhNOpHOXZ6z3nPgmtwndTOlj-PVcXG3NaG4Vs/edit?usp=sharing).

The list of pros/cons we ended up collecting (attendees suggested) during the meeting was written [here](https://hackmd.io/AYyrgWMQRXuOKl_VsxUbCA?edit).

Here is the summary of all pros/cons we found out, including the ones mentioned above:

### Introduce intent-driven API

Positive sides:

- User intentions/use cases/desires are expressed intentionally in this API. Meaning users, **without prior knowledge**, will quickly find what they need. This is one of the most important points regarding UX.
    - IDE autocompletion also helps.
- API methods, as being described as intentions, can avoid potential breaking changes.
    - New [publish/subscribe change](https://github.com/asyncapi/spec/issues/618) could mean no breaking change in the parser.
- Adds a layer of abstractions, so you dont need to learn 100% of the specification..
- Can access the raw data that follows the specification.
- Helps everyone (not just coders) better understand the tool. 

Negative sides:

- We add another layer of "abstraction" that users need to learn.
    - You have to know/understand how the intent API maps the spec within their intent functions.
- There can be some different interpretations between the spec and the API which forces us to clarify in the documentation. For example, the fact Operations return a list of Channels in the API, but in the spec a single Operation can only be attached to one single channel. In this case, it is done in that way to avoid possible breaking changes in the future, where we could consider an Operation can be linked to several channels.
- If you provide the raw data, more API's to maintain.

### Introduce parser-api

Positive sides:

- Can ensure compatibility between languages.
- Unify user experience for users
    - Decrease time for zero to hero (based on https://github.com/asyncapi/shape-up-process/issues/58)

Negative sides:

[...]

PR #442, opened 2022-01-12
ci: add next-major branch to the list of branches to release

smoya:
**Description**

Relates to https://github.com/asyncapi/parser-js/issues/401.
This change will allow `next-major` branch to be released on each merge, so users can use and test the new work during the development phase.

**Related issue(s)**

cc @derberg @jonaslagoni


package.json
```diff
@@ -90,6 +90,10 @@
       {
         "name": "2021-09-release",
         "prerelease": true
+      },
+      {
+        "name": "intent-api",
+        "prerelease": true
```
jonaslagoni, line 96: Can CI handle double prerelease @derberg? 🤔

derberg: yes, it is ready as it was required for React component and the 2.0 prerelease phase with `next` branch. 

We only need to agree on bumping strategy. In React for example every single release candidate goes to `html-template`, so on master we had to make sure that whenever you release to `master`, main release is not bumped in `html-template` -> https://github.com/asyncapi/asyncapi-react/blob/master/.github/workflows/bump.yml#L32

package.json
```diff
@@ -90,6 +90,10 @@
       {
         "name": "2021-09-release",
         "prerelease": true
+      },
+      {
+        "name": "intent-api",
```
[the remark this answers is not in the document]
jonaslagoni: `TS`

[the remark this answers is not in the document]
derberg: `any`way 🤷🏼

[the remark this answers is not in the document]
jonaslagoni: Truly, TS is just a far better linter than JSDoc 🤷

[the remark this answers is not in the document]
smoya: > please rename it to `next`. This is standard recommended by `semantic-release` and npm. And release pipeline supports it.
> 
> also from user perspective, at the end it is about next major release. We will not release only intent API in 2.0 of the Parser, there are some other stuff I bet we want to push in

Then I think it's better to use `next-major` as also specified in semantic-release doc.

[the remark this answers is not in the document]
smoya: It's done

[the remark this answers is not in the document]
derberg: but why 😄 if recommended is `next` 😄 and then you need to update global workflow everywhere https://github.com/asyncapi/.github/blob/master/.github/workflows/if-nodejs-release.yml#L12. Why not just `next` and follow the best practices we got from working with react already?
 
Screenshot 2022-01-12 at 18 11 10


@jonaslagoni actually I did not understand your question properly, now saw below after changes from Sergio.
```
      {
        "name": "2021-09-release",
        "prerelease": true
      },
      {
        "name": "next-major",
        "prerelease": true
      }
```
Tbh I'm not sure, we need to check docs of `semantic-release` package to be 100% sure.

package.json
derberg: @jonaslagoni but I guess we do not want to bump it automatically anywhere. I don't see a point at this moment

package.json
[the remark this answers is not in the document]
smoya: `next-major` is standard as well in semantic release. It is intended to be used for that, major releases, which it obviously fits better to our case. We wont block other future releases (minor or patches) may happen before the major.

I didnt checked the ci script you linked until this right moment.

Its not about discarding what you recommend to do randomly 😅. I just read your comment, learned about semantic release `next` branch mechanism, and thats where I found that the `next-major` tag its standard as well as `release`.

package.json
derberg: all templates follow guidelines to set that they are always compatible with generator prior 2.0
so yeah, pushing new Parser would be breaking change, unless you are thinking about release candidates in generator too 🤔

jonaslagoni: I am 🙂

derberg: well, I love the idea!
but 😄 I don't think my bumper supports custom branches 😆

package.json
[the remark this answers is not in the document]
smoya: Btw, I created this PR https://github.com/asyncapi/.github/pull/121 so we add all default semantic-release branches to the release workflow.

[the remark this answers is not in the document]
derberg: @smoya did you maybe have a chance to check docs, of `semantic-release` if it comes to supporting multiple prereleases? I bet it is possible, but just to be 100% sure

package.json
derberg: @smoya ping as not sure if you saw it. I think this one is not just nice to have, but tbh a must-have, otherwise testing new functions in templates will be complicated

package.json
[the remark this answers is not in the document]
smoya: > @smoya did you maybe have a chance to check docs, of `semantic-release` if it comes to supporting multiple prereleases? I bet it is possible, but just to be 100% sure

It does not explicitly say it, however, I understand it happens as I don't see why it wouldn't.
The release workflow will run against just one branch, so the script will check if the branch needs to be pre-released or not based on the `prerelease: true` property in `package.json`.

package.json
derberg: @smoya @jonaslagoni good news, I just released v4 of the bumper action

https://github.com/derberg/npm-dependency-manager-for-your-github-org

Now it supports new prop for config - `base_branch`

So once we merge this PR, and create branch `next-major`, please open another PR against this new release branch and make sure that the bump.yml workflow:
- its version in `next-major` branch will react to commits to `next-major` and not `master`
- add `base_branch: next-major`. This will force bumper to bump Parser in dependants that also have `next-major` branch, and PR with bump will be opened against `next-major`, which means that after merge the release candidate will be published for a given project, if configured

makes sense?

smoya: That's awesome! Yeah makes totally sense. Thanks for the work you did in the bumper action!
Let's do that thing you suggest then!

package.json
```diff
@@ -90,6 +90,10 @@
       {
         "name": "2021-09-release",
         "prerelease": true
+      },
```
jonaslagoni, line 93: Should this not be removed?

smoya: Why?

jonaslagoni: Truee, but kinda "wasteful" to create a new  PR just for that now that you are already changing it 😆

jonaslagoni (approved):

smoya:
@asyncapi-bot why don't you merge my PR? :(

package.json
smoya: It's done here: https://github.com/asyncapi/parser-js/pull/468 @derberg @jonaslagoni

Issue #434, opened 2022-01-05
Support `security` field at Operation level


smoya:
#### Reason/Context
Part of https://github.com/asyncapi/spec/pull/584.

#### Description
Support `security` field containing an array of SecurityRequirement at Operation level.

As this is a good candidate to be in `v2.3.0` release, the PR should be created from and targeting `2022-01-release` branch.

As far as I see, the change we need to do is to add a new function called `security` into the `operation` model. I think we can copy the one from the `server` found in https://github.com/asyncapi/parser-js/blob/2022-01-release/lib/models/server.js#L65-L72

sekharbans-ebay:
> #### Reason/Context
> Part of [asyncapi/spec#584](https://github.com/asyncapi/spec/pull/584).
> 
> #### Description
> Support `security` field containing an array of SecurityRequirement at Operation level.
> 
> As this is a good candidate to be in `v2.3.0` release, the PR should be created from and targeting `2022-01-release` branch.
> 
> As far as I see, the change we need to do is to add a new function called `security` into the `operation` model. I think we can copy the one from the `server` found in https://github.com/asyncapi/parser-js/blob/2022-01-release/lib/models/server.js#L65-L72

Ack. Is simply copying the function ok? Would you in addition like ServerSecurityRequirement (server-security-requirement.js) renamed to 'SecurityRequirement' and reused for both server and operation - or a OperationSecurityRequirement equivalent created?

magicmatatjahu:
@sekharbans-ebay Hi!

> Ack. Is simply copying the function ok? Would you in addition like ServerSecurityRequirement (server-security-requirement.js) renamed to 'SecurityRequirement' and reused for both server and operation - or a OperationSecurityRequirement equivalent created?

as I see `server-security-requirement.js` is empty class, so you can create equivalent the `OperationSecurityRequirement` class. If in the future something additional will appear in the `ServerSecurityRequirement` class, we will create a separate class that will be extended both in `ServerSecurityRequirement` and in `OperationSecurityRequirement` classes, don't worry :)

sekharbans-ebay:
Thanks so much Lokesh!
Hi @smoya, @magicmatatjahu, @derberg  - request review  please

magicmatatjahu:
Handled by https://github.com/asyncapi/parser-js/pull/505

PR #445, opened 2022-01-13
feat: add security field at the operation level

LokeshRishi3:
**Description**
We are using AsyncAPI specifications for publishing the subscriber specific details of our topics and associated payloads for our HTTP push use cases.

Our subscribable channels(or topics) are protected by a set of OAuth scopes (authorization code or client credentials) - that a potential subscriber needs to be aware of.

Here is a sample contract.

We are able to accommodate almost every detail of interest to an integrator/subscriber (THANK YOU) - with the exception of being able to call out the oauth scopes necessary for the subscription to the topic - i.e. the security aspect at the channel or rather channel/operation level.

The current specification is perhaps more aligned with a broker based topology - but ours is a pure push use case (to registered webhooks) and we need an ability to call out any security aspects at the channel operation level.

**Related issue(s)**
Resolves https://github.com/asyncapi/parser-js/issues/434

### DIFF
--- a/API.md
+++ b/API.md
@@ -259,6 +259,7 @@
             * [.hasTraits()](#module_@asyncapi/parser+Operation+hasTraits) ⇒ <code>boolean</code>
             * [.messages()](#module_@asyncapi/parser+Operation+messages) ⇒ <code>Array.&lt;Message&gt;</code>
             * [.message()](#module_@asyncapi/parser+Operation+message) ⇒ <code>Message</code>
+        * [.OperationSecurityRequirement](#module_@asyncapi/parser+OperationSecurityRequirement) ⇐ <code>Base</code>
         * [.PublishOperation](#module_@asyncapi/parser+PublishOperation) ⇐ <code>Operation</code>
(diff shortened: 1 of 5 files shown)


derberg:
@LokeshRishi3 hey, thanks a lot for this one. Are you also coming from ebay with the same use case we discuss in the proposal? So far your PR follows approach when on operation level, only operation security applies, but we are getting conclusions there that server security might apply too, so `server` function should then return not only operation sec requirements but also the ones provided in the associated server. 

Please join discussion https://github.com/asyncapi/spec/pull/584

LokeshRishi3:
@derberg Yes! 🙂
Did you mean:
> ~~server~~ security function should then return not only operation sec requirements but also the ones provided in the associated server.

Also, I realized that I did not add additional tests for the `parse` method and it seems like to add those we first need to publish a new spec with operation-level security in `@asyncapi/specs`. Is that something we will do only after [#584](https://github.com/asyncapi/spec/pull/584) is resolved?

derberg:
yes, you need to wait for the outcome from the `spec` repo discussion

PR #449, opened 2022-01-18
fix: throw meaningful errors

### LINKED ISSUE #157: Throw meaningfull error when parser of specific schemaFormat is not available

ritik307:
**Description**

- Added error messages to `parser.js`

**Related issue(s)**

Fixes #157


README.md
```diff
@@ -196,6 +196,7 @@ This package throws a bunch of different error types. All errors contain a `type
 |`impossible-to-register-parser`| None | Registration of custom message parser failed.
 |`schema-validation-errors`| `parsedJSON`, `validationErrors` | Schema of the payload provided in the AsyncAPI document is not valid with AsyncAPI schema format.
 |`fetch-url-error`| None | The URL provided for fetching AsynAPI document is invalid.
+|`schema-parser-not-registered`| `detail` | At least, one of the schemas in the AsyncAPI document is in a format that the given parser instance doesn't support. Add a missed format parser through the `registerSchemaParser` function.
```
derberg, line 199:
```suggestion
|`schema-parser-not-registered`| `detail` | At least, one of the schemas in the AsyncAPI document is in a format that the given parser instance doesn't support. Add missing parser plugin through the `registerSchemaParser` function.
```

lib/parser.js
```diff
@@ -5,10 +5,24 @@ const asyncapi = require('@asyncapi/specs');
...
+  toJS,
+  findRefs,
+  getLocationOf,
+  improveAjvErrors,
+  getDefaultSchemaFormat,
+  groupValidationErrors,
+} = require('./utils');
 const AsyncAPIDocument = require('./models/asyncapi');
 
+// const DEFAULT_SCHEMA_FORMAT = 'application/vnd.apache.avro;version=1.9.0';
```
derberg, line 25: left over?

lib/parser.js
```diff
@@ -64,49 +78,78 @@ async function parse(asyncapiYAMLorJSON, options = {}) {
...
+        validationErrors: [
+          getLocationOf('/asyncapi', asyncapiYAMLorJSON, initialFormat),
+        ],
       });
     }
 
     if (options.applyTraits === undefined) options.applyTraits = true;
 
-    const refParser = new $RefParser;
+    const refParser = new $RefParser();
```
derberg, line 111: why is it needed?

derberg, line 111: I bet this is breaking tests

derberg (commented):
I think I found your problem, have a look at my comments

I have to say that it is super hard to review this PR because of all the autoformatted changes. Can we do something about it?

ritik307:
@derberg I am not able to figure out what to pass inside `groupValidationErrors` function in `lib/parser.js`

lib/parser.js
```diff
@@ -192,52 +238,104 @@ function getValidator(version) {
...
+      type: 'schema-parser-not-registered',
+      title: 'The specification has an unsupported schema format',
+      detail:
+        'You shall not pass, no schema parser available that can parse this schema. The good news is that we have one for Avro, just install @asyncapi/avro-schema-parser and use registerSchemaParser to register it with the current parser',
+      validationErrors: groupValidationErrors(
+        'servers',
+        'server does not have a corresponding schema object',
+        new Map(),
+        fileFormat,
+        parser
```
derberg, line 308: here you should pass information if original AsyncAPI file passed into parser was YAML or JSON. It is in `fileFormat` variable.

lib/parser.js
```diff
@@ -192,52 +238,104 @@ function getValidator(version) {
...
+    throw new ParserError({
+      type: 'schema-parser-not-registered',
+      title: 'The specification has an unsupported schema format',
+      detail:
+        'You shall not pass, no schema parser available that can parse this schema. The good news is that we have one for Avro, just install @asyncapi/avro-schema-parser and use registerSchemaParser to register it with the current parser',
+      validationErrors: groupValidationErrors(
+        'servers',
+        'server does not have a corresponding schema object',
+        new Map(),
+        fileFormat,
```
derberg, line 307: it should be `originalAsyncAPIDocument`, the oryginal file as string. It is needed because returned error will have to indicate where exactly in the file (line, column) the error is in, so original document is needed. So later in the UI you can highlight the error

lib/parser.js
```diff
@@ -192,52 +238,104 @@ function getValidator(version) {
...
+  const parser = PARSERS[String(schemaFormat)];
+
+  if (!parser) {
+    throw new ParserError({
+      type: 'schema-parser-not-registered',
+      title: 'The specification has an unsupported schema format',
+      detail:
+        'You shall not pass, no schema parser available that can parse this schema. The good news is that we have one for Avro, just install @asyncapi/avro-schema-parser and use registerSchemaParser to register it with the current parser',
+      validationErrors: groupValidationErrors(
+        'servers',
```
derberg, line 304: it is not server, it is `channels`

lib/parser.js
```diff
@@ -192,52 +238,104 @@ function getValidator(version) {
...
+        'server does not have a corresponding schema object',
+        new Map(),
```
derberg, line 306: this is a map of errors, so definitely cannot be empty.

you choose to use `groupValidationErrors`, which is much better from initial implementation. 

Why?

because you will not throw error immediately when it is found. The intention with `groupValidationErrors` is to use it in error when you want to throw it after thing through all messages. 

without such approach, you would get error for one message, then you fix, then error for another message, and fix, and so on, not so nice.

JSDoc say this about the map
```
map of error elements cause the validation error might happen in many places in the document. 
The key should have a path information where the error was found, the value holds information about error element but it is not mandatory
 ```

so the key is the json path pointing to the affected place in the document. Later it ends up in error object under `location.jsonPointer`. You will need to use `tilde` util for sure. 

So from your test example, 

- first you need to make `validateAndConvertMessage` aware of the channel name and operation where validated message is located. 
- Then in this map, the key would be something like `tilde(${channelName})/${operationName}/message/schemaFormat
- and value is, I don't remember but probably broken value 🤷🏼

does it make sense?

File: lib/parser.js [hunk 8]
```diff
@@ -192,52 +238,104 @@ function getValidator(version) {
...
+  const defaultSchemaFormat = getDefaultSchemaFormat(
+    parsedAsyncAPIDocument.asyncapi
+  );
   const schemaFormat = msg.schemaFormat || defaultSchemaFormat;
 
-  await PARSERS[String(schemaFormat)]({
+  const parser = PARSERS[String(schemaFormat)];
+
+  if (!parser) {
+    throw new ParserError({
```
derberg, line 298: when using `groupValidationErrors`, do not throw error immediately. Let parser go through all messages, so they you throw error a the end, with all broken `schemaFormats` from all messages. Look into `customValidators` how they use it.

I think it will be trick, but yeah, try

derberg (changes requested):
I provided a few hints. Sorry, I just realized this code related to validation that I wrote is complicated and provided JSDoc is lame 😞 

Lemme know if you need more help. And please remember to remove all the autoformatted parts before next review as is is really hard to review

ritik307:
@derberg I have made changes as you have suggested but I am still getting errors while running the test and am not sure about the implementation of throwing all the errors at once.😅

lib/parser.js
```diff
@@ -206,11 +210,29 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
 
-async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload) {
+async function validateAndConvertMessage(msg, channelName ,originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload) {
   //check if the message has been parsed before
   if (xParserMessageParsed in msg && msg[String(xParserMessageParsed)] === true) return;
   const defaultSchemaFormat = getDefaultSchemaFormat(parsedAsyncAPIDocument.asyncapi);
   const schemaFormat = msg.schemaFormat || defaultSchemaFormat;
+  const parser = PARSERS[String(schemaFormat)];
+  const errorMap = new Map();
+  errorMap.set(`${tilde(channelName)}/${Operation.name}/${msg}/${schemaFormat}`, schemaFormat);
```
derberg, line 220: `validateAndConvertMessage` is called twice, because in AsyncAPI document, message can be in 2 locations:
- under channel/operation/message -> in this case function is called in an `Object.entries` iteration where you have access to `opName`, so you can actually pass it to `validateAndConvertMessage` instead of doing `Operation.name`
- under components/messages -> in this case there is no operation and error location will not be `${tilde(channelName)}/${Operation.name}/${msg}/${schemaFormat}` but under components like `messages/${msgName}/${schemaFormat}`

so yeah, something I did not explain before is that you need to figure out first if you are dealing with a message on `channels` or `components`. So I would definitely not add `channelName` argument to `validateAndConvertMessage` in the middle of the list of arguments but at the end as option maybe, like that it can be either `{ channelName: "myName" }` or `{ component: true }` and then you can easily check what option is passed and what to do in the code.

What do you think?

ritik307:
Hi @derberg
I got your point and have made some changes kindly look into them.😊😊

derberg (commented):
please convert it to regular PR so we can have tests running in a PR. Also please remove autoformatting also from `test/parse_test.js` file 😅

ritik307:
> please convert it to regular PR so we can have tests running in a PR. Also please remove autoformatting also from `test/parse_test.js` file sweat_smile

Done! 😁

ritik307:
@derberg  Done! 😁

lib/parser.js
```diff
@@ -16,6 +17,9 @@ const PARSERS = {};
 const xParserCircle = 'x-parser-circular';
 const xParserMessageParsed = 'x-parser-message-parsed';
 
+const schemaValidationErrors = [];
```
magicmatatjahu, line 20: My 50 cents, where do you "clear" that errors? As I see in the changed code, that array preserve state between `parse` execution.

ritik307:
```
if (schemaValidationErrors.length) {
        const errors = schemaValidationErrors;
        schemaValidationErrors.length = 0;
        throw new ParserError({
          type: validationError,
          title: 'Channel validation failed',
          parsedJSON,
          validationErrors: errors,
        });
```
@magicmatatjahu Will using the above be a good idea because I can't clear the array after the throw command. Please suggest if any other better solution exists.

ritik307:
@derberg @magicmatatjahu I have implemented the requested changes but failed the `npm run test` and I am not able to figure out how I can resolve the TypeErrors I am getting. Kindly help me out.

magicmatatjahu (changes requested):
I'm sorry, but I won't accept the current solution with the global array `schemaValidationErrors` and the reason is simple: parsing is done in an asynchronous way and if we run the function 100 or even more times in a row it may happen that the array `schemaValidationErrors` will have some data (or will be deleted) at the moments when other independent parsing is done. This can happen in our `server-api` project as well as in `studio`. If we still want to have `schemaValidationErrors` we need create that array in local scope (e.g. at the beginning of the `parser` function) and the pass to the corresponding functions. Global values should be independent.

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+    errorMap.set(`${tilde(options.channelName)}/${name}/${msg}/${schemaFormat}`, schemaFormat);
+  }
+  
+  if (!parser) {
+    schemaValidationErrors.push(new ParserError({
+      type: 'schema-parser-not-registered',
+      title: 'The specification has an unsupported schema format',
+      detail:
+        'You shall not pass, no schema parser available that can parse this schema. The good news is that we have one for Avro, just install @asyncapi/avro-schema-parser and use registerSchemaParser to register it with the current parser',
+      validationErrors: groupValidationErrors(
```
derberg, line 234: please check `validationErrors` in tests too

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+  else{
+    errorMap.set(`${tilde(options.channelName)}/${name}/${msg}/${schemaFormat}`, schemaFormat);
+  }
+  
+  if (!parser) {
+    schemaValidationErrors.push(new ParserError({
+      type: 'schema-parser-not-registered',
+      title: 'The specification has an unsupported schema format',
+      detail:
+        'You shall not pass, no schema parser available that can parse this schema. The good news is that we have one for Avro, just install @asyncapi/avro-schema-parser and use registerSchemaParser to register it with the current parser',
```
derberg, line 233:
```suggestion
        'Parser is not able to parse messages as they are provided in not supported schema. Parser can be extended to parse other schemas, you can also use existing custom schema parsers for Avro, RAML Data Types and OpenAPI Schema. Install one of them, @asyncapi/avro-schema-parser, @asyncapi/openapi-schema-parser or @asyncapi/ramldt-schema-parser and use registerSchemaParser function to register it with the this parser',
```

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+      validationErrors: groupValidationErrors(
+        'channels',
+        'channel does not have a corresponding schema object',
```
derberg, line 236:
```suggestion
        'uses unsupported schema format',
```

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
-async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload) {
+async function validateAndConvertMessage(msg,name ,originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload, options, schemaValidationErrors) {
   //check if the message has been parsed before
   if (xParserMessageParsed in msg && msg[String(xParserMessageParsed)] === true) return;
   const defaultSchemaFormat = getDefaultSchemaFormat(parsedAsyncAPIDocument.asyncapi);
   const schemaFormat = msg.schemaFormat || defaultSchemaFormat;
+  const parser = PARSERS[String(schemaFormat)];
+  const errorMap = new Map();
+  if(options.component){
+    errorMap.set(`messages/${name}/${schemaFormat}`, schemaFormat);
```
derberg, line 222:
```suggestion
    errorMap.set(`messages/${name}`, schemaFormat);  
```

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
   validateOperationId(parsedJSON, asyncapiYAMLorJSON, initialFormat, OPERATIONS);
 
-  await customComponentsMsgOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options);
-  await customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options);
+  await customComponentsMsgOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options, schemaValidationErrors);
+  await customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options, schemaValidationErrors);
 }
 
-async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload) {
+async function validateAndConvertMessage(msg,name ,originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload, options, schemaValidationErrors) {
```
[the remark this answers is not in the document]
derberg: Please change approach with name in general, this is different depending if component or not, so don't pass it under the same variable but also maybe move to options. From usage, to make it clear on variable name level, just let it work something like `name.component` and `name.operation` or `options.opName` and `options.messageName`

test/parse_test.js
```diff
@@ -861,6 +861,18 @@ it('should apply `x-parser-spec-parsed` extension', async function() {
   await expect(parsedSpec.json()[String(xParserSpecParsed)]).to.equal(true);
 });
 
+it('should throw schema-parser-not-registered error on unrecognized schema', async function () {
+  const notRegisteredFormatFile = fs.readFileSync(path.resolve(__dirname, './wrong/unknown-schema-format.yaml'), 'utf8');
+  const errorObject = {
```
derberg, line 866:
```suggestion
  const expectedErrorObject = {
```

test/wrong/unknown-schema-format.yaml
```diff
@@ -0,0 +1,9 @@
+asyncapi: 2.1.0
+info:
+  version: "1.0.0"
+  title: "Test"
+channels:
+  mychannel:
+    publish:
+      message:
+        schemaFormat: "UnknownSchemaFormat"
```
derberg, line 9: extend this test file with also broken message at the `component` level, and also add `subscribe` operation with message that uses `oneOf` and there are 2 messages with wrong format. We need to test all these cases

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
   //check if the message has been parsed before
   if (xParserMessageParsed in msg && msg[String(xParserMessageParsed)] === true) return;
   const defaultSchemaFormat = getDefaultSchemaFormat(parsedAsyncAPIDocument.asyncapi);
   const schemaFormat = msg.schemaFormat || defaultSchemaFormat;
+  const parser = PARSERS[String(schemaFormat)];
+  const errorMap = new Map();
```
derberg, line 220: this whole error map creation is related to `if (!parser) {` only, so better move it there for clarity

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+  await customChannelsOperations(parsedJSON, asyncapiYAMLorJSON, initialFormat, options, schemaValidationErrors);
 }
 
-async function validateAndConvertMessage(msg, originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload) {
+async function validateAndConvertMessage(msg,name ,originalAsyncAPIDocument, fileFormat, parsedAsyncAPIDocument, pathToPayload, options, schemaValidationErrors) {
   //check if the message has been parsed before
   if (xParserMessageParsed in msg && msg[String(xParserMessageParsed)] === true) return;
   const defaultSchemaFormat = getDefaultSchemaFormat(parsedAsyncAPIDocument.asyncapi);
   const schemaFormat = msg.schemaFormat || defaultSchemaFormat;
+  const parser = PARSERS[String(schemaFormat)];
```
derberg, line 219:
```suggestion
  const customSchemaParser = PARSERS[String(schemaFormat)];
```

I think more clear name

lib/parser.js
```diff
@@ -202,15 +207,39 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+        'channels',
+        'channel does not have a corresponding schema object',
+        errorMap,
+        originalAsyncAPIDocument,
+        parser
+      ),
+    }));
+  }
 
   await PARSERS[String(schemaFormat)]({
```
derberg, line 244: this is the reason you tests fail, because even though you are catching errors above, you still go and execute parser that might not exist. This await and also rest of the function should go into `else` so it is called only if `parser` is really there.

also you can call `customSchemaParser` that you have above and not `PARSERS[String(schemaFormat)]` again

derberg (changes requested):
I found why your tests fail, also added some comments to improve naming of variables and text of error messages.

good work so far 🚀

ritik307:
https://github.com/asyncapi/parser-js/pull/449#discussion_r793945425
@derberg  I am not clear with the following change as `errorMap` is also used at lines **222** and **225**

ritik307:
hi @derberg While implementing the requested changes I am confused about some requested changes like 

- https://github.com/asyncapi/parser-js/pull/449#discussion_r793922835 (I am sure whether I have implemented it correctly or not)
- https://github.com/asyncapi/parser-js/pull/449#discussion_r793945425 (I am not clear with the following change as `errorMap` is also used at lines **222** and **225**  )

**Screenshot of error i am currently facing while running `npm run test`**
ss1

lib/parser.js
```diff
@@ -202,28 +206,54 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+      type: validationError,
+      title: 'The document uses a schema format that parser does not support',
+      detail:
+        'Parser is not able to parse messages as they are provided in not supported schema. Parser can be extended to parse other schemas, you can also use existing custom schema parsers for Avro, RAML Data Types and OpenAPI Schema. Install one of them, @asyncapi/avro-schema-parser, @asyncapi/openapi-schema-parser or @asyncapi/ramldt-schema-parser and use registerSchemaParser function to register it with the this parser',
+      validationErrors: groupValidationErrors(
+        'channels',
+        'uses unsupported schema format',
+        errorMap,
+        originalAsyncAPIDocument,
+        customSchemaParser
```
derberg, line 238:
```suggestion
        fileFormat
```

derberg:
@ritik307 

> https://github.com/asyncapi/parser-js/pull/449#discussion_r793945425 (I am not clear with the following change as errorMap is also used at lines 222 and 225 )

sorry,  didn't make myself clear. This is what I meant exactly (notice change in variables names, you had some errors there):

```
  if (!customSchemaParser) {
    const errorMap = new Map();
    if (options.component) {
      errorMap.set(`messages/${options.messageName}`, schemaFormat);  
    } else {
      errorMap.set(`${tilde(options.channelName)}/${options.opName}/message/${schemaFormat}`, schemaFormat);
    }
```

> https://github.com/asyncapi/parser-js/pull/449#discussion_r793922835 (I am sure whether I have implemented it correctly or not)

Sorry, can't figure out which comment you're pointing too

> Failing Tests

the thing is that in your `unknown-schema-format.yaml` you have:
```
      message:
        oneOf:
          - $ref: "#/components/messages/testMessage1"
          - $ref: "#/components/messages/testMessage2"
```

But you did not add anything under components.messages in this document. You should have something like:
```
asyncapi: 2.1.0
info:
  version: "1.0.0"
  title: "Test"
channels:
  mychannel:
    publish:
      message:
        schemaFormat: "UnknownSchemaFormat"
    subscribe:  
      message:
        oneOf:
          - $ref: "#/components/messages/testMessage1"
          - $ref: "#/components/messages/testMessage2"
          
components:
  messages:
    testMessage1:
      schemaFormat: "UnknownSchemaFormat1"
    testMessage2:
      schemaFormat: "UnknownSchemaFormat2"
```

Then extend test `should throw schema-parser-not-registered error on unrecognized schema` to check exactly the error message, including validation errors that indicate the error position in the document (location)


ah, and do not worry about conflicts with bundle.js, you can actually not push any changes to this file. Just resolve conflicts in favor of master

derberg:
Even if you apply all my suggestions it is still not throwing validation errors. 

Looking quickly at code I think the problem is that in `validateAndConvertMessage` you push to `schemaValidationErrors` entire `new ParserError` while you should only push `errorMap`. Then you throw error separately in `customComponentsMsgOperations` and `customChannelsOperations`. Why not throw one aggregated error under `customDocumentOperations` after running custom operations?

ritik307:
> Then you throw error separately in `customComponentsMsgOperations` and `customChannelsOperations`. Why not throw one aggregated error under `customDocumentOperations` after running custom operations? `type`, `title` and `detail` are always the same, just `validationErrors` must list all the errors, in case of your test, there should be array of 3

Sorry for troubling you again and again I have made changes according to my understanding but I am unable to figure this out. 🤕 Kindly help.😭

derberg:
@ritik307 

> Sorry for troubling you again and again I have made changes according to my understanding but I am unable to figure this out. 🤕 Kindly help.😭

no worries, it is not a trivial task, I'm here to help. Sorry for not explaining it better.

you are doing it great, so now you throw errors under `customDocumentOperations` https://github.com/asyncapi/parser-js/pull/449/files#diff-2f45270c7389933f71b380538ff52e106ee5cb66a0b573d9967467dfdbc11c02R212 whis is the way to go IMHO (you only forgot about grouping validation errors, using the helper method), but just throw error only if `schemaValidationErrors` is not empty, because errors should be throw only in case at least one validation error was spotted.

And in my comment, what I meant is, that if you have ☝🏼 then there is no longer a need to throw parser error separately under `customComponentsMsgOperations` and `customChannelsOperations` because `validateAndConvertMessage` is already pushing all errors to  `schemaValidationErrors`. Does it make sense now?

magicmatatjahu:
@ritik307 Łukasz is on 2 weeks vacation, so if you will have any problems, please ping me, not him :)

ritik307:
Got it 👍

derberg:
how is it going here? 😄

ritik307:
@derberg @magicmatatjahu sorry I was caught up with some college work 😅. I have made some changes kindly review them.🙂

lib/parser.js
```diff
@@ -202,28 +206,57 @@ async function customDocumentOperations(parsedJSON, asyncapiYAMLorJSON, initialF
...
+    }));
+    schemaValidationErrors.length=0;
+    throw new ParserError({
+      type: validationError,
+      title: 'The document uses a schema format that parser does not support',
+      detail:'Parser is not able to parse messages as they are provided in not supported schema. Parser can be extended to parse other schemas, you can also use existing custom schema parsers for Avro, RAML Data Types and OpenAPI Schema. Install one of them, @asyncapi/avro-schema-parser, @asyncapi/openapi-schema-parser or @asyncapi/ramldt-schema-parser and use registerSchemaParser function to register it with the this parser',
+      parsedJSON,
+      validationErrors: errors,
+    });
+  }
```
derberg, line 225:
```suggestion
    if (schemaValidationErrors.size) {
    throw new ParserError({
      type: validationError,
      title: 'The document uses a schema format that parser does not support',
      detail: 'Parser is not able to parse messages as they are provided in not supported schema. Parser can be extended to parse other schemas, you can also use existing custom schema parsers for Avro, RAML Data Types and OpenAPI Schema. Install one of them, @asyncapi/avro-schema-parser, @asyncapi/openapi-schema-parser or @asyncapi/ramldt-schema-parser and use registerSchemaParser function to register it with the this parser',
      parsedJSON,
      validationErrors: groupValidationErrors(
        'channels',
        'uses unsupported schema format',
        schemaValidationErrors,
        asyncapiYAMLorJSON,
        initialFormat
      )
    });
  }
```

derberg: This is only half the working code. This will now trigger good errors so you need to adjust tests accordingly

The only thing that doesn't work in this code is line 219 -> `'channels'`. It is hardcoded to channels even though sometimes as you know error might be not in message under channels but also in message under components, so sometimes instead of `'channels'` you must pass `'components'`.  

We need to figure out how to do this....so first please apply my suggestions, fix tests and what should left for us to think about is this channels vs components thingy. I think the only way to solve it will be to refactor `groupValidationErrors` somehow

derberg (changes requested):
hey, left some comments

also please make sure that you always run `npm run lint` before you submit something for review

magicmatatjahu:
@ritik307 Do you have time for further contribution? Any problems/issues?

ritik307:
@magicmatatjahu Sorry for the late response. Yes, I do have time to contribute but not as much as I had before 😔.  As of now, I will only be able to contribute on weekends.

derberg:
@ritik307 no worries, just keep in mind that with this one you are almost done, so do not lose the time that you've already put in this PR. It is a great addition.

ritik307:
Sorry for the delay. I made the changes but didn't forget to push the code. I am still facing some doubts though. Kindly help me out

ritik307:
@derberg Sorry for the miscommunication. I meant I have pushed the code can you review it.

test/parse_test.js
```diff
@@ -861,6 +861,41 @@ it('should apply `x-parser-spec-parsed` extension', async function() {
...
+          endColumn: 1,
+          endOffset: 0,
+          startOffset: 0,
+        }
+      }
+    ],
+  };
+  await checkErrorWrapper(async () => {
+    await parser.parse(notRegisteredFormatFile, { path: __filename });
+  }, errorObject);
```
derberg, line 896:
```suggestion
  }, expectedErrorObject);
```

derberg:
@ritik307 you definitely need to run tests before you push code 😄 
Please have a look and update the tests with proper info

derberg:
@ritik307 you were pretty advanced here, do not want to continue?

magicmatatjahu:
I am closing due to lack of response.

PR #450, opened 2022-01-19
fix: export Schema model

dschwank:
**Description**
So far a `ts-ignore` and a direct import was necessary to make use of the Schema - like it is done in the [asyncapi-react](https://github.com/asyncapi/asyncapi-react/blob/next/library/src/components/__tests__/Schema.test.tsx#L8-L9) package. To get rid of this workaround, we have to export the `Schema`. This is done within this PR.

**Related issue(s)**

- none

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -8,6 +8,7 @@ const ParserError = require('./errors/parser-error');
 const { validateChannels, validateTags, validateServerVariables, validateOperationId, validateServerSecurity } = require('./customValidators.js');
 const { toJS, findRefs, getLocationOf, improveAjvErrors, getDefaultSchemaFormat } = require('./utils');
 const AsyncAPIDocument = require('./models/asyncapi');
+const Schema = require('./models/schema');
 
 const OPERATIONS = ['publish', 'subscribe'];
 //the only security types that can have a non empty array in the server security item
@@ -33,6 +34,7 @@ module.exports = {
   registerSchemaParser,
   ParserError,
   AsyncAPIDocument,
+  Schema,
 };
 
 /**


benlangfeld:
This is important to [be able to use @asyncapi/react-component > 1.0.0-next.31](https://github.com/backstage/backstage/pull/8973) to get the important security fix from https://github.com/asyncapi/asyncapi-react/pull/522. It would be really great if it could be merged and released with similar urgency to that change @magicmatatjahu

magicmatatjahu:
@dschwank @benlangfeld Hi! I don't understand one thing, why exporting Schema class is so important for security issues?

benlangfeld:
> @dschwank @benlangfeld Hi! I don't understand one thing, why exporting Schema class is so important for security issues?

It's that the latest version of this gem is not usable in Backstage without this change, so it's blocking all Backstage users from getting the fix for the CVEs.

magicmatatjahu:
> It's that the latest version of this package is not usable in Backstage without this change, so it's blocking all Backstage users from getting the fix for the CVEs.

But that absolute import in the `react-component` (for Schema class) is used since `next-1` version (you can check https://github.com/asyncapi/asyncapi-react/blob/v1.0.0-next.1/library/src/helpers/schema.ts#L3), so I'm wondering why is this a real problem? As I know Backstage uses new version of component from `next-26` version so that model didn't make any problems :) Is it due to some webpack plugin or a what? Also please have in mind that some parts of highlights.js are also imported by absolute paths like here https://github.com/asyncapi/asyncapi-react/blob/next/library/src/helpers/marked.ts#L7

And also very important thing: component has two versions, with and without `parser-js`, because in some envs there isn't needed for parser, like for our [studio](https://github.com/asyncapi/studio) or [html-template](https://github.com/asyncapi/html-template) and CJS modules doesn't support treeshaking, so if we will go with direction when we will import Schema class from root of package, then that version of component without parser will not work as expected because the final bundle will contain the parser anyway - this is mainly about the large amount of extra kbs.  

I really want to understand your problem 😅

benlangfeld:
I'm afraid I don't know the full answer to that and will have to defer to @dschwank .

dschwank:
Thanks a lot for the response @magicmatatjahu!

Just one note first - the issue with `marked` is a side-effect of the typing issue. Currently we can't update the dependency because of this.

Regarding the typing issue which this PR tries to fix - this issues started with this commit: https://github.com/asyncapi/asyncapi-react/commit/5412033f565b2363fa41ce5aefe0191a9c4d8a34

With [this](https://github.com/asyncapi/asyncapi-react/commit/5412033f565b2363fa41ce5aefe0191a9c4d8a34#diff-c5edc24c040b703062de0e2a66c843f2432f8f6f347a27bf17276a6f6630b45bR244) the `SchemaClass` is exported as return type, but there is no type definition for that class. So before, with the `any` the `schema.d.ts` looked like this:

```ts
import { ChannelParameter, ServerVariable, Schema } from '@asyncapi/parser';
...
export declare class SchemaHelpers {
    ...
    static jsonToSchema(value: any): any;
    ...
}
```

But with [this](https://github.com/asyncapi/asyncapi-react/commit/5412033f565b2363fa41ce5aefe0191a9c4d8a34#diff-c5edc24c040b703062de0e2a66c843f2432f8f6f347a27bf17276a6f6630b45bR244) change, the `schema.d.ts` looks like this:

```ts
import { ChannelParameter, ServerVariable, Schema } from '@asyncapi/parser';
import SchemaClass from '@asyncapi/parser/lib/models/schema';
...
export declare class SchemaHelpers {
    ...
    static jsonToSchema(value: any): SchemaClass;
    ...
}
```

The troublemaker is this line here `import SchemaClass from '@asyncapi/parser/lib/models/schema';`, because the type does not exist. To fix this, we could either switch back to the old `any` case, or export the `SchemaClass` here in the parser.

So far my understanding of the issue 😉  I hope that explains the issue - let me know if something is not clear!

magicmatatjahu:
@dschwank Thanks for explanation!

> The troublemaker is this line here import SchemaClass from '@asyncapi/parser/lib/models/schema';, because the type does not exist. To fix this, we could either switch back to the old any case, or export the SchemaClass here in the parser.

I didn't think of this change in those terms! I will keep it in mind for the future. ParserJS itself is going to be rewritten on TS as well as ESM so we'll fix that sooner or later, but as I wrote, parser-js uses CJS which makes treeshaking difficult so I won't agree to this solution of exporting Scheme class from root of the package. if you can change this type in your PR react-component to `any` I'll accept it quickly :)

If I can, I have a question, from what I understand you import this `SchemaHelpers` class in your code that's why it throws errors right?

dschwank:
Thanks @magicmatatjahu!

> I didn't think of this change in those terms! I will keep it in mind for the future.

All good, that's why we are all a community - we together build things, break things and make them work again 😃

> ParserJS itself is going to be rewritten on TS as well as ESM so we'll fix that sooner or later, but as I wrote, parser-js uses CJS which makes treeshaking difficult so I won't agree to this solution of exporting Scheme class from root of the package

To be honest switching back to `any` is not my preference, but you have a good point with the CJS treeshaking! So I am fine with it! Also nice that you plan to switch to TypeScript here! I was about to create a ticket to have better typing support, but with the planned rewrite that would be obsolete! If we can help somehow, let us know!

> if you can change this type in your PR react-component to any I'll accept it quickly :)

Awesome - I just updated this PR https://github.com/asyncapi/asyncapi-react/pull/527

> If I can, I have a question, from what I understand you import this SchemaHelpers class in your code that's why it throws errors right?

No, we only import the `AsyncApi` component like this `import AsyncApi from '@asyncapi/react-component';` , which is working fine, but the `tsc` with having `--skipLibCheck false` fails while checking the dependencies 😬

dschwank:
As discussed, closing this issues in favor of the fix here: https://github.com/asyncapi/asyncapi-react/pull/527

magicmatatjahu:
@dschwank 

> I was about to create a ticket to have better typing support, but with the planned rewrite that would be obsolete! If we can help somehow, let us know!

Thanks! Atm we have PR to enable releasing for next version like in react-component (PR: ) and then we should start rewriting. Help is more than welcome, but don't rush, hah 😅 

> No, we only import the AsyncApi component like this import AsyncApi from '@asyncapi/react-component'; , which is working fine, but the tsc with having --skipLibCheck false fails while checking the dependencies 😬

Ahh, ok, I understand :) Good to know about such a problem, I will have it in my mind! Thanks again! :) React-component with fix should be released about 10-15 minutes :)

PR #453, opened 2022-01-21
feat: initial intent API implementation

jonaslagoni:
**Description**

- ...
- ...
- ...

**Related issue(s)**


jonaslagoni:
The PR is ready to have its first review. I have made a lot of assumptions and decisions, so please review them and provide feedback as you see fit. 

Remember this PR is just to set the foundation of the next changes.

jonaslagoni:
And @derberg I might need your help dialing down why browser tests fail, as I don't see any reason why they should 🧐 Any hints as to how to debug it?

fmvilas:
Alright, so after playing with the API a bit, here's my feedback.

I think the method names are super long and might cause confusion or make the API not enjoyable at all. I'm leaving next some ideas that came to my mind while playing with it. They came to me in this order. I'm leaving all the steps in case visualizing my thiking process helps:

```js
// Using it as a client (MQTT, Kafka, WS client, etc.)

const parser = require('./lib')
const client = await parser.parse('...')

client.publishOperations() // Returns all the subscribe operations
client.publishableChannels() // Returns all the channels with subscribe operations

// Using it as a server (HTTP server, WS server, anything else?)

const parser = require('./lib')
const server = await parser.parseAsServer('...')

server.publishOperations() // Returns all the publish operations
server.publishableChannels() // Returns all the channels with publish operations
```

Notice `parseAsServer`. It should only work on `v2.x` documents. Or at least work with `v3.x` too but leave a message saying it's not necessary and it's not recommended. For `v3.x` simply use `parse` all the time. With this, we're signaling old users we care about them but we're not going to make v3 experience worst just because some people don't want to upgrade. Instead, we make the DX gracefully degrade for v2 users. Result: 1st benefit if you upgrade to v3.

But since in v3 we'd probably rename verbs to `send` and `receive` to avoid confusions with pub/sub pattern,
it can be as follows:

```js
// Using it as a client (MQTT, Kafka, WS client, etc.)

const parser = require('./lib')
const client = await parser.parse('...')

client.sendOperations() // Returns all the subscribe operations
client.sendableChannels() // Returns all the channels with subscribe operations

// Using it as a server (HTTP server, WS server, anything else?)

const parser = require('./lib')
const server = await parser.parseAsServer('...')

server.sendOperations() // Returns all the publish operations
server.sendableChannels() // Returns all the channels with publish operations
```

However, IMHO, `sendableChannels` (and all the sendable/receivable methods) don't look great. What if
we just do this:

```js
// Using it as a client (MQTT, Kafka, WS client, etc.)

const parser = require('./lib')
const client = await parser.parse('...')

client.operations() // Returns all the operations
client.operations.filterBy({ send: true }) // Returns all the subscribe operations
client.operations.filterBy({ receive: true }) // Returns all the publish operations
client.operations.filterBy({ send: true, receive: true }) // Returns all the operations
client.channels() // Returns all the channels
client.channels.filterBy({ send: true }) // Returns all the channels with subscribe operations
client.channels.filterBy({ receive: true }) // Returns all the channels with publish operations
client.channels.filterBy({ send: true, receive: true }) // Returns all the channels
client.channels.filterBy({ send: false, receive: false }) // Returns no channels

// Using it as a server (HTTP server, WS server, anything else?)

const parser = require('./lib')
const server = await parser.parseAsServer('...')

server.operations() // Returns all the operations
server.operations.filterBy({ send: true }) // Returns all the publish operations
server.operations.filterBy({ receive: true }) // Returns all the subscribe operations
server.operations.filterBy({ send: true, receive: true }) // Returns all the operations
server.channels() // Returns all the channels
server.channels.filterBy({ send: true }) // Returns all the channels with publish operations
server.channels.filterBy({ receive: true }) // Returns all the channels with subscribe operations
server.channels.filterBy({ send: true, receive: true }) // Returns all the channels
server.channels.filterBy({ send: false, receive: false }) // Returns no channels
```

That said, I'm aware this API might not be portable to all languages so I propose two things:

* We do something like `server.channels().filterBy(...)` instead of `server.channels.filterBy(...)`. Or;
* We define language bindings (as in mappings) of the Parser API. It is, the API will be super similar across languages but not exactly the same. E.g., Go will require us to have `Channels()` instead of `channels()`. That's fine, we can leave it defined that's the way to Go :wink:. And in the case of JS/TS, you can have the structure defined in the examples above.

What do you think?

derberg:
@jonaslagoni have a look at https://github.com/asyncapi/parser-js/blob/master/test/browser_test.js#L11 and you can also add some old school `console.log` to https://github.com/asyncapi/parser-js/blob/master/test/sample_browser/index.html 😄

jonaslagoni:
> What do you think?

@fmvilas I actually love the last suggestion! I get the long function name problem and agree we should probably fix that somehow.

However, I would suggest we change the interface a bit in order to make the UX better, especially if we consider generator use-case.

What if we adapted it for:
```js
const parser = require('./lib')
// const { client, server } = await parser.parse('...')
const parsedDocument = await parser.parse('...')
const client = parsedDocument.client
const server = parsedDocument.server
```

This way, you have access to all perspectives, when required, and you don't need to re-parse documents to gain the opposite view (generator can't assume what templates require in terms of perspectives). What do you think about that change?

Not sure about the `client`/`server`, but yea, another problem for another day.

fmvilas:
At first, I thought: "oh! that's even better!" but then realized it has a couple of issues:

* It doesn't make sense in the v3 world where this client/server dichotomy is not present anymore. Therefore, we would be dragging bad v2 decisions to v3. I think v2 UX should "gracefully degrade" instead.
* This UX is more confusing for non-servers developers, i.e., the majority? Imagine someone having to choose `server` because they're developing the application instead of a client of the application. They're also developing another client that connects to a broker but they have to choose `server` instead of `client`. In other words, we move the confusion from the spec to the parser, instead of abstracting it.

In my proposal, I could have done `parseAsClient` and `parseAsServer` but I intentionally didn't do it, so `parse` remains there forever and only those in v2 will have to think about the client/server dichotomy. Fair enough. If they want this dichotomy removed, they can upgrade to v3. See what I'm trying to do here?

After writing these lines, I thought maybe we can have `parseAsServer` and `parseAsImplementer`. They both interpret the document in the same way but `parseAsServer` would feel natural for those creating a server (WS, HTTP, etc.) and `parseAsImplementer` would feel natural for those implementing the application/service being described instead of a client of it. Still, the confusion is there but we immediately deprecate both methods and recommend people migrate to v3. Over time, we can get rid of them and problem solved.

smoya:
I understand having `Application` and `Client` prefixes on method names can be seen as a penalty to the DX. However, they were added curiously because of DX;  users won't ever need to know which "role" or "point of view" (we didn't agree on a name for that afaik) was set. Instead, methods are deterministic becoming less error-prone (I'd say totally error-prone-less if that word even exists 😆 ).

On that side, I would keep advocating for having deterministic methods.

If we see that the API is becoming complicated because of v2, we can always make the intent-driven API to only work with v3 or greater, meaning v2 users will keep using the old parser-js API (we keep the code in there for some time but we don't support new features) and only those in v3 would be able to use the new API.

If we don't want to go in that direction, another solution to keep methods deterministic could be to skip the `Application` prefix (as this is the only "point of view" on v3 spec) and keep the `Client` one. In that way, v2 users will just use all the `Client` prefixed methods (default "point of view" on v2 spec) until we decide to remove support for it.

smoya:
> However, IMHO, `sendableChannels` (and all the sendable/receivable methods) don't look great. What if we just do this:
> 
> ```js
> // Using it as a client (MQTT, Kafka, WS client, etc.)
> 
> const parser = require('./lib')
> const client = await parser.parse('...')
> 
> client.operations() // Returns all the operations
> client.operations.filterBy({ send: true }) // Returns all the subscribe operations
> client.operations.filterBy({ receive: true }) // Returns all the publish operations
> client.operations.filterBy({ send: true, receive: true }) // Returns all the operations
> client.channels() // Returns all the channels
> client.channels.filterBy({ send: true }) // Returns all the channels with subscribe operations
> client.channels.filterBy({ receive: true }) // Returns all the channels with publish operations
> client.channels.filterBy({ send: true, receive: true }) // Returns all the channels
> client.channels.filterBy({ send: false, receive: false }) // Returns no channels
> 
> // Using it as a server (HTTP server, WS server, anything else?)
> 
> const parser = require('./lib')
> const server = await parser.parseAsServer('...')
> 
> server.operations() // Returns all the operations
> server.operations.filterBy({ send: true }) // Returns all the publish operations
> server.operations.filterBy({ receive: true }) // Returns all the subscribe operations
> server.operations.filterBy({ send: true, receive: true }) // Returns all the operations
> server.channels() // Returns all the channels
> server.channels.filterBy({ send: true }) // Returns all the channels with publish operations
> server.channels.filterBy({ receive: true }) // Returns all the channels with subscribe operations
> server.channels.filterBy({ send: true, receive: true }) // Returns all the channels
> server.channels.filterBy({ send: false, receive: false }) // Returns no channels
> ```
> 
> That said, I'm aware this API might not be portable to all languages so I propose two things:
> 
> * We do something like `server.channels().filterBy(...)` instead of `server.channels.filterBy(...)`. Or;
> * We define language bindings (as in mappings) of the Parser API. It is, the API will be super similar across languages but not exactly the same. E.g., Go will require us to have `Channels()` instead of `channels()`. That's fine, we can leave it defined that's the way to Go 😉. And in the case of JS/TS, you can have the structure defined in the examples above.
> 
> What do you think?

This is IMHO far away from the initial idea of creating an Intent-Driven API since I don't see the intents anymore. Filtering per see is not an intent but a detail of implementation on how to handle your intent. E.g. you would need all the channels you can send messages to, and that's the real intent. 
I'm not saying we should stick with the Intent-Driven API at any cost. I'm more than happy to adapt it to whatever is better for DX and to avoid breaking changes.

If we feel is a naming issue, we can move our efforts to solve that thing, e.g. `sendableChannels` could be named `ChannelsForSend` (yeah, not cool. just the first occurrence in my brain).

fmvilas:
> This is IMHO far away from the initial idea of creating an Intent-Driven API since I don't see the intents anymore. Filtering per see is not an intent but a detail of implementation on how to handle your intent. E.g. you would need all the channels you can send messages to, and that's the real intent.

@smoya You're right. I think it's a matter of translating the `filterBy` function to methods with all the possibilities. Could be something like:

```
// Using it as a client (MQTT, Kafka, WS client, etc.)

const parser = require('./lib')
const client = await parser.parse('...')

client.operations.filterBySend() // Returns all the subscribe operations
client.operations.filterByReceive() // Returns all the publish operations
client.operations.all() // Returns all the operations

client.channels.filterBySend() // Returns all the channels with subscribe operations
client.channels.filterByReceive() // Returns all the channels with publish operations
client.channels.all() // Returns all the channels

// Using it as a server (HTTP server, WS server, anything else?)

const parser = require('./lib')
const server = await parser.parseAsServer('...')

server.operations.filterBySend() // Returns all the publish operations
server.operations.filterByReceive() // Returns all the subscribe operations
server.operations.all() // Returns all the operations

server.channels.filterBySend() // Returns all the channels with publish operations
server.channels.filterByReceive() // Returns all the channels with subscribe operations
server.channels.all() // Returns all the channels
```

I think the intent is clearer now while still avoiding names such as "sendableChannels" or "channelsForSend". Also, I think it's safe to assume there will always be `servers`, `channels`, `operations`, and `messages`. These entities can be used to scope other methods (like in the example above). This way we don't end up bloating the "global scope" with too many "getXXXofYYY()" method names.

> If we see that the API is becoming complicated because of v2, we can always make the intent-driven API to only work with v3 or greater, meaning v2 users will keep using the old parser-js API (we keep the code in there for some time but we don't support new features) and only those in v3 would be able to use the new API.

In an ideal world, yes, I'd be all for it. However, we want people to start migrating to intent-driven API before v3 is out. Just so when it's released it will cause no change for them. I know they're going to have little time to do it though. Definitely, something to think about.

smoya:
@jonaslagoni If you are happy, I can take over this and apply the latest suggestion @fmvilas made, which I kinda like it and I feel it will simplify the API ergo improve DX. WDYT?

jonaslagoni:
> @jonaslagoni If you are happy, I can take over this and apply the latest suggestion @fmvilas made, which I kinda like it and I feel it will simplify the API ergo improve DX. WDYT?

Please do! Let me know if you need anything 🙂

smoya:
I added a new commit with the implementation for the suggestions @fmvilas made [here](https://github.com/asyncapi/parser-js/pull/453#issuecomment-1038974100). 
I ended up creating an intermediate object for channels, operations, and messages.

smoya:
I think we would also need to introduce methods for getting objects by their ID, like `document.channel(<channel_id>)`,  `document.operation(<operation_id>)`, `document.server(<server_name>)`.

smoya:
- Added `IntentServers` object with same style as `IntentChannels`, `IntentOperations`, and ` IntentMessages). 
- Added a `id()` method on the following models: 
  - Channel -> returns key on the map for `2.x.x`
  - Operation -> returns `operation id` for `2.x.x`
  - Message -> returns `uid` for `2.x.x`
  - Server -> returns key on the map for `2.x.x`
- Removed `document.channel(<channel_name>)` method in favor of `document.channels().filterById(<name>)`.

smoya:
I'm thinking about introducing an `isEmpty()` method to all those intermediate objects, so we can remove the `document.hasChannels`-like methods. 
The point is that `isEmpty` is the opposite of what we previously had (emptiness vs fullness) so not sure what others think about.

smoya:
> I wonder if we shouldn't be generating operation ids for those operations that don't have one defined. This way we can make sure we never return `undefined` in any version.

Yeah, I think it makes sense. Would you also suggest we write it as a rule for all parser implementations (Parser-API) ? So the return type should be only `string` and  never empty then.

fmvilas:
> Would you also suggest we write it as a rule for all parser implementations (Parser-API) ?

I think so. Not sure how far we should get into implementation details other than the API surface but definitely something to take into account. In any case, if we say that operations MUST always have an ID, implementations will have to generate them 😄

smoya:
@jonaslagoni would you mind converting this issue into a draft? As this is just a POC, I don't think it makes sense to keep it as a regular PR.

PR #465, opened 2022-02-02
fix: do not use the Array.prototype.flat function

magicmatatjahu:
**Description**

In some NodeJS version we cannot use `Array.prototype.flat` function. I fix that and switch to implementation with `push` function.

**Related issue(s)**
Example problem in this PR https://github.com/asyncapi/website/pull/579

### DIFF
--- a/lib/asyncapiSchemaFormatParser.js
+++ b/lib/asyncapiSchemaFormatParser.js
@@ -43,19 +43,18 @@ async function parse({ message, originalAsyncAPIDocument, fileFormat, parsedAsyn
  * @private
  */
 function getMimeTypes() {
-  const mimeTypes = [];
-  ['2.0.0', '2.1.0', '2.2.0', '2.3.0'].forEach(version => {
-    mimeTypes.push(...[
-      `application/vnd.aai.asyncapi;version=${version}`,
-      `application/vnd.aai.asyncapi+json;version=${version}`,
-      `application/vnd.aai.asyncapi+yaml;version=${version}`,
-    ]);
-  });
-  mimeTypes.push(...[
+  const mimeTypes = [
     'application/schema;version=draft-07',
     'application/schema+json;version=draft-07',
     'application/schema+yaml;version=draft-07',
-  ]);
+  ];
+  ['2.0.0', '2.1.0', '2.2.0', '2.3.0'].forEach(version => {
+    mimeTypes.push(
+      `application/vnd.aai.asyncapi;version=${version}`,
+      `application/vnd.aai.asyncapi+json;version=${version}`,
+      `application/vnd.aai.asyncapi+yaml;version=${version}`,
+    );
+  });
   return mimeTypes;
 }
 
--- a/lib/asyncapiSchemaFormatParser.js
+++ b/lib/asyncapiSchemaFormatParser.js
@@ -43,15 +43,20 @@ async function parse({ message, originalAsyncAPIDocument, fileFormat, parsedAsyn
  * @private
  */
 function getMimeTypes() {
-  return [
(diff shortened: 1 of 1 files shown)


lib/asyncapiSchemaFormatParser.js
```diff
@@ -43,15 +43,20 @@ async function parse({ message, originalAsyncAPIDocument, fileFormat, parsedAsyn
...
+      `application/vnd.aai.asyncapi;version=${version}`,
+      `application/vnd.aai.asyncapi+json;version=${version}`,
+      `application/vnd.aai.asyncapi+yaml;version=${version}`,
+    ]);
+  });
+  mimeTypes.push(...[
+    'application/schema;version=draft-07',
+    'application/schema+json;version=draft-07',
+    'application/schema+yaml;version=draft-07',
+  ]);
```
jonaslagoni, line 58: You could simplify it a bit and do this: 

```suggestion
  const mimeTypes = ['application/schema;version=draft-07',
    'application/schema+json;version=draft-07',
    'application/schema+yaml;version=draft-07'];

  ['2.0.0', '2.1.0', '2.2.0', '2.3.0'].forEach(version => {
    mimeTypes.push(
      `application/vnd.aai.asyncapi;version=${version}`,
      `application/vnd.aai.asyncapi+json;version=${version}`,
      `application/vnd.aai.asyncapi+yaml;version=${version}`
    );
  });
```

magicmatatjahu: I wanna have `application/schema;version=draft-07` etc at the end of array as in original array :)

magicmatatjahu: but ok :)

jonaslagoni: Ahh right 🙂 So this? 

```suggestion
  const mimeTypes = [];

  ['2.0.0', '2.1.0', '2.2.0', '2.3.0'].forEach(version => {
    mimeTypes.push(
      `application/vnd.aai.asyncapi;version=${version}`,
      `application/vnd.aai.asyncapi+json;version=${version}`,
      `application/vnd.aai.asyncapi+yaml;version=${version}`
    );
  });
  mimeTypes.push('application/schema;version=draft-07',
    'application/schema+json;version=draft-07',
    'application/schema+yaml;version=draft-07'
  );
```

magicmatatjahu: Ok, we can have that `application/schema;version=draft-07` at the beginning :)

magicmatatjahu: I propagated suggestions.

jonaslagoni:
I guess we cant test this can we? To ensure this cannot be introduced again?

magicmatatjahu:
> I guess we cant test this can we? To ensure this cannot be introduced again?

In the current implementation (with `flat`) it pass in all tests, so 🤷🏼  Probably we should downgrade NodeJS version to test it, but is it worth?

jonaslagoni:
> In the current implementation (with `flat`) it pass in all tests, so 🤷🏼 Probably we should downgrade NodeJS version to test it, but is it worth?

Depends on why we think the website deployment failed and whether it's a use-case we have to account for in the parser. Maybe create a follow-up issue afterward?

jonaslagoni (approved):

magicmatatjahu:
> Depends on why we think the website deployment failed and whether it's a use-case we have to account for in the parser. Maybe create a follow-up issue afterward?

Maybe we should run workflows as matrix, but not for systems but for NodeJS versions, in this repo and in others? cc 
@derberg

PR #468, opened 2022-02-07
ci: set next-major branch as base branch for bump action

smoya:
**Description**

This PR changes the base branch for the bump action so it only applies to dependants of the`next-major` branch, a branch that was recently created to work on the next major version (https://github.com/asyncapi/parser-js/issues/401).

Note that the target branch is (and should be) `next-major`. 

**Related issue(s)**

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -15,6 +15,7 @@ on:
 
 jobs:
   bump:
+    if: startsWith(github.event.commits[0].message, 'chore(release):')
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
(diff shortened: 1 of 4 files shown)


smoya:
I need to confirm this PR does what I expect it does. Would you @derberg mind clarifying the following? 🙏 

Any change made to `next-major` branch needed to be released by semantic-release, will trigger a version bump on all repositories that depend on that `next-major` release tag. Is that right?
On the other side, `next-major` branch would need to stay up-to-date with `master`. Is this something done automatically by any other action, or rather we need to update it manually?

Thanks!

smoya:
Branch is now up-to-date with `master` so by what @derberg mentioned in https://github.com/asyncapi/parser-js/pull/468#issuecomment-1035270509, it seems we are good to go now.

smoya:
Wait... why did the bot merged this PR when there was no 👍  from any Code Owner 🤯  ????? I added the `/rtm` because in case an Owner approved it, to get it merged automatically (as there was nothing else from my side to change).

smoya:
Pinging Code Owners @fmvilas @magicmatatjahu @jonaslagoni @derber.

I think some config is missing in this repository regarding branch protection. A PR shouldn't be merged if there is no +1 of at least one Code Owner. This should be configured in the repository settings, under branch protection rules iirc.

magicmatatjahu:
Why did you remove `next-branch`? I think that we can always change config for given branch if branch exists 😅

jonaslagoni:
> I think some config is missing in this repository regarding branch protection. A PR shouldn't be merged if there is no +1 of at least one Code Owner. This should be configured in the repository settings, under branch protection rules iirc.

It should 👍

PR #472, opened 2022-02-11
feat: release for version 3.0.0 of the spec

smoya:

Also, the reason to create this PR a long time before the release is to enable automation (bot keeps upstream branches always up to date with master) that we have in place to regularly update the release branch with whatever is changed in the master branch. So nobody has to do it manually.


jonaslagoni:
What the f... 😆

jonaslagoni:
@smoya might want to mark this as a draft? 🤔

smoya:
@jonaslagoni I'm not sure if CI automation for upstream branches applies to draft PR's. I think it does https://github.com/asyncapi/parser-js/blob/master/.github/workflows/autoupdate.yml ?

Let's see.

Issue #482, opened 2022-03-07
Migrate project to TypeScript


smoya:
#### Reason/Context
[TypeScript](https://www.typescriptlang.org/) provides a really good set of features on top of Javascript. It is very opinionated and used widely across the whole OSS scene.

Some of the features I consider very interesting here are Static (or strict) typing and OOP syntax models including interfaces, classes, enums, etc.

All those things improve not only the Developer Experience but codes stability and resiliency by:
- Early catching errors.
- Improving code reusability.
- Improving code extensibility.
- Making (unit) testing a more thorough thing.
- Having a nice IDE integration.
- Etc.

Definitely, it drives us through the road of [Producing software that can be easily extended, reusable, and highly testable](https://github.com/asyncapi/community/discussions/212).

Considering the [Road to v2.0.0](https://github.com/asyncapi/parser-js/issues/481) and the fact major changes are going to happen, such as a complete rewrite of the API of the parser (See https://github.com/asyncapi/parser-js/pull/453), or the rewrite we plan to do for the parsing logic integrating [Spectral](https://github.com/stoplightio/spectral) (See https://github.com/asyncapi/parser-js/issues/477), I can't think of a better time to move forward with this task.

#### Description
To rewrite/migrate the entire `parser-js` library in TypeScript.
We should create PR's targeting `next-major` branch if we finally want this to happen in `v2.0.0`.

This development also includes porting the new Parser-API based on the latest implementation that can be found here: https://github.com/asyncapi/parser-js/pull/497#issuecomment-1072418690

magicmatatjahu:
@smoya I am 100% in favor of this! I can of course help with the rewriting. We can split it so that you take care of models and intent API and I will take care of validation (by spectral) and parsing. WDYT?

smoya:
> @smoya I am 100% in favor of this! I can of course help with the rewriting. We can split it so that you take care of models and intent API and I will take care of validation (by spectral) and parsing. WDYT?

That's awesome Maciej! Thanks for jumping in 💯 . Let's zap all previous code so we can freshly start with this! https://github.com/asyncapi/parser-js/pull/483

Souvikns:
@smoya would love to help, I have been reading parser for some time now and I wanted to get started with the project but didn't know where to start, I think this is the best time to get started since we are rewriting the whole parser. Just let me know how I can help.

smoya:
[...]

PR #485, opened 2022-03-08
refactor: add info and dependant models

smoya:
**Description**

This PR adds Info and the dependant models (Contact, License). It also allows to retrieve Info from the document root.
The models are based on the new Intent API (even though Info looks the same by now).

Pending things that can be out of this PR:

- How to handle retrocompatibility.

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/482

### DIFF
--- a/src/models/asyncapi.ts
+++ b/src/models/asyncapi.ts
@@ -7,7 +7,6 @@ export class AsyncAPIDocument extends BaseModel {
     }
 
     info(): Info {
-        const doc = this.json("info");
-        return doc && new Info(doc);
+        return new Info(this.json("info"));
     }
 }
--- a/src/models/asyncapi.ts
+++ b/src/models/asyncapi.ts
@@ -1,3 +1,13 @@
 import { BaseModel } from "./base";
+import { Info } from "./info";
 
-export class AsyncAPIDocument extends BaseModel {}
+export class AsyncAPIDocument extends BaseModel {
+    version(): string {
+        return this.json("asyncapi");
+    }
+
+    info(): Info {
+        const doc = this.json("info");
+        return doc && new Info(doc);
+    }
+}
--- a/src/models/info.ts
+++ b/src/models/info.ts
@@ -19,12 +19,12 @@ export class Info extends BaseModel {
         return this.json("termsOfService");
     }
 
-    contact(): Contact {
+    contact(): Contact | undefined {
         const doc = this.json("contact");
         return doc && new Contact(doc);
     }
 
-    license(): License {
+    license(): License | undefined {
         const doc = this.json("license");
         return doc &&  new License(doc);
     }
--- a/src/models/info.ts
+++ b/src/models/info.ts
@@ -1,0 +1,31 @@
+import { BaseModel } from "./base";
+import { Contact } from "./contact";
+import { License } from "./license";
+
+export class Info extends BaseModel {
+    title(): string {
+        return this.json("title");
+    }
+
+    version(): string {
+        return this.json("version");
+    }
+
+    description(): string {
+        return this.json("description");
+    }
+
+    termsOfService(): string {
+        return this.json("termsOfService");
+    }
+
+    contact(): Contact {
+        const doc = this.json("contact");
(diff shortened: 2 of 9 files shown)


fmvilas (approved):

src/models/asyncapi.ts
```diff
@@ -1,3 +1,13 @@
...
 
-export class AsyncAPIDocument extends BaseModel {}
+export class AsyncAPIDocument extends BaseModel {
+    version(): string {
+        return this.json("asyncapi");
+    }
+
+    info(): Info {
+        const doc = this.json("info");
+        return doc && new Info(doc);
```
magicmatatjahu, line 11: info is required object so you don't need to check that

```suggestion
        return new Info(this.json("info"));
```

magicmatatjahu (changes requested):

src/models/info.ts
```diff
@@ -0,0 +1,31 @@
...
+    termsOfService(): string {
+        return this.json("termsOfService");
+    }
+
+    contact(): Contact {
+        const doc = this.json("contact");
+        return doc && new Contact(doc);
+    }
+
+    license(): License {
```
magicmatatjahu, line 27: We have change return type to `Contact | undefined`. Why? Because some fields are optional, so in TS tools people will have info that they should check if value exist or not, if you return only `Contact` you tell: `Hey, you will have always Contact instance` and it's not true.

I don't know how we will handle situation for v2 and v3, example: `channels` will be optional from v3 (in 99%, yes?) so in v2 people will have always list channels, but in v3 they will have undefined or list. We have to discuss it.

```suggestion
    contact(): Contact | undefined {
        const doc = this.json("contact");
        return doc && new Contact(doc);
    }

    license(): License | undefined {
```

magicmatatjahu (changes requested):

src/models/asyncapi.ts
smoya: I applied this but I'm not very convinced about it right now. Shouldn't we handle this anyway for the matter of consistency and testing?

magicmatatjahu: But what do you mean? in v2 `info` is required so it cannot be undefined so `info` method should always return `Info` instance. Or maybe I don't understand something. 🤔 ]

derberg:
@magicmatatjahu sorry I didn't get it why it can't be released as release candidate? what react component has to do here?

src/models/asyncapi.ts
smoya: Remember we are building here an API based on intents. `Info` model is not a 1:1 with the `info` key (even though it is right now). 

I think it's not the fact we check if it's empty or not what it "shocks"me. It's the differences between major spec versions what and how to handle them. But this is out of this PR and will work right after merging it so sooner the better. So Nvm by now :P

magicmatatjahu:
@derberg due to the fact that in react component the first dozen of PRs were not released because the component itself did not work. Each PR was a new component rewritten to make PR easier. It is the same here. So the first PRs on `next-major` should not be released, because nothing works yet and it will work maybe in 2 weeks or a month. Of course I wanna release `next-major` as prerelease but not now.

derberg:
@magicmatatjahu oh, ok, now it makes sense. I did not review this PR so was not aware what is the status. I totally agree that every release should include "working code", same with prereleases, something that can actually be used and integrated

src/models/asyncapi.ts
magicmatatjahu: ok i understand. However, it would be necessary to have a consensus on how to do it in relation to the AsyncAPI version because it will probably be the most difficult task here.

It was a mistake that I didn't participate at all in the discussion about IntentAPI, because e.g. in react-component or markdown-template intent api won't be used at all, so I'm still thinking about 1:1 of API into specification structure.

magicmatatjahu:
@derberg No worries :)

@smoya So I suggest to use `refactor` for first PRs, ok?

magicmatatjahu (approved):
LGTM

Issue #481, opened 2022-03-03
Road to v2.0.0


magicmatatjahu:
This issue contains list of possible features that we wanna introduce in the `v2.0.0` of ParserJS. Have in mind that some of them may not be implemented. You can also write your own proposals/features/PRs in the comments, I will update then the main list.

> Note: Mainly PRs should be merged to the `next-major` branch. Additionally, some of the following issues will only be fixable/implementable once the other issues are resolved.

- [x] https://github.com/asyncapi/parser-js/issues/482
- [x] https://github.com/asyncapi/parser-js/issues/401
  - [x] https://github.com/asyncapi/parser-js/pull/453
  - [X] https://github.com/asyncapi/parser-js/issues/201
- [x] https://github.com/asyncapi/parser-js/issues/477
  - [X] https://github.com/asyncapi/parser-js/issues/329 - Spectral supports that out of the box
  - [ ] https://github.com/asyncapi/parser-js/issues/315 - we need to wait for that PR https://github.com/asyncapi/spec-json-schemas/pull/239. Spectral will supports bindings validation out of the box, but if we need, we can improve error messages for that.
  - [X] https://github.com/asyncapi/parser-js/issues/99
    - https://github.com/stoplightio/spectral/pull/2126
- [X] https://github.com/asyncapi/parser-js/issues/480
  - https://github.com/asyncapi/parser-js/pull/579
- [x] https://github.com/asyncapi/parser-js/issues/421
- [X] https://github.com/asyncapi/parser-js/issues/321
  - https://github.com/asyncapi/parser-js/pull/542
- [X] https://github.com/asyncapi/parser-js/issues/157
  - https://github.com/asyncapi/parser-js/pull/579

## Nice To Have
- [x] https://github.com/asyncapi/parser-js/issues/294
- [x] https://github.com/asyncapi/parser-js/issues/405
  - https://github.com/stoplightio/spectral/pull/2262
- [ ] https://github.com/asyncapi/parser-js/issues/404 - we need to check that in v2 by simple tests
- [ ] https://github.com/asyncapi/parser-js/issues/403 - we need to check that in v2 by simple tests
- [ ] https://github.com/asyncapi/parser-js/issues/360 - we need to check that in v2 by simple tests
- [ ] https://github.com/asyncapi/parser-js/issues/224 - we need to check that in v2 by simple tests
- [ ] https://github.com/asyncapi/parser-js/issues/50
[...]

PR #489, opened 2022-03-10
refactor: init custom parser functionality

magicmatatjahu:
**Description**

- init custom parser functionality: `registerSchemaParser` etc.
- init custom operations like `applyTraits` or `parseSchemas`.
- without tests, I will add them soon.

**Related issue(s)**
Part of #481 
Part of #482

### DIFF
--- a/src/custom-operations/apply-traits.ts
+++ b/src/custom-operations/apply-traits.ts
@@ -1,0 +1,60 @@
+import { JSONPath } from 'jsonpath-plus';
+
+import { xParserOriginalTraits } from '../constants';
+import { mergePatch } from '../utils';
+
+const v2TraitPaths = [
+  // operations
+  '$.channels.*.[publish,subscribe]',
+  '$.components.channels.*.[publish,subscribe]',
+  // messages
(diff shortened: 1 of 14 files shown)


jonaslagoni (commented):
No tests? 😄 

> without tests, I will add them soon.

Ah 😆

src/custom-operations/apply-traits.ts
```diff
@@ -0,0 +1,60 @@
+import { JSONPath } from 'jsonpath-plus';
+
+import { xParserOriginalTraits } from '../constants';
+import { mergePatch } from '../utils';
+
+const v2TraitPaths = [
```
smoya, line 6: `v2` means `v2` of the spec here? If so, shall we extract the paths to their own files so we can do the same for `v3` spec ?

magicmatatjahu: We will have only 2 custom operations and I don't think so that we should split them to separate files. We can do it later, if we will have more logic for each version.

smoya (approved):
LGTM!

PR #496, opened 2022-03-16
refactor: add metadata to models

magicmatatjahu:
**Description**

- add metadata to model like `pointer` (JSON Pointer), `asyncapi` data with detailed AsyncAPI document, `parent` which points to the parent model
- add helpers (similar we have in old implementation) for creating map/list of given model etc.
- add unit tests
- add missed mixins to models

**Related issue(s)**
Part of #481 
Part of #482


magicmatatjahu:
@smoya Why do we need metadata and a pointer for example? E.g. in react component and also in studio we make anchors by which people can move between given sections of specification (in html preview and also in editor). Current solution in studio is custom and not very optimal. By adding this small thing to the model implementation we have the ability to specify where the object is located and it is version agnostic. In the future we can always add other things to the metadata that will be needed. Write what you think about it :)

smoya (approved):
LGTM!

PR #497, opened 2022-03-16
refactor: add models for server object

Souvikns:
**Description**
- add `server` model
- add unit tests

**Related issue(s)**

Part of https://github.com/asyncapi/parser-js/issues/482


Souvikns:
I am opening this for review, I think I am done with most of the work, will be adding tests now.

src/models/server-security-requirement.ts
```diff
@@ -0,0 +1,5 @@
+import { BaseModel } from "./base";
+
+export interface ServerSecurityRequirementInterface extends BaseModel {
+
+}
```
magicmatatjahu, line 5: `ServerSecurityRequirementInterface` is unnecessary because it should not be a model, but only type https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#securityRequirementObject

src/models/server.ts
```diff
@@ -0,0 +1,14 @@
...
+import {BindingsMixinInterface, DescriptionMixinInterface, SpecificationExtensionsMixinInterface} from './mixins';
+
+export interface ServerInterface extends BaseModel, BindingsMixinInterface, DescriptionMixinInterface, SpecificationExtensionsMixinInterface {
+    url(): string;
+    protocol(): string;
+    protocolVersion(): string
+    variables(): Record<string, ServerVariableInterface>;
+    variable(name: string): ServerVariableInterface;
+    hasVariables(): boolean;
+    security(): [ServerSecurityRequirementInterface] | undefined;
```
magicmatatjahu, line 13: `url` and `protocol` are required, rest of fields are optional, so they should return `... | undefined`.

Also I opt to have methods like:

```ts
hasVariables(): boolean;
hasVariables(name: string): boolean;

variables(): Record<string, ServerVariableInterface> | undefined;
variables(name: string): ServerVariableInterface | undefined;
```

cc @smoya 

also type like `[ServerSecurityRequirementInterface]` is a tuple, not array of the items.

test/models/v2/server.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { ServerV2 } from '../../../src/models';
+
+describe('Server model', function () {
```
magicmatatjahu, line 3: We should write test for every model :) Please remember about that.

magicmatatjahu (changes requested):
Good, but we need some improvements. Please @smoya look at my comments, because we probably should change Parser API.

magicmatatjahu (changes requested):

src/models/asyncapi.ts
```diff
@@ -5,10 +5,13 @@ import { AsyncAPIDocumentV2 } from "./v2";
 import { AsyncAPIDocumentV3 } from "./v3";
 
 import { ExternalDocsMixinInterface, SpecificationExtensionsMixinInterface, TagsMixinInterface } from "./mixins";
+import { ServerInterface } from "./server";
 
 export interface AsyncAPIDocumentInterface extends BaseModel, ExternalDocsMixinInterface, SpecificationExtensionsMixinInterface, TagsMixinInterface {
   version(): string;
   info(): InfoInterface;
+  servers(): Record<string, ServerInterface>;
+  server(name: string): ServerInterface;
```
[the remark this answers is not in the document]
Souvikns: Kind of confused on this one,  are we not following the https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#asyncapidocument it has `server(name)` and `servers`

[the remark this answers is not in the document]
magicmatatjahu: I know it may be strange, but we can change this API if we think there is a better solution :)

src/models/asyncapi.ts
```diff
@@ -5,10 +5,16 @@ import { AsyncAPIDocumentV2 } from "./v2";
...
 import { ExternalDocsMixinInterface, SpecificationExtensionsMixinInterface, TagsMixinInterface } from "./mixins";
+import { ServerInterface } from "./server";
 
 export interface AsyncAPIDocumentInterface extends BaseModel, ExternalDocsMixinInterface, SpecificationExtensionsMixinInterface, TagsMixinInterface {
   version(): string;
   info(): InfoInterface;
+  servers(): Record<string, ServerInterface>;
+  hasServers(): boolean;
+  serverNames(): Array<string>;
+  server(name: string): ServerInterface;
```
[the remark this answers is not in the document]
smoya: We did an initial implementation in plain JS in this PR https://github.com/asyncapi/parser-js/pull/453.
Btw, that PR is like a first implementation draft on the parser-api and we should stick as much as we can with it.

In particular, this is how we solve all the collection of models:

We create an intermediate model, in this case `servers` which has some methods on it, like `isEmpty()`. 
You can see: https://github.com/asyncapi/parser-js/pull/453/files#diff-e871b9dd78d395cf3023273d1db969ee7638c3f5079fb1ac9323a921f46d6016

src/models/server.ts
smoya: Please also read https://github.com/asyncapi/parser-js/pull/497#discussion_r830002712.

Intermediate models are the solution.

smoya:
@Souvikns I'm not sure if I mentioned already, if not I'm so sorry, but @jonaslagoni and me did an initial implementation of the new Intent API but in the current plain javascript parser. It is here and has been through several iterations getting feedback: https://github.com/asyncapi/parser-js/pull/453

So we would need to port that to TS "mainly".

magicmatatjahu:
@smoya I didn't know about that 😆

src/models/asyncapi.ts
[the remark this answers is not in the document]
magicmatatjahu: Sorry I didn't know about that 😆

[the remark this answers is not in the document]
smoya: Yeah, my bad. I created a new issue in parser-api to let parser-api to keep up with that implementation:  https://github.com/asyncapi/parser-api/issues/53

Souvikns:
I have a question @smoya 

In `operations.js` file here https://github.com/jonaslagoni/parser-js/blob/259446323a31786a867f3714457445278eecbf45/lib/models/operations.js#L26 we have `operations.type()` but I don't see any implementation of it in https://github.com/jonaslagoni/parser-js/blob/feature%2Fmock_implementation/lib/models/operation.js also I can't find any definition of `type` in operation object in spec - https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#operationObject 

Can you give me some insight on this?

magicmatatjahu:
@Souvikns `type` is type of operation. In 2.x.x it will be `publish/subscribe`, in the 3.x.x it will be `send/receive` :) You must treat the operation key as a type.

magicmatatjahu:
@Souvikns At the moment you should only focus on 2.x.x not on 3.x.x (we don't have any merged PR for 3.x.x yet), so in the ^2.2.x versions you have `servers` field in the ChannelObjects so you should retrieve all operations defined for given server. Empty `servers` field means that all operations defined in all channels belong to the server, and vice versa, defined `servers` means that given channel belongs only to the given server(s) :)

magicmatatjahu:
@Souvikns But atm we don't have implemented operations and other needed models so please skip methods (create only signature) related to them, ok?

Souvikns:
@magicmatatjahu should I be waiting for this https://github.com/asyncapi/parser-js/pull/499 since I think I need to use the `collection`  class for the `servers` class?

magicmatatjahu:
@Souvikns As I know Sergio now is making review, so yes, please wait :)

smoya:
> @Souvikns As I know Sergio now is making review, so yes, please wait :)

It is done, and there is no blocker for merging (there is one thing we can change later after merging). So it's just a fact of merging it now 🚀

magicmatatjahu:
Merged :)

src/models/v2/server.ts
```diff
@@ -0,0 +1,43 @@
+import { Mixin } from '../utils';
+import { BaseModel } from '../base';
+import { ServerInterface } from '../server';
+import { DescriptionMixin } from './mixins/description';
+import { BindingsMixin } from './mixins/bindings';
+
+export class Server extends Mixin(BaseModel, DescriptionMixin, BindingsMixin) implements ServerInterface {
+    id(): string {
+        return this.json('id');
+    }
```
magicmatatjahu, line 10: We had discussion in Slack that `id` should not be a part of `_json` :) I also gave you example how it should work - https://github.com/asyncapi/parser-js/blob/next-major/src/models/v2/mixins/extensions.ts#L11

```suggestion
  constructor(
    private readonly _id: string,
    _json: Record<string, any>,
  ) {
    super(_json);
  }

  id(): string {
    return this._id;
  }
```

src/models/v2/server.ts
```diff
@@ -0,0 +1,43 @@
...
+        return this.json('protocolVersion');
+    }
+
+    hasUrl(): boolean {
+        return !!this.json('url');
+    }
+
+    url(): string {
+        return this.json('url');
+    }
```
magicmatatjahu, line 42: `protocol` and `url` will not be 99.9% optional fields in the specification, even in 3.x.x or 4.x.x, so please remove `hasUrl` and `hasProtocol`.

```suggestion
    protocol(): string | undefined {
        return this.json('protocol');
    }

    hasProtocolVersion(): boolean {
        return !!this.json('protocolVersion');
    }

    protocolVersion(): string {
        return this.json('protocolVersion');
    }

    url(): string {
        return this.json('url');
    }
```

src/models/v2/server.ts
```diff
@@ -0,0 +1,43 @@
...
+        return this.json('id');
+    }
+
+    hasName(): boolean {
+        return !!this.json('name');
+    }
+
+    name(): string | undefined {
+        return this.json('name');
+    }
```
magicmatatjahu, line 18: @smoya what is the `name` method in the server - https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#server?

src/models/v2/asyncapi.ts
```diff
@@ -17,4 +19,8 @@ export class AsyncAPIDocument
   info(): InfoInterface {
     return new Info(this._json.info);
   }
+
+  servers(): ServersInterface {
+    return new Servers(createArrayFromMap(this._json.servers));
+  }
```
magicmatatjahu, line 25: You are not creating a valid `Servers` instance. This class accepts `Server` instance arrays, not an array of normal objects:

```suggestion
  servers(): ServersInterface {
    return new Servers(
      Object.entries(this._json.servers).map(([serverName, server]) => new Server(serverName, server))
    );
  }
```

You have to import `Server` class.

src/models/v3/asyncapi.ts
```diff
@@ -4,16 +4,22 @@ import { Info } from "./info";
...
     return this.json("asyncapi");
   }
 
   info(): Info {
     return new Info(this.json("info"));
   }
+
+  servers(): ServersInterface {
+    return new Servers(this._json.servers);
+  }
```
magicmatatjahu, line 24: You are not creating a valid `Servers` instance. This class accepts `Server` instance arrays, not an array of normal objects:

```suggestion
  servers(): ServersInterface {
    return new Servers(
      Object.entries(this._json.servers).map(([serverName, server]) => new Server(serverName, server))
    );
  }
```

You have to import `Server` class.

src/models/v3/server.ts
```diff
@@ -0,0 +1,43 @@
+import { Mixin } from '../utils';
+import { BaseModel } from '../base';
+import { ServerInterface } from '../server';
+import { DescriptionMixin } from './mixins/description';
+import { BindingsMixin } from './mixins/bindings';
+
+export class Server extends Mixin(BaseModel, DescriptionMixin, BindingsMixin) implements ServerInterface {
```
magicmatatjahu, line 7: This same comment as for `v2` version about `url` and `protocol`.

src/models/v2/server.ts
smoya: The `name` is the key in the map of servers.

magicmatatjahu: then what is `id` then?

smoya: > the key in the map of servers

magicmatatjahu: ok, so `name` and `hasName` is out @Souvikns

src/models/v2/server.ts
Souvikns: I did this because I modeled the utils function to convert object to array something like this 

```
export function createArrayFromMap(json: Record<string,any>){
  const ArrayObject = [];
  for (const [key, value] of Object.entries(json)) {
    value['id'] = key;
    ArrayObject.push(value);
  };

  return ArrayObject;
}
```

so I thought maybe I don't need the add a separate id variable can just get it to format it. Main reason was so that the `v2` and `v3` have minimal changes, because to add seperate `_id` parameter now we have to pass something like `new Servers(id, this._json)`

src/models/v2/server.ts
Souvikns: So then what is `id`? I was thinking that `id` is the key

smoya: I think it makes sense to use ID for all objects like Channel, Server, Operation, etc, instead of `path()` for Channels, `name` for Servers, etc
If we add a ID method for all of those models, no matter if we later change the source of them between versions.
I know the parser-api doesn’t reflect that but we learnt a lot during all those months :)

Souvikns:
I am getting code smell warnings from `sonarcloud` for semicolons what should I do?

src/models/v2/server.ts
magicmatatjahu: v3 changes may be much bigger than they are now (to be honest we don't have anything approved so v3=v2 now), so please don't look at this proposal from Fran as something final.

About `id`. Have in mind that if someone will run `json()` should get the value that is defined for the object in the specification (not including traits because we have to combine them) without any additional fields that are not defined in the object, so what you want to get with adding the `id` field is nothing more than changing the original json. I know that adding an id as an argument to a constructor can be weird, but in this case the id is outside the object, and in v3 it can be part of the object and it will be easier to handle.

magicmatatjahu:
> I am getting code smell warnings from sonarcloud for semicolons what should I do?

as I see, it's from my PR (weird that sonarcloud didn't check it on my PR). I will remove it from my next, no worries.

src/models/v2/server.ts
```diff
@@ -0,0 +1,42 @@
...
+        return this.json('protocolVersion');
+    }
+
+    hasUrl(): boolean {
+        return !!this.json('url');
+    }
+
+    url(): string {
+        return this.json('url');
+    }
```
magicmatatjahu, line 41:
```suggestion
    protocol(): string | undefined {
        return this.json('protocol');
    }

    hasProtocolVersion(): boolean {
        return !!this.json('protocolVersion');
    }

    protocolVersion(): string {
        return this.json('protocolVersion');
    }

    url(): string {
        return this.json('url');
    }
```

test/models/v2/server.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+describe('Server Model', function () {
+
+    describe('.id()', function () {
+        it('should return name if present', function () {
+            expect(docItem.id()).toMatch('development');
+        })
+
+        it('should return undefined if name is not present', function () {
+            expect(emptyItem.id()).toMatch('');
+        })
```
magicmatatjahu, line 62: Unnecessary test.

```suggestion
```

test/models/v2/server.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    describe('.hasProtocol()', function () {
+        it('should return true if protocol is present', function () {
+            expect(docItem.hasProtocol()).toBeTruthy();
+        })
+
+        it('should return false if protocol is not present', function () {
+            expect(emptyItem.hasProtocol()).toBeFalsy();
+        })
+    })
```
magicmatatjahu, line 73: Unnecessary test.

```suggestion
```

test/models/v2/server.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    describe('protocol()', function () {
+        it('should return protocol ', function () {
+            expect(docItem.protocol()).toMatch(doc.development.protocol);
+        })
+
+        it('should return undefined when protocol is missing', function () {
+            expect(emptyItem.protocol()).toBeUndefined();
+        })
```
magicmatatjahu, line 82: Unnecessary test.

```suggestion
```

test/models/v2/server.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    describe('.hasUrl()', function () {
+        it('should return true if url is not missing', function () {
+            expect(docItem.hasUrl()).toBeTruthy();
+        })
+
+        it('should return false when url is missing', function () {
+            expect(emptyItem.hasUrl()).toBeFalsy();
+        })
+    })
```
magicmatatjahu, line 113: Unnecessary test.

```suggestion
```

test/models/v2/server.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    describe('.url()', function () {
+        it('should return url', function () {
+            expect(docItem.url()).toMatch(doc.development.url);
+        })
+
+        it('should return undefined when url is missing', function () {
+            expect(emptyItem.url()).toBeUndefined();
+        })
```
magicmatatjahu, line 122: Unnecessary test.

```suggestion
```

magicmatatjahu (changes requested):
Overall good, but I see that we have redundant unit tests and also in `v2` version we have unnecessary methods. I create suggestions :)

magicmatatjahu:
@Souvikns Good, last thing because I see that sonarcloud fails on duplicated code - please create `.sonarcloud.properties` file in the root with content:

```sonarcloud.properties
# Disable specific duplicate code since it would introduce more complexity to reduce it.
sonar.cpd.exclusions=src/models/**/*.ts
```

and we can merge :)

magicmatatjahu (approved):
LGTM! Thanks for contribution! :)

PR #501, opened 2022-03-21
feat: release for version 2.4.0 of the spec

smoya:

Also, the reason to create this PR a long time before the release is to enable automation (bot keeps upstream branches always up to date with master, see this [action](https://github.com/asyncapi/spec/blob/master/.github/workflows/autoupdate.yml)) that we have in place to regularly update the release branch with whatever is changed in the master branch. So nobody has to do it manually.


smoya:
@magicmatatjahu would you mind merging `master` branch into `2022-04-release` right directly in `2022-04-release` branch without PR? Otherwise, we will keep having the same issue we faced in https://github.com/asyncapi/spec/pull/740#issuecomment-1110854100 ?

smoya:
@magicmatatjahu thanks

magicmatatjahu (approved):

magicmatatjahu (approved):

PR #511, opened 2022-04-04
refactor: add Parser class

magicmatatjahu:
**Description**

Add `Parser` class. Having a different instance (of Parser) for validation/parsing is important from the point of view of tools for which the rules for Spectral will often change, e.g. `server-api`. Once the 2.0.0 version of the parser is released, people should be able to upload their own rules. This will also allow us to optimize the validation on the side of e.g. the Studio. 

**Related issue(s)**
Part of #481 
Part of #482


magicmatatjahu:
@jonaslagoni Thanks!

PR #517, opened 2022-04-13
chore: add new mimetype

jonaslagoni:
**Description**

todo:

- [x] Retarget branch


magicmatatjahu:
It will be needed (at least) in a few months, so why make PR now?

jonaslagoni:
> It will be needed (at least) in a few months, so why make PR now?

Hopefully, everything will be set up this week so we can start moving proposals to RFC-3. This means we need these changes to be ready to be merged into `next-major-spec`, once created 🙂

derberg (approved):

PR #521, opened 2022-04-14
chore: point @asyncapi/specs package to 2.14.0-2022-04-release.1

smoya:
Wondering if I could use some wildcard to not having to do this change every time we wanna merge a change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,11 +6,11 @@
   "packages": {
     "": {
       "name": "@asyncapi/parser",
-      "version": "1.14.1",
+      "version": "1.15.0-2022-04-release.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@apidevtools/json-schema-ref-parser": "^9.0.6",
-        "@asyncapi/specs": "^2.13.0",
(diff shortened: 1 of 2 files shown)


derberg:
yeah, we always tag releases - it should be `2022-04-release`
https://www.npmjs.com/package/@asyncapi/specs

but tbh I do not think it will do what you need, as package-lock will anyway lock to specific version, and without update there, it will still use the “old one”

derberg (approved):

smoya:
> yeah, we always tag releases - it should be `2022-04-release`
> https://www.npmjs.com/package/@asyncapi/specs
> 
> but tbh I do not think it will do what you need, as package-lock will anyway lock to specific version, and without update there, it will still use the “old one”

Yeah, right. A manual `npm i` will be neded.

derberg:
@smoya I just noticed you did it as `chore` so release candidate with it was not released 😄

smoya:
> @smoya I just noticed you did it as `chore` so release candidate with it was not released 😄

Facepalming but https://github.com/asyncapi/parser-js/pull/522

PR #526, opened 2022-04-19
feat!: release for version 3.0.0 of the spec

jonaslagoni:

Also, the reason to create this PR a long time before the release is to enable automation (bot keeps upstream branches always up to date with master, see this [action](https://github.com/asyncapi/spec/blob/master/.github/workflows/autoupdate.yml)) that we have in place to regularly update the release branch with whatever is changed in the master branch. So nobody has to do it manually.

### DIFF
--- a/.github/workflows/link-check-cron.yml
+++ b/.github/workflows/link-check-cron.yml
@@ -17,7 +17,7 @@ jobs:
 
     # Checks the status of hyperlinks in .md files
     - name: Check links
-      uses: gaurav-nelson/github-action-markdown-link-check@v1
+      uses: derberg/github-action-markdown-link-check@temporary-fix
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'yes'
@@ -29,8 +29,8 @@ jobs:
     - name: Report workflow run status to Slack
       uses: 8398a7/action-slack@v3
       with:
-          status: ${{ job.status }}
-          fields: repo,message,action,eventName,ref,workflow
+        status: ${{ job.status }}
+        fields: repo,action,workflow
       env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_DOCS_CHANNEL }}
       if: failure() # Only, on failure, send a message on the Slack Docs Channel (if there are broken links)
--- a/.github/workflows/link-check-pr.yml
+++ b/.github/workflows/link-check-pr.yml
@@ -6,14 +6,17 @@ name: Check Markdown links
 on: 
   pull_request_target:
     types: [synchronize, ready_for_review, opened, reopened]
+    paths:
+      - '**.md'
 
 jobs:
   External-link-validation-on-PR:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - name: Checkout repo
+      uses: actions/checkout@v3
     - name: Check links
-      uses: gaurav-nelson/github-action-markdown-link-check@v1
+      uses: derberg/github-action-markdown-link-check@temporary-fix
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'yes'
--- a/package-lock.json
+++ b/package-lock.json
@@ -1608,9 +1608,9 @@
       }
     },
     "node_modules/async": {
-      "version": "2.6.3",
-      "resolved": "https://registry.npmjs.org/async/-/async-2.6.3.tgz",
-      "integrity": "sha512-zflvls11DCy+dQWzTW2dzuilv8Z5X/pjfmZOWba6TNIVDm+2UDaJmXSOXlasHKfNBs8oo3M0aT50fDEWfKZjXg==",
+      "version": "2.6.4",
+      "resolved": "https://registry.npmjs.org/async/-/async-2.6.4.tgz",
+      "integrity": "sha512-mzo5dfJYwAn29PeiJ0zvwTo04zj8HDJj0Mn8TD7sno7q12prdbnasKJHhkm2c1LgrhlJ0teaea8860oxi51mGA==",
       "dev": true,
       "dependencies": {
         "lodash": "^4.17.14"
@@ -16610,9 +16610,9 @@
       "dev": true
     },
     "async": {
-      "version": "2.6.3",
-      "resolved": "https://registry.npmjs.org/async/-/async-2.6.3.tgz",
-      "integrity": "sha512-zflvls11DCy+dQWzTW2dzuilv8Z5X/pjfmZOWba6TNIVDm+2UDaJmXSOXlasHKfNBs8oo3M0aT50fDEWfKZjXg==",
+      "version": "2.6.4",
+      "resolved": "https://registry.npmjs.org/async/-/async-2.6.4.tgz",
+      "integrity": "sha512-mzo5dfJYwAn29PeiJ0zvwTo04zj8HDJj0Mn8TD7sno7q12prdbnasKJHhkm2c1LgrhlJ0teaea8860oxi51mGA==",
       "dev": true,
       "requires": {
         "lodash": "^4.17.14"
--- a/package.json
+++ b/package.json
@@ -94,6 +94,10 @@
       {
         "name": "next-major",
         "prerelease": true
+      },
+      {
+        "name": "next-major-spec",
+        "prerelease": true
       }
     ],
     "plugins": [


jonaslagoni:
This is gonna be a lovely merge conflict resolvement 🙄

fmvilas:
> This is gonna be a lovely merge conflict resolvement 🙄

It's been a lovely one, yeah ❤️

fmvilas:
Do we really need to release a major version with this PR? 🤔 It's adding on top of the existing API.

And, if we don't have to release a major version, why don't we merge whatever degree of v3 support that we have now? As long as we don't "advertise" it as stable, it should be fine, right? I'm trying to get rid of this long-lived branch.

smoya:
> Do we really need to release a major version with this PR? 🤔 It's adding on top of the existing API.
> 
> And, if we don't have to release a major version, why don't we merge whatever degree of v3 support that we have now? As long as we don't "advertise" it as stable, it should be fine, right? I'm trying to get rid of this long-lived branch.

Indeed, the idea behind the Parser-API was to avoid breaking changes as much as possible. Afaik, the API for v3 doesn't break anything. Is there any new feature added that Im missing the reason this is wanted to be released as major instead of minor? Cc @jonaslagoni

jonaslagoni:
> Do we really need to release a major version with this PR? 🤔 It's adding on top of the existing API.

Depends on whether https://github.com/asyncapi/parser-js/pull/785 adds a breaking change to the parser API. Havent finished it yet so not 100% sure.

Ideally, this should be a simple `feat` now and not a breaking change (since we already released the parser API changes).

>. And, if we don't have to release a major version, why don't we merge whatever degree of v3 support that we have now? As long as we don't "advertise" it as stable, it should be fine, right? I'm trying to get rid of this long-lived branch.

Agree.

src/models/channel.ts
```diff
@@ -7,7 +7,7 @@ import type { ServersInterface } from './servers';
 
 export interface ChannelInterface extends BaseModel, BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface {
   id(): string;
-  address(): string;
+  address(): string | null | undefined;
```
jonaslagoni, line 10: This is technically a breaking change to the parser API

jonaslagoni (commented):
Also, we might have to go through the files in this PR once before merging. Seems like it contains files no longer used. And there might be further changes we need to inspect.

Ontop of that, it seems like v3 does contain breaking changes to the parser API, see above comment in the code.

fmvilas:
Yeah, that's technically a breaking change. Now the question would be if we really want to be picky because nobody is using this version of the parser yet.

fmvilas:
Or actually, let me refrain from what I said above. What's the issue with releasing a new major version? I'm happy to bump the parser to v3 already, even though the life of v2 has been so short. What if we move the content of this branch to `next-major` so we can start getting release candidates on npm? The purpose of `next-major` was to release v2 but that's already done and this branch is going to produce a breaking change so then it's indeed a "next major". What do you think?

jonaslagoni:
I am all for releasing it ASAP as a major version 👍 Let's just clean up this PR first before merging it 😄

.github/workflows/if-nodejs-version-bump.yml
```diff
@@ -35,6 +35,13 @@ jobs:
           node-version: "${{ steps.lockversion.outputs.version }}"
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Setup Node.js
```
jonaslagoni, line 39: This should not be changed.

docs/api.md
```diff
@@ -0,0 +1 @@
+TBD
```
jonaslagoni, line 1: Huh?

src/models/v3/asyncapi.ts
```diff
@@ -1,7 +1,164 @@
...
+          messagesData.push(messageData);
+          messages.push(message);
+        }
+      });
+    });
+    return new Messages(messages);
+  }
+
+  schemas() {
+    return null as any;
```
jonaslagoni, line 93: This seems wrong

src/schema-parser/openapi-schema-parser.ts
```diff
@@ -0,0 +1,106 @@
+import Ajv from 'ajv';
+import { schemaV3 } from './openapi/schema_v3';
```
jonaslagoni, line 2: Didnt schema parsers get released separately and removed from the parser repo?

src/schema-parser/avro-schema-parser.ts
```diff
@@ -0,0 +1,354 @@
+import avsc from 'avsc';
```
jonaslagoni, line 1: Didnt schema parsers get released separately and removed from the parser repo?

src/schema-parser/raml-schema-parser.ts
```diff
@@ -0,0 +1,59 @@
+import yaml from 'js-yaml';
+import * as lib from 'webapi-parser';
+
+/* eslint-disable */
```
jonaslagoni, line 4: Didnt schema parsers get released separately and removed from the parser repo?

src/schema-parser/openapi/schema_v3.ts
```diff
@@ -0,0 +1,275 @@
+/* eslint-disable */
+// From https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/schemas/v3.0/schema.json
+export const schemaV3 ={
```
jonaslagoni, line 3: Pretty sure this can be removed as well

jonaslagoni:
Only thing I would wait for is https://github.com/asyncapi/parser-js/pull/785 as it most likely contain another breaking change in the API.

@fmvilas can you check out the above comments I made? Then I get 785 solved ASAP.

jonaslagoni:
Just occurred to me we also need to inspect the spectral rules

fmvilas:
> 

It doesn't appear here: https://www.npmjs.com/package/@asyncapi/parser?activeTab=versions.



> Only thing I would wait for is #785 as it most likely contain another breaking change in the API.
> 
> @fmvilas can you check out the above comments I made? Then I get 785 solved ASAP.

I'll go through them 👍 



> Just occurred to me we also need to inspect the spectral rules

What do you mean? I had to fix a few related things but not sure it's enough.

jonaslagoni:
> What do you mean? I had to fix a few related things but not sure it's enough.

Pretty sure that the spectral rules are only for v2: https://github.com/asyncapi/parser-js/tree/master/src/ruleset

This is fine for the first release I guess, just need to double-check how it works.

smoya:
@jonaslagoni @fmvilas I'm totally against releasing a new version (in this case a major version) that adds no value for users. This adds confusion now and in the future when people want to use the parser for v3 and strictly need to point to a version greater than `3.<whatever>` instead of just `v3.0`.

> It doesn't appear here: https://www.npmjs.com/package/@asyncapi/parser?activeTab=versions.

So then we should make it happen IMHO and go with that path.

src/schema-parser/avro-schema-parser.ts
smoya: For some reason, those are back here but they are removed from `master` branch 🤷

smoya: PR created removing them again https://github.com/asyncapi/parser-js/pull/802

.github/workflows/if-nodejs-version-bump.yml
smoya: I've spend time trying to figure out which merge was wrong, but couldn't replicate it.
I manually checked the code and I came up with the following PR that need to be merged to fix this:

- https://github.com/asyncapi/parser-js/pull/802
- https://github.com/asyncapi/parser-js/pull/804

docs/api.md
smoya: I've spend time trying to figure out which merge was wrong, but couldn't replicate it.
I manually checked the code and I came up with the following PR that need to be merged to fix this:

- https://github.com/asyncapi/parser-js/pull/802
- https://github.com/asyncapi/parser-js/pull/804

smoya:
@jonaslagoni I notice the prefix for the PR title includes the `!` so this will be considered as a major release. Do you know what's the BC here? I can't find any other commit including the exclamation mark in the message 🤔 and I don't recall atm where could the BC exist. Maybe in some of those `refactor:` commits.

jonaslagoni:
Dont think it's needed any longer no 🙂

src/models/v3/asyncapi.ts
smoya: Fixed in https://github.com/asyncapi/parser-js/pull/836

fmvilas:
Is there anything preventing us from releasing this already? We may want to wait for https://github.com/asyncapi/spec-json-schemas/pull/208 but other than that, it looks ready.

fmvilas (dismissed):
From my perspective, this can already be merged.

jonaslagoni:
Think this is ready to be reviewed and merged right? 🤔 @derberg @smoya @magicmatatjahu

smoya (approved):

smoya (approved):

fmvilas:
Let's go GitHub Actions. Show me what's you're capable of 💥

smoya:
> Let's go GitHub Actions. Show me what's you're capable of 💥 

Pleeeeeaaae!🙏

smoya:
> 2023-12-01T09:24:32.888Z INFO  Skipping PR merge, blocking label present: do-not-merge

PR #529, opened 2022-04-20
chore: fix merge conflicts with master branch

smoya:


derberg:
problem is that now we do not have `dist` directory. you need to create `bundle.js` from master version otherwise tests will fail

smoya:
> problem is that now we do not have `dist` directory. you need to create `bundle.js` from master version otherwise tests will fail

`dist`dir restored from `master` branch.

derberg (approved):

PR #476, opened 2022-02-21
feat: add methods and tests for serverVariables

danielkocot:
**Related issue(s)**
asyncapi/spec#707, asyncapi/spec#717, asyncapi/spec-json-schemas#174

PR #531, opened 2022-04-21
fix: broken package-lock.json

smoya:
**Description**

A malformed `package-lock.json` file has been merged on https://github.com/asyncapi/parser-js/pull/476/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R63.

We didn't realize it until it was merged. See https://github.com/asyncapi/parser-js/runs/6110565450?check_suite_focus=true
I think the whole set of github actions didn't run on the PR, maybe due to the fact it was the first time the author made a PR (and owner had to approve, but didn't happen).

Anyway, this PR is fixing it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@asyncapi/parser",
-      "version": "1.15.0-2022-04-release.1",
+      "version": "1.15.0-2022-04-release.2",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


jonaslagoni (approved):
This can happen indeed when there is no branch protection ensuring the jobs must go through. Same with at least one review, i.e. I could merge this one right now, and so can you by using `/rtm` as it does not require anything.
billede


Going forward with `next-spec` and `next-major-spec` this will be avoided as we no longer create branches as we go and no need to keep updating branch protection rules.

PR #542, opened 2022-05-04
refactor: port message, operation, correlation-id and security-requirements models

magicmatatjahu:
**Description**

- port `Message`, `MessageTrait`, `Operation`, `OperationTrait`, `CorrelationId` and `SecurityRequirements` models
- add unit tests

**Related issue(s)**
Part of #481 
Part of #482 

cc @smoya @Souvikns


jonaslagoni (approved):
👍

Souvikns (approved):
I haven't read all the files

src/models/channel-parameter.ts
```diff
@@ -1,3 +1,11 @@
 import type { BaseModel } from "./base";
+import type { DescriptionMixinInterface, ExtensionsMixinInterface } from "./mixins";
+import type { SchemaInterface } from "./schema";
 
-export interface ChannelParameterInterface extends BaseModel {}
+export interface ChannelParameterInterface extends BaseModel, DescriptionMixinInterface, ExtensionsMixinInterface {
+  id(): string;
+  hasSchema(): boolean;
```
smoya, line 7: If this should stay, should be added to https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#channelparameter

src/models/channel-parameter.ts
```diff
@@ -1,3 +1,11 @@
...
+  schema(): SchemaInterface | undefined;
+  hasLocation(): boolean;
+  location(): string | undefined;
```
smoya, line 10: Same as before, If this should stay, should be added to https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#channelparameter

src/models/message-trait.ts
```diff
@@ -1,3 +1,24 @@
 import type { BaseModel } from "./base";
+import type { CorrelationIdInterface } from "./correlation-id";
+import type { MessageExamplesInterface } from "./message-examples";
+import type { BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface } from './mixins';
+import type { SchemaInterface } from "./schema";
 
-export interface MessageTraitInterface extends BaseModel {}
+export interface MessageTraitInterface extends BaseModel, BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface {
```
smoya, line 7: There are some missing methods from the Parser-API that I think should be implemented here (otherwise, discarded from the Parser-API).

- hasKnownSchemaFormat() : `boolean`
- schemaFormat() : `string`

See https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#message

src/models/operation-trait.ts
```diff
@@ -1,3 +1,14 @@
 import type { BaseModel } from "./base";
+import type { BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface } from './mixins';
+import type { OperationKind } from "./operation";
+import type { SecurityRequirementsInterface } from "./security-requirements";
 
-export interface OperationTraitInterface extends BaseModel {}
+export interface OperationTraitInterface extends BaseModel, BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface {
+  id(): string;
+  kind(): OperationKind;
```
smoya, line 8: This should be added into the Parser-API https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#operation, as well as removing some of the methods related to Publish/Subscribe

smoya: Also, `kind` VS `type`. For example, in Security Requirements, we use `type` https://github.com/asyncapi/parser-js/pull/542/files#diff-dfe2dcce2b8c529ef868faf21548e2ba16163b0c4b9ef33129ac286346f2e008R15

src/models/v2/message-trait.ts
```diff
@@ -0,0 +1,98 @@
...
+  constructor(
+    private readonly _id: string,
+    _json: Record<string,any>,
+    _meta: ModelMetadata = {} as any
+  ) {
+    super(_json, _meta);
+  }
+
+  id(): string {
+    return this.messageId() || this._id;
```
smoya, line 30: Noice!!! 👍

src/models/v3/security-requirements.ts
```diff
@@ -0,0 +1,16 @@
+import { Collection } from '../collection';
+
+import type { SecurityRequirementsInterface } from '../security-requirements';
+import type { SecurityRequirementInterface } from '../security-requirement';
+
+export class SecurityRequirements extends Collection<SecurityRequirementInterface> implements SecurityRequirementsInterface {
+  override get(id: string): SecurityRequirementInterface | undefined {
+    // return this.collections.find(trait => trait.id() === id);
```
smoya, line 8: Is this a TODO?

src/models/channel-parameter.ts
magicmatatjahu: The channel parameter has a location so it has to stay 😄

src/models/message-trait.ts
magicmatatjahu: I didn't add it on purpose for a simple reason. Defining `schemaFormat` at the messag level on a lot of problems from a schema reuse point of view, and I will try to move schemaFormat to the Schema object - then this will be removed. Here is my issue for 3.0.0 https://github.com/asyncapi/spec/issues/622. If we want it I can add it, but I hope that in 99% we will remove it.

smoya (changes requested):
Mainly looks good to me! 🎉 Left some comments and suggestions

src/models/operation-trait.ts
magicmatatjahu: You're right, we have inconsistencies. `type` in SecurityRequirements and here `kind`. In 3.0.0 we have for example called it as `action` (https://github.com/asyncapi/spec/issues/618) so maybe we should call it `action`?

src/models/v2/message-trait.ts
magicmatatjahu: Here I have a problem, because on one side we call it `id`, and next to it we have `messageId`. Maybe we should rename this method to `uid` and for `id` return the value for `messageId`? Same for other objects - where we call the method `id` we should call it `uid`?

src/models/v3/security-requirements.ts
magicmatatjahu: Yes, and no. Because of the fact that these objects do not have `id` I wonder if we should return `undefined`, or change this collection to a regular array and not collection class.

src/models/message-trait.ts
smoya: It makes sense indeed. How would you implement that? I think this should be then done in this PR

src/models/channel-parameter.ts
smoya: Would you mind then creating a PR on Parser-API repo?

src/models/v2/message-trait.ts
smoya: The `messageId` is the real **unique id** (uid) across the document, so I think it would be the other way around.
Anyway, I see the issue you are mentioning. 
Maybe, adding a new `UniqueIdentifiableInterface`, with the `uid` method would make sense. In the case of message, to return `this.messageId()`, `this.operationId` for Operation, etc.
The `id` field will then just return the key of the map which the object belongs to. 

Makes sense?

src/models/v3/security-requirements.ts
smoya: Good point. Can't we do just directly use `Record<string, { schema: SecuritySchemeInterface, scopes: Array<string> }>` ?

src/models/operation-trait.ts
smoya: Indeed, `action` sounds like a better fit.

src/models/v3/security-requirements.ts
magicmatatjahu: We cannot, `security` field (in security and operation) is an array, so we ca change it to the 

```ts
Array<Record<string, { schema: SecuritySchemeInterface, scopes: Array<string> }>>
```

src/models/v2/message-trait.ts
magicmatatjahu: It makes sense, but I wonder if we need 3 methods, `messageId`, `id` and `uid`. 

The `id` could be for `messageId`
The `uid` for the `messageId` or key or some generated unique id for given reference.

src/models/message-trait.ts
magicmatatjahu: I can add `schemaFormat` in this PR to `Message` model, but we can always remove it.

smoya: Yeah, makes sense.

I like the `message.payload().schemaFormat()` you suggested in https://github.com/asyncapi/spec/issues/622. Maybe we can move forward with it later on, and now just adding an issue to work on it next?

src/models/v2/message-trait.ts
smoya: I think we are considering a different thing for what the `id` value is.

From my pov, `id` is just the key on the map. For example, `myMessage` in the following example:

```yaml
components:
  messages:
    myMessage:
      # etc
```

Of course, currently messages declared inside the operation won't have an id (at least on spec v2.x) but this could change for next versions (and makes sense to me).

src/models/v3/security-requirements.ts
smoya: I think makes sense

magicmatatjahu:
@smoya I propagated suggestions from your review. I didn't add `uid` and stay with `id` and `messageId`. I think that now it's good approach, we can always change it when we will see how 3.0.0 look like.

I also updated `parser-api` https://github.com/asyncapi/parser-api/pull/64

src/models/message-trait.ts
magicmatatjahu: Issue :) https://github.com/asyncapi/parser-js/issues/544

PR #543, opened 2022-05-04
feat: new options to generate nice $id (#516)

### LINKED ISSUE #516: Better x-parser-schema-id
#### Reason/Context
If there is not proper "x-parser-schema-id", meaning those "anonym..",
[Modelina ](https://github.com/asyncapi/modelina) will end up with unusable classes.
But this is an issue for all code generator where you are not owner of the schema.

There are 2 Issues:
- at the moment proper $id will only be generated by first level schemas. All sub schemas used in arrays or similar loose their name.
- For string enum the $id could be generated by the field name, if not set. 

I would suggest this to be enabled by an options to make it non breaking.

GreenRover:
This should be used when using it with a code generator like modelina

**Description**

If there is not proper "x-parser-schema-id", meaning those "anonym..",
[Modelina ](https://github.com/asyncapi/modelina) will end up with unusable classes.
But this is an issue for all code generator where you are not owner of the schema.

There are 2 Issues:

at the moment proper $id will only be generated by first level schemas. All sub schemas used in arrays or similar loose their name.
For string enum the $id could be generated by the field name, if not set.
I would suggest this to be enabled by an options to make it non breaking.

**Related issue(s)**
Resolves #516


derberg:
Before I do deeper review can you please explain the idea behind this function.

- you expose it as standalone function, why not using it only in `asyncapi.js` during construction? like we do with anonymous
- you invoke this function only on `components.schemas`, why not on `message.payload` too ?

GreenRover:
> Before I do deeper review can you please explain the idea behind this function.
As explained in latest community meeting where you was in vacation ;-)
The main idea/problem to be solved is:
- i am not owner of the schema. But responsible for the tooling
-  a schema not needs to have $id to be valid
- caused by a bug if there are no $id only for root schemas the schema name would be used but not for childs. This happends caused by dereference.
- Avery schemas without $id will be converted by modelina to AnnonymousClass-123.java
- Speaking class names is very important for a developer.
- My opinion : punish a api consumer for bad api design is not god
The intention is: make modelina generate best quality code as possible regardless of the api quality.
Yes i see the generated class names may change when refactoring the api. But again: lets do it as god as possible. Perfect will not be possible.

> * you expose it as standalone function, why not using it only in `asyncapi.js` during construction? like we do with anonymous
Can you give me a line number to an example please?
> * you invoke this function only on `components.schemas`, why not on `message.payload` too ?
God point i will change it

GreenRover:
> you expose it as standalone function, why not using it only in asyncapi.js during construction? like we do with anonymous
Ok i remoted the `generateIdsInSpec` from `types.d.ts` now it should be the same as `dereference`.
A anonymous function is god here:
- it is to big to be Inlined (code quality)
- it is recursive, this would be extra hard to read if being anonymous 

> you invoke this function only on components.schemas, why not on message.payload too ?
Was changed.

derberg:
> As explained in latest community meeting where you was in vacation ;-)

Oh, sorry, I did not make myself clear with my question. Btw I watched the recording 😉 
So yeah, I think all parser maintainers are aware of the issue with `anonymous` id's. There is this issue -> https://github.com/asyncapi/parser-js/issues/176

> Before I do deeper review can you please explain the idea behind this function.

☝🏼 was just an intro sentence for questions in bullet points 😓 

---

Ok, so just to make sure, before I share feedback to the code.

The current idea is: user passes special option to get `$id` generated in schemas, instead of the anonymous ones. And the assignment of ids is done before construction of `AsyncAPIDocument`.

My question is: what is the use case to do it this way? We already assign ids here -> https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L39-L42 ? Why not improve current assignment, which is already handling recursion and will not fail with documents that contain circular refs.

Maybe I'm missing something. Maybe you could provide example AsyncAPI document before and after Ids assignment?

> you invoke this function only on components.schemas, why not on message.payload too ?
Was changed.

you did it for `components.messages` and I was referring to `message.payload`

---

last but not least, please do not force push as it is hard to see what actually has changed since last review 🙏🏼 

@jonaslagoni @fmvilas looping you in as you were at the community call discussing this with @GreenRover

GreenRover:
> We already assign ids here -> https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L39-L42 
Do it here would be most reasonable but will not work. Because the `dereference` will convert from:
```yaml
asyncapi: 2.0.0
info:
  title: Account Service
  version: '1.0.0'

channels:
  user/signedup:
    subscribe:
      operationId: emitUserSignUpEvent
      message:
        $ref : '#/components/messages/UserSignedUp'

components:
  messages:
    UserSignedUp:
      name: userSignedUp
      title: User signed up event
      summary: Inform about a new user registration in the system
      contentType: application/json
      payload:
        $ref: '#/components/schemas/userSignedUpPayload'

  schemas:
    userSignedUpPayload:
      type: object
      properties:
        firstName:
          type: string
          description: "foo"
        lastName:
          type: string
          description: "bar"
        email:
          $ref: '#/components/schemas/eMail'
        createdAt:
          $ref: '#/components/schemas/dateTime'
    dateTime:
      type: string
      format: date-time 
    eMail:
      type: string
      format: email
      description: "baz"
```

to:
```yaml
asyncapi: 2.0.0
info:
  title: Account Service
  version: '1.0.0'

channels:
  user/signedup:
    subscribe:
      operationId: emitUserSignUpEvent
      message:
        name: userSignedUp
        title: User signed up event
        summary: Inform about a new user registration in the system
        contentType: application/json
        payload:
          type: object
          properties:
            firstName:
              type: string
              description: "foo"
            lastName:
              type: string
              description: "bar"
            email:
              type: string
              format: date-time 
            createdAt:
              type: string
              format: email
              description: "baz"

components:
  messages:
    UserSignedUp:
      name: userSignedUp
      title: User signed up event
      summary: Inform about a new user registration in the system
      contentType: application/json
      payload:
        type: object
        properties:
          firstName:
            type: string
            description: "foo"
          lastName:
            type: string
            description: "bar"
          email:
            type: string
            format: date-time 
          createdAt:
            type: string
            format: email
            description: "baz"

  schemas:
    userSignedUpPayload:
      type: object
      properties:
        firstName:
          type: string
          description: "foo"
        lastName:
          type: string
          description: "bar"
        email:
          type: string
          format: date-time 
        createdAt:
          type: string
          format: email
          description: "baz"
    dateTime:
      type: string
      format: date-time 
    eMail:
      type: string
      format: email
      description: "baz"
```

Meaning this line:
https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L39
makes no sense at all. Because what ever was changed in `components.schemas` needs to be exact the same as in `components.messages` and `channels[].message`  because everything was copied over.

In short,  where the id generation is done actually is totally nice to read code but it is to late.

Second point: after `dereference`  i lost in `channels[].message`  the information that the schema part was named `components/schemas/eMail` originally. So i last one of the best hint for generation anonymous ids.

What modelina is using the `channels`
https://github.com/asyncapi/modelina/blob/4e8d9768bf3c311ba5732c0bbe5f1851df63a71b/src/processors/AsyncAPIInputProcessor.ts#L32
https://github.com/asyncapi/parser-js/blob/dcee64944343f6fb74566e3905680d133946332f/lib/models/asyncapi.js#L170

GreenRover:
@derberg I changed it to use 
https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L39-L42

But now the code is harder to read.
Please see the test case for an example.

Clean up git history will be done when merge ready.

lib/parser.js
```diff
@@ -6,7 +6,7 @@ const $RefParser = require('@apidevtools/json-schema-ref-parser');
 const mergePatch = require('tiny-merge-patch').apply;
 const ParserError = require('./errors/parser-error');
 const { validateChannels, validateTags, validateServerVariables, validateOperationId, validateServerSecurity, validateMessageId } = require('./customValidators.js');
-const { toJS, findRefs, getLocationOf, improveAjvErrors, getDefaultSchemaFormat } = require('./utils');
+const { toJS, findRefs, getLocationOf, improveAjvErrors, getDefaultSchemaFormat, ucFirst } = require('./utils');
```
derberg, line 9: we do not use ucFirst here anymore

lib/utils.js
```diff
@@ -320,3 +320,13 @@ utils.getUnknownServers = (parsedJSON, channel) => {
...
   return `application/vnd.aai.asyncapi;version=${asyncapiVersion}`;
 };
+
+/**
+ * returns a string with the first character of string capitalized, if that character is alphabetic.
+ * 
+ * @function ucFirst
+ * @param {String} word 
+ */
+utils.ucFirst = (word) => {
```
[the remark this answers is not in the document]
derberg: also please mark it as private with `@private` so we do not get it in API.md

derberg (changes requested):
sorry for keeping you waiting, I was at a conference and it delayed me

- I left some individual comments
- please solve merge conflicts, `dist/bundle.js` can be removed as it is not needed to be present in the repo anymore
- please fix code smells from sonar
- > Clean up git history will be done when merge ready. -> no need, we squash and merge
- `x-parser-schema-id-level` needs to be documented in readme

once all is addressed I'll chase another reviewer to have a look

lib/utils.js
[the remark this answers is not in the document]
GreenRover: done

lib/parser.js
GreenRover: removed

GreenRover:
@derberg Can you please review again?

lib/anonymousNaming.js
```diff
@@ -43,22 +49,95 @@ function assignUidToParameterSchemas(doc) {
...
+        assignUidToComponentSchemasRecursive(schema, key);
+      } else {
+        schema.json()[String(xParserSchemaId)] = key;
+      }
     }
   }
 }
 
-/**
+function assignUidToComponentSchemasRecursive(schema, key) {
```
derberg, line 72: please add jsdoc like for other functions,

derberg (commented):
@GreenRover some code smells are still there https://sonarcloud.io/project/issues?id=asyncapi_parser-js&pullRequest=543&resolved=false&types=CODE_SMELL

I also left 2 more comments

@jonaslagoni can you please have a look as well, especially that we are touching the part that you refactored in the past and easier for you will be to notice potential future bugs that I do not see

jonaslagoni (commented):
I tried to use the `zbos_mqtt-all-asyncapi` file, because all it generates are anonyms id's. It seems like it still does, even though I expected it to now have more meaningful id's.

```js
const inputYAMLVeryComplexSchema2 = fs.readFileSync(path.resolve(__dirname, './good/zbos_mqtt-all-asyncapi.json'), 'utf8');
it('should generate ids 2', async function () {
  const result = await parser.parse(inputYAMLVeryComplexSchema2, { path: __filename, genererateIdInSchema: true });
  const json = JSON.stringify(result.json(), null, 2).trim();
  fs.writeFileSync(path.resolve(__dirname, './good/zbos_mqtt-all-asyncapi-output.json'), json);
});
```

Am I missing something? 🤔 

Had two comments otherwise, but nothing major 🙂

types.d.ts
```diff
@@ -194,12 +194,12 @@ declare module "@asyncapi/parser" {
         hasCircular(): boolean;
         /**
          * Traverse schemas in the document and select which types of schemas to include.
-         * By default all schemas are iterated
+        By default all schemas are iterated
```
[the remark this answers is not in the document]
GreenRover: removed from commit

lib/anonymousNaming.js
GreenRover: done

README.md
```diff
@@ -206,6 +206,7 @@ The parser uses custom extensions to define additional information about the spe
 - `x-parser-original-schema-format` holds information about the original schema format of the payload. You can use different schema formats with the AsyncAPI documents and the parser converts them to AsyncAPI schema. This is why different schema format is set, and the original one is preserved in the extension.
 - `x-parser-original-payload` holds the original payload of the message. You can use different formats for payloads with the AsyncAPI documents and the parser converts them to. For example, it converts payload described with Avro schema to AsyncAPI schema. The original payload is preserved in the extension.
 - [`x-parser-circular`](#circular-references)
+- `x-parser-schema-id-level` Is internally used to generate use friendly `x-parser-schema-id` if option `genererateIdInSchema` is set. It indicates the traversal level of a schema. This is used to prefer name of well-defined schemas.
```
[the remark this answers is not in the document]
GreenRover: i use your proposal

GreenRover:
> Am I missing something

The sample `zbos_mqtt-all-asyncapi` defines 2 schema: "percentage" and "key" but all the big rest is defined anonymous in the `channels` section. The approach only works if you use `components.schemas` and only reference your schema at `channels` and `components.messages` section. Otherwise there is no good starting name and the whole system gives up.

jonaslagoni:
> @derberg @jonaslagoni What is the editor and editor settings you use for this project?
> vscode will always auto format all files for me, caused by the unequal coding styles.

I never use any auto-formatting (something that can be disabled in VSCode, search for `format on save`), and just fix whatever warn/errors the linter gives me.

lib/anonymousNaming.js
```diff
@@ -43,22 +49,95 @@ function assignUidToParameterSchemas(doc) {
...
+    schema,
+    key,
+    {
+      parentId: '',
+      level: 0,
+      propOrIndexModifier: (propertyName, schema, options) => {
+        if (typeof propertyName === 'number') {
+          // It is not save to generate ids with arrays containing multiple types
+          return null;
+        }
```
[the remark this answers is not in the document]
GreenRover: I improved comment

GreenRover:
@derberg @jonaslagoni I updated the PR and tried to respect all your inputs. Can you review again?
There is still one open change request in githup but i have no clue have to acknowledge the change request.

README.md
```diff
@@ -206,6 +206,7 @@ The parser uses custom extensions to define additional information about the spe
 - `x-parser-original-schema-format` holds information about the original schema format of the payload. You can use different schema formats with the AsyncAPI documents and the parser converts them to AsyncAPI schema. This is why different schema format is set, and the original one is preserved in the extension.
 - `x-parser-original-payload` holds the original payload of the message. You can use different formats for payloads with the AsyncAPI documents and the parser converts them to. For example, it converts payload described with Avro schema to AsyncAPI schema. The original payload is preserved in the extension.
 - [`x-parser-circular`](#circular-references)
+- `x-parser-schema-id-level` is internally used to generate use friendly `x-parser-schema-id` if `genererateIdInSchema` option is passed to the parser. It indicates the traversal level of a schema. This is used to prepare the name of well-defined schemas.
```
smoya, line 209:
```suggestion
- `x-parser-schema-id-level` is internally used to generate user-friendly `x-parser-schema-id` if `genererateIdInSchema` option is passed to the parser. It indicates the traversal level of a schema. This is used to prepare the name of well-defined schemas.
```

I would like to also explain what "the traversal level of a schema" is in this context if possible.

lib/anonymousNaming.js
```diff
@@ -43,42 +49,170 @@ function assignUidToParameterSchemas(doc) {
...
+        options.parentType = subSchema.type();
+        options.level++;
+
+        if (!subSchema.$id()) {
+          if (subSchema.json()[String(xParserSchemaIdLevel)] && subSchema.json()[String(xParserSchemaIdLevel)] < options.level) {
+            // After deref we get a tree, always prefer names of parent root schema definitions.
+
+            // The usage of $ref will lead to having the same schema multiple times in the tree caused by deref.
+            // By walking through the tree, each level deper in the tree gets a higher xParserSchemaIdLevel
+            // So if we find a genereated id with a higher xParserSchemaIdLevel the the current, this id was generated because the current schema was used via $ref as subelement in another schema.
```
smoya, line 156:
```suggestion
            // So if we find a generated id with a higher xParserSchemaIdLevel the the current, this id was generated because the current schema was used via $ref as subelement in another schema.
```

Also I don't really understand what the sentence says after `xParserSchemaIdLevel` word.

lib/anonymousNaming.js
```diff
@@ -43,42 +49,170 @@ function assignUidToParameterSchemas(doc) {
...
+            // and we prefer the generated id that is clother to a rood node in tree.
```
smoya, line 157:
```suggestion
            // and we prefer the generated id that is closer to a root node in the tree.
```

Is that what you meant?

lib/anonymousNaming.js
```diff
@@ -43,42 +49,170 @@ function assignUidToParameterSchemas(doc) {
...
+            return false;
+          }
+
+          if (subSchema.type() === 'array') {
+            // You want the shorter on the child, this is used by the code generator to define the class name.
```
smoya, line 162: I'm sorry to say this, but I don't get what you mean by:

> You want the shorter on the child

Not wanting to sound picky, really. I just want to understand this piece of code.

lib/anonymousNaming.js
```diff
@@ -43,42 +49,170 @@ function assignUidToParameterSchemas(doc) {
...
+          } else {
+            subSchema.json()[String(xParserSchemaId)] = propOrIndex;
+          }
+          subSchema.json()[String(xParserSchemaIdLevel)] = options.level;
+        }
+      },
+      schemaTypesToIterate: [
+        'objects',
+        'arrays',
+        'allOf' // Will inheritate the parent id, because it is a joined object.
```
smoya, line 188:
```suggestion
        'allOf' // Will inherit the parent id, because it is a joined object.
```

lib/anonymousNaming.js
GreenRover: If you have the array and the array child. As in the example below this line.
There will be 2 ids generated.
The id with the shorter name you want to have on the array child because this is going to be a class name when code is generated. The array it self is just a class member and will be our of interested when it comes to code generation.

In short: you want shorted speaking names for your generated classes.

GreenRover:
@[smoya](https://github.com/smoya) Thanks for all your spell checks and improvement suggestions!

test/good/asyncapi-very-complex-schema-output.json
```diff
@@ -0,0 +1,876 @@
...
+                          "A++",
+                          "A+",
+                          "A",
+                          "B",
+                          "C",
+                          "D",
+                          "E",
+                          "F"
+                        ],
+                        "x-parser-schema-id": "<anonymous-schema-9>"
```
jonaslagoni, line 108: Code generation would generate an enum here, which would have an anonymous schema id.

test/good/asyncapi-very-complex-schema-output.json
```diff
@@ -0,0 +1,876 @@
...
+                      }
+                    },
+                    "x-parser-schema-id": "<anonymous-schema-8>"
```
jonaslagoni, line 111: Code generation would generate an "class" here, which would have an anonymous schema id.

test/good/asyncapi-very-complex-schema-output.json
```diff
@@ -0,0 +1,876 @@
...
+                  },
+                  {
+                    "type": "object",
+                    "properties": {
+                      "demo": {
+                        "type": "string",
+                        "x-parser-schema-id": "<anonymous-schema-11>"
+                      }
+                    },
+                    "x-parser-schema-id": "<anonymous-schema-10>"
```
jonaslagoni, line 142: Code generation would generate a "class" here, which would have an anonymous schema id.

test/good/asyncapi-very-complex-schema-output.json
```diff
@@ -0,0 +1,876 @@
...
+                  },
+                  {
+                    "type": "object",
+                    "properties": {
+                      "demo": {
+                        "type": "string",
+                        "x-parser-schema-id": "<anonymous-schema-13>"
+                      }
+                    },
+                    "x-parser-schema-id": "<anonymous-schema-12>"
```
jonaslagoni, line 173: Code generation would generate a "class" here, which would have an anonymous schema id.

jonaslagoni (changes requested):
You have a linter error 🙂 

Just want to highlight some schemas that something like Modelina would generate models from, that still have anonymous ids. Not sure it's something we want to fix, just wanted to highlight it.

Otherwise it looks fine to me 👍

test/good/asyncapi-very-complex-schema-output.json
GreenRover: It is a child of an anyOf. I have no clue how to solve this.

test/good/asyncapi-very-complex-schema-output.json
GreenRover: It is a child of an oneOf. I have no clue how to solve this.

test/good/asyncapi-very-complex-schema-output.json
GreenRover: Edge case: allOf child enum. Should be something like.
Could be something like: streetLightLightbulbEcologically but i am not sure if a as code owner would be happy with the name. IF some one would like to have that being solved. That there needs be be another PR.

GreenRover:
PR takes to long for me. In my opinion it is better to have small improvements than discuss ages and have no improvements at all.
I will live with my fork.

PR #547, opened 2022-05-06
refactor: port channel, channel-parameter and schema models

magicmatatjahu:
**Description**

Port `Channel`, `Channel-parameter` and `Schema` models.

**Related issue(s)**
Part of #481 
Part of #482


magicmatatjahu:
@smoya Don't worry, if Fran/Jonas reply to our thoughts I will update my PR and add also the missing methods from this PR :) Thanks for review!

PR #572, opened 2022-07-19
feat: parseFromUrl does not resolve relative references (#504)

### LINKED ISSUE #344: Parser does not work out of the box in non-node environments (websites)
#### Describe the bug
If the parser is utilized in a website the `path` parameters is a requirement to set, in order to make the library work.

#### How to Reproduce
Given the AsyncAPI document:
```
{
    "asyncapi": "2.1.0",
    "info": {
        "title": "Streetlights API",
        "version": "1.0.0",
        "description": "The Smartylighting Streetlights API allows you\nto remotely manage the city lights.\n",
        "license": {
            "name": "Apache 2.0",
            "url": "https://www.apache.org/licenses/LICENSE-2.0"
        }
    },
    "servers": {
        "mosquitto": {
            "url": "mqtt://test.mosquitto.org",
            "protocol": "mqtt"
        }
    },
    "channels": {
        "light/measured": {
            "publish": {
                "summary": "Inform about environmental lighting conditions for a particular streetlight.",
                "operationId": "onLightMeasured",
                "message": {
                    "name": "LightMeasured",
                    "payload": {
                        "type": "object",
                        "$id": "LightMeasured",
                        "properties": {
                            "id": {
                                "type": "integer",
                                "minimum": 0,
                                "description": "Id of the streetlight."
                            },
                            "lumens": {
                                "type": "integer",
                                "minimum": 0,
                                "description": "Light intensity measured in lumens."
                            },
                            "sentAt": {
                                "type": "string",
                                "format": "date-time",
                                "description": "Date and time when the message was sent."
                            }
                        }
                    }
                }
            }
        }
    }
}
```

and the following code:
```
const parsedInput = await parse(input)
```
which forces the `path` parameter:
```
const parsedInput = await parse(input, {path: './'})
```

This is because the following code: https://github.com/asyncapi/parser-js/blob/aba3ed25a58e1eb36b77d5e7dd724202ca737f64/lib/parser.js#L45

Where in a browser environment `process.cwd()` returns `/` and `path.sep` returns `/` which together (`//`) is an incorrect path and it will throw an exception once it begins the dereference process.

#### Expected behavior
Expected the parser to work out of the box without any modification and ensure the default path is always valid.

### LINKED ISSUE #504: parseFromUrl does not resolve relative references
#### Describe the bug
Given the use of `parseFromUrl` I would expect that if the remote document contained relative references, they would be resolved based on the URL passed to `parseFromUrl` 🤔 

#### How to Reproduce

Given a reference such as `"$ref": "./components/parameters.json#/server_id"` in an AsyncAPI document located at `https://raw.githubusercontent.com/org/repo/main/asyncapi.json` I would expect it to fetch the parameter from `https://raw.githubusercontent.com/org/repo/main/components/parameters.json#/server_id` instead of trying to fetch it locally on the machine.

aeworxet:
- `getBaseUrl()` was made a separate util to make code reuse possible.
- URL validation in it is not performed because `node-fetch` performs its own validation at fetch time, so no repetition of this task is made. The function only ensures that `url` has type of `string` and lets `node-fetch` deal with the rest.
- Condition `(typeof window !== 'undefined' && !options.hasOwnProperty('path'))` was added for the case if `getBaseUrl()` gets called directly from inside of a browser (not from `parseFromUrl()`, but from `<script>`) AND, amongst all, object `options` doesn't already have in it its own property `path` with value. It is assumed in this case that `getBaseUrl()` was not executed yet, so it's executed.
If `getBaseUrl()` gets called from inside of a browser AND object `options` HAS in it its own property `path` with a value - OK then, somehow it got that value, so `getBaseUrl()` does not get executed. This construction was inspired by test 'parse from string' in `./test/sample_browser/index.html`.
It is also assumed that, at the end of the day, user could have provided his own version of `options.path` - injected it at some point, for example.
- `<`/`>` in `<anonymous-schema-x>` were considered custom HTML tags by browser's parser and didn't appear on screen

making automated tests that assert visible elements' content to fail. So they are changed to HTML entities `&lt;`/`&gt;` before insertion into the DOM and are replaced back to `<`/`>` by browser's parser at the insertion time, thus allowing automated tests querying elements' content to pass.
- Assertion tests are made big intentionally, so they test big chunks of functionality at once, and if the test fails, one, using difftool, can quickly peek what was parsed wrong and get an idea of where to look.
- Updated `puppeteer` to latest version (v17.0.0 at the moment of writing) just to make it more up to date.

Fixes #344
Fixes #504


derberg (commented):
@aeworxet thanks for the PR, there is one code smell to fix, please have a look https://sonarcloud.io/project/issues?id=asyncapi_parser-js&pullRequest=572&resolved=false&types=CODE_SMELL

@fmvilas @jonaslagoni @magicmatatjahu 
hey folks, I'm not gonna jump into review of this one. I synced with @aeworxet offline several times to discuss possible solution, and yeah, I'm not the best person to approve what I already consulted and like 😆 

so yeah, please review, especially @jonaslagoni that reported the request.

tl;dr it is about having similar solution for refs that we already have for parser, when parsing AsyncAPI file from the file system, to set base path to refs. In case of this PR it is about setting base path for refs but when parsing AsyncAPI file from URL or in a browser

aeworxet:
@derberg 
Fixed.

tapan3:
We are facing same issue, so curious to know when this PR could be merged and available to use?

jonaslagoni (approved):
LGTM 👍

derberg (approved):
@jonaslagoni no, I was most interested with your input as you did report the related issue.

we just need to update PR description properly so GitHub closes related issues.

So, you think we can close https://github.com/asyncapi/parser-js/issues/344 with this one too?

jonaslagoni:
> @jonaslagoni no, I was most interested with your input as you did report the related issue.
> 
> we just need to update PR description properly so GitHub closes related issues.
> 
> So, you think we can close #344 with this one too?

You right, yea, I think we can, at least that's what I am gonna do 👍

jonaslagoni:
@aeworxet just added the following to automatically close the issues once released 🙂 
```
Fixes https://github.com/asyncapi/parser-js/issues/344
Fixes https://github.com/asyncapi/parser-js/issues/504
```

Issue #480, opened 2022-03-03
New interface for custom schema parser


magicmatatjahu:
> It will introduce breaking change so this should be done when rewriting the parser to the IntentAPI and TS

#### Reason/Context

Currently, custom parsers for the massage's payload (and in the `3.0.0` AsyncAPI version this will be made available to all schemas) have two methods: `parse()`, which parses the given format into an AsyncAPI JSON Schema, which is later used in ModelAPI (in `2.0.0` version of lib it will be IntentAPI), and `getMimeTypes()` which returns the mime types that specify the given schema format. We are missing 3 functions `validate`, which would return possible errors if a given schema has syntax errors.

#### Description

We should update all custom schema parsers to have that third function and it should be mandatory.

Due to the fact that RAML/OpenAPISchema/AVRO can be described using YAML or JSON we can easily validate them using JSON Schema and AJV and then combine the validation with Spectral from Stoplight (see more https://github.com/asyncapi/parser-js/issues/477).

cc @jonaslagoni @smoya

magicmatatjahu:
I wonder if there should be also a method that would validate the example based on the schema (use case https://github.com/asyncapi/studio/issues/200). We can of course convert the format to AsyncAPI JSON Schema and validate on that, but I wonder if it would make sense. It should be discussed.

smoya:
@magicmatatjahu I would love to work on this issue so I can get more knowledge on how Stoplight integrates with the Parser. It would need, however, to happen next week. Hope you are ok with that @magicmatatjahu 🙏

magicmatatjahu:
Issue is resolved in v2 ParserJS - now in https://github.com/asyncapi/parser-js/tree/next-major branch

PR #573, opened 2022-07-21
refactor: add AsyncAPI Schema Parser

smoya:
**Description**

This PR adds the AsyncAPI Schema Parser, which is compatible with Spectral validation functions.

**Related issue(s)**


src/types.ts
```diff
@@ -16,4 +17,5 @@ export interface DetailedAsyncAPI {
 }
 
 export type Diagnostic = ISpectralDiagnostic;
-export type SchemaValidateResult = IFunctionResult;
+export type SchemaValidateError = IFunctionResult;
```
smoya, line 20: I changed the name to be more explicit about what it is in fact... just an error.

package.json
```diff
@@ -52,6 +52,8 @@
     "@stoplight/spectral-functions": "^1.5.1",
     "@stoplight/spectral-parsers": "^1.0.1",
     "@stoplight/spectral-rulesets": "^1.4.3",
+    "@types/json-schema": "^7.0.11",
```
magicmatatjahu, line 55: Packages from `@types/` should be installed in the dev-dependencies.

src/types.ts
magicmatatjahu: It's not always an error, because from spectral you can also return info or warning severity, by this spectral team called it `IFunctionResult` not `IFunctionError`. Please change to the previous one.

src/schema-parser/asyncapi-schema-parser.ts
```diff
@@ -8,12 +15,63 @@ export function AsyncAPISchemaParser(): SchemaParser {
...
+  try {
+    ajv.compile(schema);
+  } catch (error: any) {
+    if (error! instanceof Error) {
+      errors = ajvToSpectralErrors(error);
+    } else {
+      // Unknown and unexpected error
+      throw error;
+    }
+  }
```
magicmatatjahu, line 31: I have thought it over again and it is not a good solution, from a simple fact. Here we validate the scheme using the normal draft 07 Schema, but in our Schema Object we have 3 additional fields that do not exist in the JSON Schema - https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#fixed-fields-21, so I recommend switching to the solution from our current parser where we validate the scheme using the schema stored in our json schemas, like here https://github.com/asyncapi/parser-js/blob/master/lib/asyncapiSchemaFormatParser.js#L68

src/types.ts
```diff
@@ -1,4 +1,5 @@
 import type { ISpectralDiagnostic, IFunctionResult } from '@stoplight/spectral-core';
+import { JSONSchema7 }from "json-schema"
```
magicmatatjahu, line 2:
```suggestion
import type { JSONSchema7 }from "json-schema"
```

magicmatatjahu (changes requested):
Good, but we need some improvements :)

src/types.ts
smoya: Rollbacked the rename! thanks for the catch

package.json
smoya: Good catch. Fixed!

src/schema-parser/asyncapi-schema-parser.ts
magicmatatjahu: @smoya Do you need help with that? :)

src/schema-parser/asyncapi-schema-parser.ts
```diff
@@ -8,12 +19,73 @@ export function AsyncAPISchemaParser(): SchemaParser {
...
+/**
+ * To validate schema of the payload we just need a small portion of official AsyncAPI spec JSON Schema, the definition of the schema must be
+ * a main part of the JSON Schema
+ * 
+ * @private
+ * @param {Object} asyncapiSchema AsyncAPI specification JSON Schema
+ * @param {Object} version AsyncAPI version.
+ * @returns {Object} valid JSON Schema document describing format of AsyncAPI-valid schema for message payload
+ */
+function preparePayloadSchema2(asyncapiSchema: AsyncAPISchema, version: string) {
```
smoya, line 75: Comes from https://github.com/asyncapi/parser-js/pull/423

src/schema-parser/asyncapi-schema-parser.ts
```diff
@@ -22,7 +94,8 @@ function getMimeTypes() {
     'application/schema+json;version=draft-07',
     'application/schema+yaml;version=draft-07',
   ];
-  ['2.0.0', '2.1.0', '2.2.0', '2.3.0'].forEach(version => {
+
+  specVersions.forEach((version: string) => {
```
smoya, line 98: Removing the need for adding a new version every time a new spec version is out...

test/schema-parser/asyncapi-schema-parser.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+            if: "this should be an if"
+          }
+        }
+      }
+    };
+
+    const result = await parser.validate(schema);
+    const expectedResult: SchemaValidateResult[] = [
+      { "message": "must be object,boolean", "path": ["components", "schemas", "schema1", "payload", "properties", "name", "if"] },
+      { "message": "must be array", "path": ["components", "schemas", "schema1", "payload", "oneOf"] }
```
smoya, line 69: I don't really get why AJV returns duplicated errors for this:

```js
[
      {
        instancePath: '/properties/name/if',
        schemaPath: '#/type',
        keyword: 'type',
        params: { type: [Array] },
        message: 'must be object,boolean'
      },
      {
        instancePath: '/oneOf',
        schemaPath: '#/type',
        keyword: 'type',
        params: { type: 'array' },
        message: 'must be array'
      },
      {
        instancePath: '/oneOf',
        schemaPath: '#/allOf/1/properties/oneOf/type',
        keyword: 'type',
        params: { type: 'array' },
        message: 'must be array'
      },
      {
        instancePath: '/properties/name/if',
        schemaPath: '#/type',
        keyword: 'type',
        params: { type: [Array] },
        message: 'must be object,boolean'
      }
    ]
```

src/schema-parser/asyncapi-schema-parser.ts
smoya: No, was reading code so I could better understand things. That's why I wrote https://github.com/asyncapi/parser-js/issues/481#issuecomment-1195965993.

Btw, the code of this PR has been updated with what I had until then. feel free to review it again 🙏

src/schema-parser/asyncapi-schema-parser.ts
magicmatatjahu, line 75:
```suggestion
function preparePayloadSchema(asyncapiSchema: AsyncAPISchema, version: string) {
```

src/schema-parser/asyncapi-schema-parser.ts
```diff
@@ -8,12 +19,73 @@ export function AsyncAPISchemaParser(): SchemaParser {
...
+  return result;
+}
+
+function ajvToSpectralResult(errors: ErrorObject[], parentPath: Array<string | number>): SchemaValidateResult[] {
+  if (parentPath === undefined) {
+    parentPath = [];
+  }
+
+  return errors.map(error => {
+    const errorPath = error.instancePath.replace(/^\//, '').split('/'); // TODO: Instance Path or Schema Path?
```
magicmatatjahu, line 41: Instance path, because we should inform about problems related on inputs, not on JSON Schema.

src/schema-parser/asyncapi-schema-parser.ts
```diff
@@ -8,12 +19,73 @@ export function AsyncAPISchemaParser(): SchemaParser {
...
 
-function validate() {
+async function validate(input: ValidateSchemaInput<unknown, unknown>): Promise<SchemaValidateResult[]> {
+  const version = input.asyncapi.semver.version
+  const validator = findSchemaValidator(version);
+
+  let result: SchemaValidateResult[] = []
+  const valid = validator(input.data);
+  if (!valid && validator.errors) {
+    result = ajvToSpectralResult(validator.errors, input.path);
```
magicmatatjahu, line 29: That function will run in async mode, and `validator` saves errors to this same validator object (between execution), so better option will be to copy the errors to new array, for safety.

```suggestion
    result = ajvToSpectralResult([...validator.errors], input.path);
```

test/schema-parser/asyncapi-schema-parser.spec.ts
magicmatatjahu: I also don't, but we shouldn't focus on that, it's not our library. Probably JSOn Schema for draft-07 uses underneath some complex logic based on `allOf` etc and by this you see duplicated errors related to this same problem but created by two separate schemas.

magicmatatjahu (changes requested):
Good, but we need some additional improvements :)

src/schema-parser/asyncapi-schema-parser.ts
magicmatatjahu: Done.

src/schema-parser/asyncapi-schema-parser.ts
smoya: nice catch! fixing

src/types.ts
magicmatatjahu: It's not resolved.

magicmatatjahu (approved):

src/types.ts
smoya: Sorry, didn't commit. It is now, please check again.

magicmatatjahu (approved):

PR #576, opened 2022-08-11
refactor: add Avro Schema Parser

smoya:
**Description**

This PR does the following:

- Ports https://github.com/asyncapi/avro-schema-parser to TypeScript.
- Makes it compliant with the [new interface](https://github.com/asyncapi/parser-js/issues/480) for Schema Parsers (new `validate()` method).

> **Note**
> There are some TODOs I left in purpose to have a discussion.

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/480

PR #577, opened 2022-08-17
refactor: add RAML Schema Parser

smoya:
**Description**

This PR does the following:

- Ports https://github.com/asyncapi/raml-dt-schema-parser to TypeScript.
- Makes it compliant with the [new interface](https://github.com/asyncapi/parser-js/issues/480) for Schema Parsers (new `validate()` method).

> **Note**
> Schema validation is done through https://github.com/raml-org/webapi-parser. It does not provide a real JSON Path value for validation errors or something that can be really transformed in all cases. By now, path will point to the message payload but we can keep investigating other solutions.

**Related issue(s)**

### DIFF
--- a/src/schema-parser/raml-schema-parser.ts
+++ b/src/schema-parser/raml-schema-parser.ts
@@ -1,0 +1,69 @@
+import { SchemaParser, ParseSchemaInput, ValidateSchemaInput } from "../schema-parser";
+import type { AsyncAPISchema, SchemaValidateResult } from '../types';
+import yaml from 'js-yaml';
+import * as lib from "webapi-parser";
+const wap = lib.WebApiParser;
+const r2j = require('ramldt2jsonschema');
+
+export function RamlSchemaParser(): SchemaParser {
+    return {
+      validate,
+      parse,
+      getMimeTypes,
+    }
+  }
+
+async function parse(input: ParseSchemaInput<unknown, unknown>): Promise<AsyncAPISchema> {
+    const message = (input.meta as any).message
+    try {
+        const payload = formatPayload(input.data);
+
+        // Draft 6 is compatible with 7.
+        const jsonModel = await r2j.dt2js(payload, 'tmpType', { draft: '06' });
+        const convertedType = jsonModel.definitions.tmpType;
+
+        message['x-parser-original-schema-format'] = input.schemaFormat || input.defaultSchemaFormat;
+        message['x-parser-original-payload'] = payload;
+        message.payload = convertedType;
+        delete message.schemaFormat;
+    } catch (e) {
+        console.error(e);
+    }
+
+    return message.payload;
+}
+
+function getMimeTypes() {
+  return [
+    'application/raml+yaml;version=1.0',
+  ];
+}
+
+async function validate(input: ValidateSchemaInput<unknown, unknown>): Promise<SchemaValidateResult[]> {
+  const payload = formatPayload(input.data);
+  const parsed = await wap.raml10.parse(payload);
+  const report = await wap.raml10.validate(parsed);
+  if (report.conforms) {
+    // No errors found.
+    return [];
+  }
+  
+  let validateResult: SchemaValidateResult[] = [];
+  report.results.forEach(result => {
+    validateResult.push({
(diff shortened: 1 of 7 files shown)


src/schema-parser/raml-schema-parser.ts
```diff
@@ -0,0 +1,69 @@
...
+async function parse(input: ParseSchemaInput<unknown, unknown>): Promise<AsyncAPISchema> {
+    const message = (input.meta as any).message
+    try {
+        const payload = formatPayload(input.data);
+
+        // Draft 6 is compatible with 7.
+        const jsonModel = await r2j.dt2js(payload, 'tmpType', { draft: '06' });
+        const convertedType = jsonModel.definitions.tmpType;
+
+        message['x-parser-original-schema-format'] = input.schemaFormat || input.defaultSchemaFormat;
```
smoya, line 25: Same as https://github.com/asyncapi/parser-js/pull/576#discussion_r943559156

fmvilas (approved):

src/schema-parser/raml-schema-parser.ts
magicmatatjahu: I wrote comment about that in PR for AVRO :) We will handle that in one place.

src/schema-parser/raml-schema-parser.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (report.conforms) {
+    // No errors found.
+    return [];
+  }
+  
+  let validateResult: SchemaValidateResult[] = [];
+  report.results.forEach(result => {
+    validateResult.push({
+      message: result.message,
+      path: input.path, // Info provided by the RAML parser doesn't provide a real path to the error.
```
magicmatatjahu, line 55: In the custom parsers path of single error should point to from perspective of given payload - in other words, payload should be treated as root of error, we will concatenate error path in one place, concatenate path of schema + path of error, so here you should return empty array.

 ```suggestion
      path: [] // Info provided by the RAML parser doesn't provide a real path to the error.
```

**Review by magicmatatjahu — COMMENTED** [review]
Great. I left two comments, please refer to them and we can merge that PR :)

File: src/schema-parser/raml-schema-parser.ts (hunk 2 above)
Comment (smoya) [reply src/schema-parser/raml-schema-parser.ts:55]: Let's say `input.path` value is: `["channels", "myChannel", "publish", "message", "payload"]`, the behaviour will be different that https://github.com/asyncapi/parser-js/pull/578/files#r952869403

Comment (magicmatatjahu) [reply src/schema-parser/raml-schema-parser.ts:55]: But there you operate on error results, in mentioned comment you create mock data for testing. `.path` of input is location on which data you operate some logic, `error.path` is path of the error, but from `.path` point of view - it's a local path that we will merge after validation 😄

Comment (smoya) [reply src/schema-parser/raml-schema-parser.ts:55]: That means, this is also wrong, right? https://github.com/asyncapi/parser-js/blob/next-major/src/schema-parser/asyncapi-schema-parser.ts#L68

Comment (smoya) [reply src/schema-parser/raml-schema-parser.ts]: Changed!

**Review by magicmatatjahu — CHANGES_REQUESTED** [review]
One thing, rest is awesome! :)

File: src/schema-parser/raml-schema-parser.ts [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+        // Draft 6 is compatible with 7.
+        const jsonModel = await r2j.dt2js(payload, 'tmpType', { draft: '06' });
+        const convertedType = jsonModel.definitions.tmpType;
+
+        message['x-parser-original-schema-format'] = input.schemaFormat || input.defaultSchemaFormat;
+        message['x-parser-original-payload'] = payload;
+        message.payload = convertedType;
+        delete message.schemaFormat;
+    } catch (e) {
+        console.error(e);
```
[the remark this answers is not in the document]
smoya: Removed!

magicmatatjahu (approved):
LGTM! 🚀 Should we wait for review of parser's maintainers?

smoya:
We can merge now as we have one approval from a code owner.

PR #578, opened 2022-08-19
refactor: add OpenAPI Schema Parser

smoya:
**Description**

This PR does the following:

- Ports https://github.com/asyncapi/openapi-schema-parser to TypeScript.
- Makes it compliant with the [new interface](https://github.com/asyncapi/parser-js/issues/480) for Schema Parsers (new `validate()` method).

**Related issue(s)**


src/schema-parser/openapi-schema-parser.ts
```diff
@@ -0,0 +1,118 @@
...
+    
+    iterateSchema(transformed);
+
+    const message = (input.meta as any).message
+    if (message !== undefined) {
+      message['x-parser-original-schema-format'] = input.schemaFormat || input.defaultSchemaFormat;
+      message['x-parser-original-payload'] = message.payload;
+      message.payload = transformed;
+      delete message.schemaFormat;
+    };
```
magicmatatjahu, line 71: It will be handled in one place, but now leave it.

src/schema-parser/openapi-schema-parser.ts
```diff
@@ -0,0 +1,118 @@
...
+            result.push({
+                message: r.message,
+                path: input.path.concat(r.path.filter(p => !["components", "schemas", "schema"].includes(p.toString()))),
+            });
+        });
+    } catch(err) {
+      console.error(err);
+    } 
+
+    return result;
```
magicmatatjahu, line 47: I have problem with that 😄 I see that you use Spectral , but I don't think so that it's needed here, because we only should validate schema shape and running spectral will only slow down our validation here. We should use here AJV and validate input data based on that JSON Schema https://github.com/stoplightio/spectral/blob/develop/packages/rulesets/src/oas/schemas/3.0.json#L313

magicmatatjahu (commented):
Great, but I have problem with `validate` function. Please refer to my comment :)

src/schema-parser/openapi-schema-parser.ts
smoya: We can use spectral and only leave the rule that uses the schema? We could in fact do the same for AsyncAPI Schema?

magicmatatjahu: But then you put every rule to the final bundle. We should avoid that. For AsyncAPI case yes, I agree we can reuse spectral but not as separate Spectral instance, but this same.

smoya: Is it really making a big difference in terms of bundle size? Because in my opinion, it simplifies and unifies the logic by using always the same tool. It also returns the expected errors, etc.

test/schema-parser/openapi/openapi-schema-parser.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        patch: 0
+      }, 
+      source: "",
+      parsed: {},
+    },
+    data: message.payload,
+    meta: {
+      message: message,
+    },
+    path: ["channels", "myChannel", "publish", "message", "payload"],
```
smoya, line 64: @magicmatatjahu as per your comment in https://github.com/asyncapi/parser-js/pull/577#discussion_r952480918,  then you mean this should not be like this and instead empty array?

magicmatatjahu: No, it's a path of message's payload (input), not path of error.

src/schema-parser/openapi-schema-parser.ts
smoya: That's fair enough 👍 Good point.

In fact, I see like a good feature request for Spectral: to have the possibility to only import few rules from a ruleset instead of all? Do you think this would be possible?

magicmatatjahu: It's not possible in this way. You can simply filter rules, but that rules will be still in final bundle.

test/schema-parser/openapi/openapi-schema-parser.spec.ts
smoya: Additionally, that means, this is also wrong, right? https://github.com/asyncapi/parser-js/blob/next-major/src/schema-parser/asyncapi-schema-parser.ts#L68

src/schema-parser/openapi-schema-parser.ts
smoya: I changed the code and now it's using AJV as you suggested.

smoya:
Pinging @derberg @fmvilas as they are codeowners of https://github.com/asyncapi/openapi-schema-parser

src/schema-parser/openapi-schema-parser.ts
```diff
@@ -0,0 +1,117 @@
+import { SchemaParser, ParseSchemaInput, ValidateSchemaInput } from "../schema-parser";
+import type { AsyncAPISchema, SchemaValidateResult } from '../types';
+import Ajv, { ErrorObject, ValidateFunction } from "ajv";
+import * as fs from 'fs';
+import * as path from 'path';
+const toJsonSchema = require('@openapi-contrib/openapi-schema-to-json-schema');
+
+const schemaV3 = JSON.parse(fs.readFileSync(path.resolve(__dirname, './openapi/schema_v3.json'), 'utf8'));
```
magicmatatjahu, line 8: You should try to avoid using native nodeJS modules. Current OpenAPI and AVRO formats we can use in browser, RAML not (parser for that format uses java lib to convert raml to json-schema), so please pass content of `schema_v3.json` inside this file.

magicmatatjahu (changes requested):
One thing, rest is awesome! :)

src/schema-parser/openapi-schema-parser.ts
magicmatatjahu: Thanks!

src/schema-parser/openapi-schema-parser.ts
smoya: Fixed!

magicmatatjahu (approved):
LGTM! 🚀 

cc @derberg @fmvilas You're codeowners of that parser, so please check PR (of course if you want).

fmvilas (approved):

PR #579, opened 2022-08-24
refactor: integrate custom schemas with parser and spectral

magicmatatjahu:
**Description**

- integrate custom schemas with parser and Spectral:
  - write custom Spectral rule to validate schemas
  - write logic to parse schemas
  - write unit tests

cc @smoya 

**Related issue(s)**
Part of #482


src/schema-parser/avro-schema-parser.ts
```diff
@@ -32,29 +34,15 @@ async function validate(input: ValidateSchemaInput<unknown, unknown>): Promise<S
...
-    message['x-parser-original-payload'] = input.data;
-    message.payload = asyncAPISchema;
-    delete message.schemaFormat; 
-
-    await handleKafkaProtocolKey();
+  const message = (input.meta as any).message
+  const key = message?.bindings?.kafka?.key;
+  if (key) {
+    const bindingsTransformed = await avroToJsonSchema(key);
+    message['x-parser-original-bindings-kafka-key'] = key;
```
smoya, line 44: 💯

src/custom-operations/parse-schema.ts
```diff
@@ -62,6 +67,6 @@ export async function parseSchemasV2(parser: Parser, detailed: DetailedAsyncAPI)
 }
 
 async function parseSchemaV2(parser: Parser, item: ToParseItem) {
-  item.value[xParserOriginalSchemaFormat] = item.input.schemaFormat;
```
smoya, line 65: I see you are removing it but I can't see where we add `xParserOriginalSchemaFormat` value back.

src/parser.ts
```diff
@@ -18,16 +19,18 @@ export class Parser {
...
+    const { spectral } = this.options || {};
     if (spectral instanceof Spectral) {
       this.spectral = spectral;
     } else {
       this.spectral = new Spectral(spectral);
     }
 
-    // TODO: fix type
-    this.spectral.setRuleset(aasRuleset as any);
+    this.registerSchemaParser(AsyncAPISchemaParser());
```
smoya, line 32: You can now register the other ones as all have been merged.

magicmatatjahu: No. As current solution, people need to register given format manually. For example, RAML works only on nodejs, so register RAML here will break browser compability.

src/custom-operations/parse-schema.ts
magicmatatjahu: We don't remove schemaFormat from message in custom parsers so we don't need reset to extension. Is it clear?

src/parser.ts
smoya: Oh, that's right! 👍  thanks

src/custom-operations/parse-schema.ts
smoya: Then we need to modify the README as well

magicmatatjahu: Ok, I will change it! :)

src/schema-parser/spectral-rule-v2.ts
```diff
@@ -0,0 +1,101 @@
...
+  return {
+    description: 'Custom schema must be correctly formatted from the point of view of the used format.',
+    formats: [aas2_0, aas2_1, aas2_2, aas2_3, aas2_4],
+    message: '{{error}}',
+    severity: 'error',
+    type: 'validation',
+    recommended: true,
+    given: [
+      // operations
+      '$.channels.*.[publish,subscribe].message',
```
smoya, line 22: I just realized all those paths are tied to AsyncAPI Spec v2.x.x. How would we make it happen for v3.x.x ?

magicmatatjahu: By separate function, we can use different "formats" for rules in this same rulesets - one function for v2, another one for v3. Don't worry :)

smoya (approved):
LGTM! Great job @magicmatatjahu 💯 🚀 🌔

PR #580, opened 2022-08-25
refactor: fix AsyncAPI Schema Parser error path

smoya:
**Description**

As mentioned in https://github.com/asyncapi/parser-js/pull/577#discussion_r952480918, the returned path should be relative to the payload and should not include the path to the payload itself. 

**Related issue(s)**


magicmatatjahu (approved):
LGTM! Thanks for fast PR :)

PR #581, opened 2022-08-26
refactor: remove mixins and add AsyncAPI TS types

magicmatatjahu:
**Description**

- remove mixins - they create a lot of problems in maintaining and, in addition, do not work well with generic types in TS.
- add TS interfaces/types for v2 spec - use that interfaces in v2 models 

I know that there're a lot of changes, but only due to fact that I had to ride off mixins and add to every model created typings.

cc @smoya 

**Related issue(s)**
Part of #93 
Part of #482


smoya:
@magicmatatjahu It was hard to me to understand those interfaces represent the actual AsyncAPI Spec by each version. I want to suggest we change the dir `interfaces` to something else that helps visualizing what it is actually. Maybe `spec` or similar?

src/interfaces/v2.ts
```diff
@@ -0,0 +1,380 @@
+import type { JSONSchema7, JSONSchema7Type } from "json-schema";
+
+export type AsyncAPIVersion = '2.0.0' | '2.1.0' | '2.2.0' | '2.3.0' | '2.4.0';
```
smoya, line 3: You can play with `@asyncapi/specs` package like we do in https://github.com/asyncapi/parser-js/pull/583/files#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R19 instead of hardcoding, so if we release `2.5.0` you don't need to update this line.

magicmatatjahu: Package `@asyncapi/specs` hasn't exported packages so we cannot "retrieve" types like here https://github.com/asyncapi/parser-js/pull/583/files#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R19

TS is a something like linter, it doesn't read types from runtime.

magicmatatjahu:
@smoya 

> @magicmatatjahu It was hard to me to understand those interfaces represent the actual AsyncAPI Spec by each version. I want to suggest we change the dir interfaces to something else that helps visualizing what it is actually. Maybe spec or similar?

Sure, `spec-types` is probably better, I don't know what `spec` would be for me if I didn't know about these types 😅

src/interfaces/v2.ts
magicmatatjahu: We can also change that type to string.

smoya: Yeah, I would vote for leaving it as string until we refactor `@asyncapi/specs` to export the types from it (if we do at some point).

magicmatatjahu: @smoya Done :)

smoya (approved):
LGTM! 🚀 🌔

PR #582, opened 2022-08-30
refactor: add schemaFormat method to the Message model

magicmatatjahu:
Add `schemaFormat` method to the `Message model` and `MessageTrait model`. Atm we don't have clarification if we wanna go forward with proposal https://github.com/asyncapi/spec/pull/797 so will add `schemaFormat` to mentioned models and (maybe) in the future deprecate that method.

cc @smoya 

Part of:
#482 
#481


smoya (approved):
LGTM!

PR #584, opened 2022-08-30
docs: add node.js and npm to README.md

### LINKED ISSUE #427: [New Doc 📑]: specify which version of npm is required to run `npm install`.
### What Dev Docs changes are you proposing?

Current `package-lock.json` file sets `lockfileVersion` to `2`. This version is only available starting from `npm 7`.
In the case you run `npm install` with a version lower than 7, it will change the version of the `package-lock.json` file to `1`.

In order to keep consistency, I think we should definitely specify that this project requires npm version >= 7 and node version >= 14 (as per our CI https://github.com/asyncapi/parser-js/blob/master/.github/workflows/if-nodejs-pr-testing.yml#L33). In fact, better to say the best scenario is that the developer sets exactly the same versions as we do so we ensure everything works as expected.

Maybe we could add a `. nodeenvrc` file to set a virtual environment for npm. See https://github.com/ekalinin/nodeenv.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tinatinz:
**Description**

- Wrote in 'Develop' section that the latest version of Node.js and npm are required

**Related issue(s)**
Fixes #427


README.md
```diff
@@ -238,6 +238,7 @@ A few advantages of this solution:
 1. Write code and tests.
 1. Make sure all tests pass `npm test`
 1. Make sure code is well formatted and secure `npm run lint`
+1. Make sure you are using the latest version of both Node.js and npm.
```
magicmatatjahu, line 241: Why latest? We run workflows with v14 https://github.com/asyncapi/parser-js/blob/master/.github/workflows/if-nodejs-pr-testing.yml#L40 Could you elaborate why you need that?

magicmatatjahu (commented):
@tinatinz Thanks for PR! I leave one comment :)

README.md
derberg: @tinatinz please specify versions like in https://github.com/asyncapi/parser-js/issues/427, so not the latest Node but node>=14 and npm>=7

and I think better to make it step one, of the instruction, so contributors know it upfront

tinatinz: > Why latest? We run workflows with v14 https://github.com/asyncapi/parser-js/blob/master/.github/workflows/if-nodejs-pr-testing.yml#L40 Could you elaborate why you need that?

Sorry, I misunderstood #427 and thought the latest version was needed. I've changed the wording and placement like @derberg mentioned.

README.md
```diff
@@ -235,6 +235,7 @@ A few advantages of this solution:
 
 ## Develop
 
+1. Make sure you are using Node.js 14.0 or higher and npm 7.0 or higher
```
derberg, line 238:
```suggestion
1. Make sure you are using Node.js 14 or higher and npm 7 or higher
```

for devs it is normal that when you refer to any possible major version, you do not need to specify `.0` but just write `14`. Actually `14.0` is a bit confusing 😉

tinatinz: gotcha, fixed it.

derberg (approved):
Great, congrats on your first GitHub PR to an open source project 👏🏼 

@smoya as you created the initial issue, is there anything you'd like to add? are you happy with the solution?

smoya (commented):
LGTM! 🚀🌖

Thanks @tinatinz for your contribution!

smoya (approved):

smoya:
I noticed the right combo should be Node >=15,  NPM >= 7., according to https://nodejs.org/en/download/releases/

2022-10-18 at 15 10 35

I'm gonna open a PR fixing this.

PR #586, opened 2022-08-31
refactor: adapt components to parser-api

smoya:
**Description**

This PR updates the `Components` model with latest changes from https://github.com/asyncapi/parser-api/pull/71.
This PR does not add either `SecurityRequirement`, or `SecurityRequirements`, which will be added in a separate PR.

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/482


src/models/v2/channel.ts
```diff
@@ -21,6 +21,7 @@ import type { ServersInterface } from "../servers";
 import type { ServerInterface } from "../server";
 
 import type { v2 } from "../../spec-types";
+import { OperationObject } from "spec-types/v2";
```
magicmatatjahu, line 24: You have imported types from `v2` version. Please use `v2.OperationObject` in places where you operate on it :)

```suggestion
```

src/models/v2/channel.ts
```diff
@@ -53,8 +54,9 @@ export class Channel extends BaseModel<v2.ChannelObject, { id: string, address:
   operations(): OperationsInterface {
     const operations: OperationInterface[] = [];
     ['publish', 'subscribe'].forEach(operationAction => {
+      let id =  this._json[operationAction as 'publish' | 'subscribe'] && (this._json[operationAction as 'publish' | 'subscribe'] as OperationObject).operationId || this.meta().id + "_" + operationAction;
```
magicmatatjahu, line 57: You don't reassign that variable, so we can make it const:

```suggestion
      const id =  this._json[operationAction as 'publish' | 'subscribe'] && (this._json[operationAction as 'publish' | 'subscribe'] as OperationObject).operationId || this.meta().id + "_" + operationAction;
```

src/models/v2/channel.ts
```diff
@@ -53,8 +54,9 @@ export class Channel extends BaseModel<v2.ChannelObject, { id: string, address:
...
       this._json[operationAction as 'publish' | 'subscribe'] && operations.push(
-        this.createModel(Operation, this._json[operationAction as 'publish' | 'subscribe'], { id: operationAction, action: operationAction, pointer: `${this._meta.pointer}/${operationAction}` }),
+        this.createModel(Operation, this._json[operationAction as 'publish' | 'subscribe'], { id: id, action: operationAction, pointer: `${this._meta.pointer}/${operationAction}` }),
```
magicmatatjahu, line 59: In JS (in Go as I recall you probably have the same feature) if the key you pass to an object, exists in the "shape" of that object then you can an avoid declaring key, so:

```suggestion
        this.createModel(Operation, this._json[operationAction as 'publish' | 'subscribe'], { id, action: operationAction, pointer: `${this._meta.pointer}/${operationAction}` }),
```

is enough :)

src/models/v2/components.ts
```diff
@@ -92,11 +114,13 @@ export class Components extends BaseModel<v2.ComponentsObject> implements Compon
     return extensions(this);
   }
 
-  protected createMap<M extends BaseModel>(itemsName: keyof v2.ComponentsObject, model: Constructor<M>): Record<string, M> {
-    return Object.entries(this._json[itemsName] || {}).reduce((items, [itemName, item]) => {
-      items[itemName] = this.createModel(model, item, { id: itemName, pointer: `/components/${itemsName}/${itemName}` })
-      return items;
-    }, {} as Record<string, M>);
+  protected createCollection<M extends Collection<any>, T extends BaseModel>(itemsName: keyof ComponentsObject, collectionModel: Constructor<M>, itemModel: Constructor<T>): M {
+    let collectionItems: T[] = [];
```
magicmatatjahu, line 118:
```suggestion
    const collectionItems: T[] = [];
```

magicmatatjahu (commented):
Great! :) I found some things that we can improve, but they aren't bugs.

src/models/v2/channel.ts
smoya: Yeah, it was a `let` because previously I didn't add a logical operator in there but an if later. Good catch!

src/models/v2/channel.ts
smoya: Fixed!

magicmatatjahu (approved):
LGTM!

PR #587, opened 2022-08-31
refactor: add extensions method to the bindings

magicmatatjahu:
**Description**

- Add missed `extensions` method to the Bindings model
- Fix types for `AsyncAPISchemaObject`
- Fix types for Collection abstract model
- Add unit tests

**Related issue(s)**
Part of #481 
Part of #482


src/models/v2/message.ts
```diff
@@ -68,7 +68,7 @@ export class Message extends MessageTrait<v2.MessageObject> implements MessageIn
           (operation.message.oneOf || []).includes(this._json)
         )) {
           operations.push(
-            this.createModel(Operation, operation, { pointer: `/channels/${tilde(channelAddress)}/${operationAction}`, action: operationAction })
+            this.createModel(Operation, operation, { id: '', pointer: `/channels/${tilde(channelAddress)}/${operationAction}`, action: operationAction as 'subscribe' | 'publish' })
```
smoya, line 71: Can we add a new type for the Operation actions for v2 ('subscribe' | 'publish') so we can avoid "hardcoding" this here but instead use such type?

magicmatatjahu: Good point, I will add! :)

src/models/v2/security-requirement.ts
```diff
@@ -1,15 +1,14 @@
 import { BaseModel } from '../base';
+
 import type { SecuritySchemeInterface } from '../security-scheme';
-import type { v2 } from "../../spec-types";
-import { SecurityRequirementInterface } from 'models/security-requirement';
-import { SecurityScheme } from './security-scheme';
+import type { SecurityRequirementInterface } from '../security-requirement';
 
-export class SecurityRequirement extends BaseModel<{}, { id: string, scheme: SecuritySchemeInterface }> implements SecurityRequirementInterface {
+export class SecurityRequirement extends BaseModel<{ scopes?: string[], scheme: SecuritySchemeInterface }, { id?: string }> implements SecurityRequirementInterface {
```
smoya, line 6: Good point, so we declare the model directly here so the `_json` can contain the whole object. 👍 right?

magicmatatjahu: Yep, you can pass `scopes` and `scheme` directly to the constructor `_json`, not to the metadata.

src/schema-parser/asyncapi-schema-parser.ts
```diff
@@ -81,7 +82,7 @@ function getSchemaValidator(version: string): ValidateFunction {
  * To validate the schema of the payload we just need a small portion of official AsyncAPI spec JSON Schema, the Schema Object in particular. The definition of Schema Object must be
  * included in the returned JSON Schema.
  */
-function preparePayloadSchema(asyncapiSchema: AsyncAPISchema, version: string): AsyncAPISchema {
+function preparePayloadSchema(asyncapiSchema: JSONSchema7, version: string): JSONSchema7 {
```
smoya, line 85: But AsyncAPISchema !== JSONSchema7, isnt? We extend from it, adding few extra fields like discriminator, etc.

src/models/v2/security-requirement.ts
smoya: Nice, good improvement 👍

src/schema-parser/asyncapi-schema-parser.ts
magicmatatjahu: Right, I need to change it to the `AsyncAPISchemaDefinition` type. Thanks!

magicmatatjahu: Done.

src/models/v2/message.ts
magicmatatjahu: Done.

magicmatatjahu:
@smoya Ok, I applied all suggestions. Could you check again?

package-lock.json
```diff
@@ -1,15708 +1,8 @@
 {
   "name": "@asyncapi/parser",
   "version": "2.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
smoya, line 4: You are using an old version of node/npm. Please restore this file to previous version.

magicmatatjahu: Done.

src/models/base.ts
```diff
@@ -1,32 +1,31 @@
...
-    protected readonly _meta: ModelMetadata & M = {} as any,
+    protected readonly _meta: ModelMetadata & M = {} as ModelMetadata & M,
   ) {}
 
-  json<T extends Record<string, any> = J>(): T;
+  json<T = J>(): T;
   json<K extends keyof J>(key: K): J[K];
   json(key?: keyof J) {
-    if (key === undefined) return this._json;
+    if (key === undefined || typeof this._json !== 'object') return this._json;
```
smoya, line 18: I don't get why we need this change. what type could be other than object? If the user provides a key, it is expected to only return a value.

magicmatatjahu: The value can also be different from the object - we have one case in our spec, schema can have a boolean, true or false value.

smoya: But what if the user provides a value for `key`? You should still assume the user wants the value for such a key, no matters the type of `this._json`.

magicmatatjahu: Right, I will fix it.

magicmatatjahu: Done, new function body:

```ts
    if (key === undefined) return this._json;
    if (this._json === null || this._json === undefined) return this._json;
    return this._json[key];
```

src/models/collection.ts
```diff
@@ -1,14 +1,28 @@
...
+    protected readonly _meta: CollectionMetadata<T> & M = {} as CollectionMetadata<T> & M,
   ) {
     super(...collections);
   }
 
-  abstract get(id: string): T | undefined;
-  abstract has(id: string): boolean;
+  get(id: string): T | undefined {
+    const possibleItem = this._meta.originalData?.[id];
+    return typeof possibleItem === 'undefined' ? this.__get(id) : possibleItem;
```
smoya, line 20: Shouldn't this be just the opposite? First we see if `__get` is declared, if so, means there is a desire from the subclass to override the whole `get`?

src/models/extension.ts
```diff
@@ -1,7 +1,7 @@
 import type { BaseModel } from "./base";
 
-export interface ExtensionInterface extends BaseModel {
+export interface ExtensionInterface<T = any> extends BaseModel {
```
smoya, line 3: Would `T` be ever any other type rather than `any` ?

magicmatatjahu: Extension can include any value, primitive, object, array.

src/models/collection.ts
magicmatatjahu: Good point, but have in mind that `__get` here is treated as fallback to retrieve data when `originalData` (which can be omitted) doesn't exist or given key also doesn't exist. With your solution we need to duplicate `const possibleItem = this._meta.originalData?.[id];` logic in every `__get` function.

smoya: What I suggest is that we do the opposite; something like:

```js
const possibleItem = this.__get(id);
return typeof possibleItem === 'undefined' ? this._meta.originalData?.[id] : possibleItem;
```

magicmatatjahu: But by this you will decrease optimization, I made looking for object first to improve read speed. Also `originalData` is optional so we can pass objects to the collections constructor only when we will have 100% sure that keys in objects === ids in collections.

src/models/extension.ts
magicmatatjahu: From slack:

You can have any value in extension, we cannot say what exactly type is in extension, but user can cast extension value to given type, so he/she can do this:
```
const extension = model.extensions().get<TwitterExtensionsShape>('twitter');
```

and then:
```
extensions.value()
```
will return

```
TwitterExtensionsShape
```

type.

src/models/collection.ts
smoya: Correct me if I'm wrong, please. Imagine having the following document:

```yaml
components:
  messages:
    test1:
      messageId: 'test'
    test2:
      messageId: 'test1'
```

What '.components().messages().get('test1')` should return? 
I expect `messageId` should have preference over the key, so I would say the returned message would be:

```yaml
test2:
      messageId: 'test1'
```

However, with your change, that won't happen since `this._meta.originalData?.[id]` will match with the other message.

src/models/extension.ts
smoya: Yeah. Makes totally sense. Thanks for clarifying!

src/models/collection.ts
magicmatatjahu: Nope, as I wrote:

```
Also `originalData` is optional so we can pass objects to the collections constructor only when we will have 100% sure that keys in objects === ids in collections.
```

so in messages/operations cases where we use `operationId` and `messageId`, we will always run `__get`.

smoya: Being hones, I'm not very convinced about the complexity of that code. Regarding the fact `originalData` is optional, it is like a hidden behavior that it is not very clear to the developer what happens when it is provided. Not even convinced it would be fixed by documenting.
Is it worth the "optimization"?

magicmatatjahu: We wrote that there is a problem with optimization the reading of data as an array, and it would be nice to have a way to speed this up using an object and keys. As for the complexity of this, it is only for internal use. The end user should not create standalone models at all that are not AsyncAPIDocument model.

magicmatatjahu: I can remove that if you want, because atm we only use it in the `extensions` case for bindings. WDYT?

smoya: I'm concerned about the Developer (of the parser) Experience, not the final user. 
I understand the benefit; kind of o(1) VS o(n) lookups, However I'm always in favor of not introducing early optimizations in code, and this one might be one, especially if not executed thinking on DX.

As a last effort to keep it, maybe at least renaming the `originalData` var to `indexedCollection` or similar and adding a comment on what this is about could help. 

Just to clarify, this could also be used for example for `servers` and `channels` (v2) right?

magicmatatjahu: I opt for `originalData` because in bindings model we need to pass original data (not changed), because then we need to read from original `bindings` field all extensions for `extensions()` method.  `indexedCollection` indicates that this data is transformed. 

> Just to clarify, this could also be used for example for servers and channels (v2) right?

Yeah, also for bindings and extensions models - in every collections where we change object to array 1:1 - we change key to the id of element.

I will remove __get, but we need `originalData` meta field.

smoya: `indexedCollection` doesn't indicate that data is transformed but just indexed by something (we can assume by id).

magicmatatjahu: Yeah, but we can even transform to the object original data - for example messages we can transform to the object, where keys will be `messageId`s.

Ok, I removed the `__get` function. We will think about optimization after release.

smoya (approved):
LGTM! 🚀

PR #591, opened 2022-09-05
refactor: handle circular references in schemas

magicmatatjahu:
**Description**

Handle circular references, which are not handled by Spectral internal ref-resolver. The solution in PR doesn't support the really strange cases related with circular references, but for the moment it is sufficient. When we release the first alpha/beta of the new API, then I will try to support all the cases with custom ref-resolver.

Also, that PR move some utils from `./utils` file to the new `./document` to prevent circular references between TS files - sometimes we need utils from `./utils` file in model file, and due to fact that we have reference to the AsyncAPIDocument model, we can case with circular ref between TS files - example with `Schema` model and `MessageTrait`.

cc @smoya 

**Related issue(s)**
Part of #481 
Part of #482


src/utils.ts
```diff
@@ -124,4 +90,38 @@ export function untilde(str: string) {
...
+
+export function retrievePossibleRef(data: any & { $ref?: string }, pathOfData: string, spec: any = {}): any {
+  if (!hasRef(data)) {
+    return data;
+  }
+
+  const refPath = serializePath(data.$ref);
+  if (pathOfData.startsWith(refPath)) { // starts by given path
+    return retrieveDeepData(spec, splitPath(refPath)) || data;
+  } else if (pathOfData.includes(refPath)) { // circular path in substring of path
```
smoya, line 103: Correct me if I'm wrong but given:

`pathOfData` = `/components/channels/test`
`refPath` = `/channels/test`

Would it  match `(pathOfData.includes(refPath)` when it shouldn't?.

magicmatatjahu: The `string.includes` in JS can be used to check if a string has a substring.

smoya: I know. What I want to raise here is that `'/components/channels/test'.includes('/channels/test')` returns `true` **but** I think it is not an expected behavior here since both are completely different paths.

smoya (approved):
LGTM!

I think we should create a new issue for keep working on a final solution regarding handling circular references; could be doing work on Spectral side, json-schema-ref-parser or whatever.

magicmatatjahu:
@smoya I created issue https://github.com/asyncapi/parser-js/issues/599

PR #592, opened 2022-09-05
refactor: add filters to some collections

smoya:
**Description**

This PR adds:
- `FilterBySend()` method to `Messages`, `Servers`, `Operations`, `Channels` models.
- `FilterByReceive()` method to `Messages`, `Servers`, `Operations`, `Channels` models.

All changes are done based on the Parser-API; last development in particular can be found in https://github.com/asyncapi/parser-api/pull/71.

**Related issue(s)**

### DIFF
--- a/test/models/v2/servers.spec.ts
+++ b/test/models/v2/servers.spec.ts
@@ -6,9 +6,31 @@ const doc = {
     protocol: 'mqtt',
     protocolVersion: '1.0.0',
     url: 'development.gigantic-server.com'
+  },
+  'development2': {
+    protocol: 'mqtt',
+    protocolVersion: '1.0.0',
+    url: 'development2.gigantic-server.com'
+  },
+  'production': {
+    protocol: 'mqtt',
+    protocolVersion: '1.0.0',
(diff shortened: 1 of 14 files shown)


test/models/v2/servers.spec.ts
```diff
@@ -46,4 +68,28 @@ describe('Servers model', function () {
       expect(servers.has('production')).toEqual(false);
     })
   })
+
+  describe('.filterBySend()', function () {
+    it('should return all servers with channels with subscribe operation', function () {
+      const servers = new Servers(serverItems);
+      expect(servers.filterBySend()).toEqual([serverItems[2], serverItems[3]]);
```
smoya, line 75: This test is failing due to some kind of sourcery comparing the message JSON value with the actual message inside the AsyncAPI parsed document. See https://github.com/asyncapi/parser-js/blob/next-major/src/models/v2/message.ts#L67

It works if comparison is done with `JSON.stringify()`, which obviously is not gonna happen since that method preserves order of keys ending in a not a fear comparison.

magicmatatjahu (approved):
LGTM! :)

PR #593, opened 2022-09-05
refactor: add possibility to pass custom resolvers (unstable)

magicmatatjahu:
**Description**

Add possibility to pass custom resolvers. With `json-schema-ref-resolvers` we have possibility to pass custom ref resolvers. Atm we use that feature in Generator and in ServerAPI. Unfortunately, ref-resolver from Spectral doesn't have such a thing, so that PR "wraps" default Spectral resolvers with adding similar functionality. Of course that feature will be used only by us, so I annotated that feature by `__unstable` prefix. Spectral team is preparing to switch to the `json-schema-ref-parser` and we will provide this functionality when Spectral will support it natively.

cc @smoya 

**Related issue(s)**
Part of #481 
Part of #482

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
         "@types/json-schema": "^7.0.11",
         "@types/lodash": "^4.14.179",
         "@types/node-fetch": "^2.6.2",
+        "@types/urijs": "^1.19.19",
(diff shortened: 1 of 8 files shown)


smoya (approved):
LGTM! 🚀🌔

PR #594, opened 2022-09-06
refactor: use @swc/jest to speedup test execution

magicmatatjahu:
**Description**

As in title. Differences:
- without `@swc`: about 30s.
- with `@swc`: about 8s.

Of course in my machine. I also removed `json` and `lcov` coverage reports - atm we don't need that. We will enable them when we will need them.

cc @smoya 

**Related issue(s)**
Part of #481
Part of #482


smoya (approved):
⚡ ⚡

PR #597, opened 2022-09-07
refactor: cleanup interfaces + add missing methods from parser-api

smoya:
**Description**

This PR does a cleanup and sync with the latest changes on `parser-api` made in https://github.com/asyncapi/parser-api/pull/71#issuecomment-1239379887.

**Related issue(s)**
- https://github.com/asyncapi/parser-js/issues/481

### DIFF
--- a/src/models/security-scheme.ts
+++ b/src/models/security-scheme.ts
@@ -2,9 +2,6 @@ import type { BaseModel } from './base';
 import type { OAuthFlowsInterface } from './oauth-flows';
 import type { DescriptionMixinInterface, ExtensionsMixinInterface } from './mixins';
 
-export type SecuritySchemaType = 'userPassword' | 'apiKey' | 'X509' | 'symmetricEncryption' | 'asymmetricEncryption' | 'httpApiKey' | 'http' | 'oauth2' | 'openIdConnect' | 'plain' | 'scramSha256' | 'scramSha512' | 'gssapi';
-export type SecuritySchemaLocation = 'user' | 'password' | 'query' | 'header' | 'header' | 'cookie';
-
 export interface SecuritySchemeInterface extends BaseModel, DescriptionMixinInterface, ExtensionsMixinInterface {
   id(): string
   hasBearerFormat(): boolean;
@@ -12,6 +9,6 @@ export interface SecuritySchemeInterface extends BaseModel, DescriptionMixinInte
   openIdConnectUrl(): string | undefined;
   scheme(): string | undefined;
   flows(): OAuthFlowsInterface | undefined;
-  type(): SecuritySchemaType;
-  in(): SecuritySchemaLocation | undefined;
+  type(): string;
+  in(): string | undefined;
 }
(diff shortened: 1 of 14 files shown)


src/models/security-scheme.ts
```diff
@@ -3,6 +3,7 @@ import type { OAuthFlowsInterface } from './oauth-flows';
 import type { DescriptionMixinInterface, ExtensionsMixinInterface } from './mixins';
 
 export type SecuritySchemaType = 'userPassword' | 'apiKey' | 'X509' | 'symmetricEncryption' | 'asymmetricEncryption' | 'httpApiKey' | 'http' | 'oauth2' | 'openIdConnect' | 'plain' | 'scramSha256' | 'scramSha512' | 'gssapi';
+export type SecuritySchemaLocation = 'user' | 'password' | 'query' | 'header' | 'header' | 'cookie';
```
magicmatatjahu, line 6: Hmm 🤔 Probably we should move that types to the `spec-types/v2`. `SecuritySchemaType` probably I wrote but we have missed `SecuritySchemaLocation` here. Could you move them?

src/models/v2/extensions.ts
```diff
@@ -6,6 +6,6 @@ import type { ExtensionInterface } from '../extension';
 export class Extensions extends Collection<ExtensionInterface> implements ExtensionsInterface {
   override get<T = any>(name: string): ExtensionInterface<T> | undefined {
     name = name.startsWith('x-') ? name : `x-${name}`;
```
magicmatatjahu, line 8:
```suggestion
  override get<T = any>(id: string): ExtensionInterface<T> | undefined {
    id = id.startsWith('x-') ? id : `x-${id}`;
```

src/models/v2/extensions.ts
```diff
@@ -6,6 +6,6 @@ import type { ExtensionInterface } from '../extension';
...
-    return this.collections.find(ext => ext.name() === name);
+    return this.collections.find(ext => ext.id() === name);
```
magicmatatjahu, line 9:
```suggestion
    return this.collections.find(ext => ext.id() === id);
```

magicmatatjahu (changes requested):

src/models/security-scheme.ts
smoya: Those are already in `spec-types/v2`. The question is, should we also add **all** of them (meaning all locations possible, no matters the version of the spec) in the interface?That was the idea behind this in `parser-api`: https://github.com/asyncapi/parser-api/pull/71/files#diff-ac35ac6e0b0ad3629802a0e04076aa18e94da614f5392e1164dcaacdfe3ee198R375.
WDYT?

magicmatatjahu: Yeah, we don't make different spec types for `2.0.0`, `2.1.0` etc. v2 spec types is for 2.x.x, so please use types from `spec-types/v2`.

smoya: > v2 spec types is for 2.x.x, so please use types from spec-types/v2

But we are talking about types located in the interface (`src/models/security-scheme.ts`). Not related with the implementation.

smoya: Do you mean we should just use `string` as type here, and in the implementation (v2, v3...) to set those values as unique for that string?

magicmatatjahu: We can even use `v2.SecuritySchemaLocation | v3.SecuritySchemaLocation` type in the future, but probably better will be `string` and then in implementation cast type to the `v2.SecuritySchemaLocation`. Please change to the `string` and move types to the `spec-types/v2`

smoya: Done!

magicmatatjahu (approved):
LGTM!

PR #604, opened 2022-09-09
refactor: add iterator for traversing an AsyncAPI doc

smoya:
**Description**
This PR adds an iterator for traversing AsyncAPI documents. It is based in the current one. See https://github.com/asyncapi/parser-js/blob/master/lib/iterators.js

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/481

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -58,7 +58,6 @@ rules:
   no-empty-character-class: 2
   no-self-compare: 2
   valid-typeof: 2
-  no-unused-vars: [2, { "args": "none" }]
   handle-callback-err: 2
   no-shadow-restricted-names: 2
   no-new-require: 2
--- a/src/iterator.ts
+++ b/src/iterator.ts
@@ -1,0 +1,249 @@
+import { AsyncAPIDocumentInterface } from 'models';
+import { ChannelInterface } from 'models/channel';
+import { ChannelParameterInterface } from 'models/channel-parameter';
+import { MessageInterface } from 'models/message';
+import { MessageTraitInterface } from 'models/message-trait';
+import { SchemaInterface } from 'models/schema';
+ 
+/**
+ * The different kind of stages when crawling a schema.  
+ */
+export enum SchemaIteratorCallbackType {
+    NEW_SCHEMA = 'NEW_SCHEMA', // The crawler just started crawling a schema.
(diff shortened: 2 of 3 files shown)


.eslintrc
```diff
@@ -58,7 +58,6 @@ rules:
   no-empty-character-class: 2
   no-self-compare: 2
   valid-typeof: 2
-  no-unused-vars: [2, { "args": "none" }]
```
smoya, line 61: Not compatible with Typescript, however we also use the compatible version so I'm just removing the support here.

src/iterator.ts
```diff
@@ -0,0 +1,249 @@
...
+    Definitions = 'definitions', // Crawl all schemas in definitions field
+}
+   
+export type TraverseOptions = {
+   callback: Function
+   schemaTypesToIterate: SchemaTypesToIterate[]
+   seenSchemas: Set<any>
+}   
+  
+export type TraverseCallback = (schema: SchemaInterface, propOrIndex: String | Number | null, callbackType: SchemaIteratorCallbackType) => void
```
magicmatatjahu, line 46:
```suggestion
export type TraverseCallback = (schema: SchemaInterface, propOrIndex: string | number | null, callbackType: SchemaIteratorCallbackType) => void
```

src/iterator.ts
```diff
@@ -0,0 +1,249 @@
...
+      traverseMessageTrait(messageTrait, options);
+    });
+  }
+}
+
+/* eslint-disable sonarjs/cognitive-complexity */
+/**
+ * Traverse current schema and all nested schemas.
+ */
+function traverseSchema(schema: SchemaInterface, propOrIndex: String | Number | null, options: TraverseOptions) { // NOSONAR
```
magicmatatjahu, line 85:
```suggestion
function traverseSchema(schema: SchemaInterface, propOrIndex: string | number | null, options: TraverseOptions) { // NOSONAR
```

magicmatatjahu (commented):
Only two comments :)

smoya:
@magicmatatjahu would you mind revieweing again? I fixed the linting issues but also improved the tests. Thanks!

magicmatatjahu (approved):

PR #607, opened 2022-09-13
refactor: add migration function from v2 to v1 API

magicmatatjahu:
**Description**

- add new iterator for v1 API - due to fact that old AsyncAPIDocument exposes traverse function in one of the method we need to create another traverse function for schemas written in v1 API.
- create `migration` function from v1 to v2 API and expose it .
- add anonymous naming for schemas and messages.
- add unit tests
- update Readme.md.
- return `document` not `parsed` from `parser.parse()` function.

cc @smoya 

**Related issue(s)**
Part of #481 
Part of #482


src/old-api/external-docs.ts
```diff
@@ -14,4 +15,36 @@ export class ExternalDocs extends SpecificationExtensionsModel<v2.ExternalDocume
   description() {
     return description(this);
   }
+
+  hasExtensions() {
```
magicmatatjahu, line 19: I had to do that due to circular references between `tag`, `externalDocs` and `mixins` files.

src/old-api/tag.ts
```diff
@@ -22,4 +23,36 @@ export class Tag extends SpecificationExtensionsModel<v2.TagObject> {
   hasExternalDocs() {
     return hasExternalDocs(this);
   }
+
+  hasExtensions() {
```
magicmatatjahu, line 27: I had to do that due to circular references between tag, externalDocs and mixins files.

README.md
```diff
@@ -300,11 +324,27 @@ For that, the Parser supports the ability to stringify a parsed AsyncAPI documen
...
 A few advantages of this solution:
+
 - The string remains as small as possible due to the use of [JSON Pointers](https://datatracker.ietf.org/doc/html/rfc6901).
 - All references (also circular) are preserved.
 
 Check [example](#example-with-stringify-and-unstringify-parsed-documentstringify).
 
+## Convert to the old API
+
+Version `2.0.0` of package introduced a lot of breaking changes, including changing the API of the returned parsed document (parser uses [Intent API](https://github.com/asyncapi/parser-api)). Due to the fact that a large part of the AsyncAPI tooling ecosystem uses a Parser with the old API and rewriting the tool for the new one can be time-consuming and difficult, the package exposes the `migrateToOldAPI()` function to convert new API to old one:
```
smoya, line 335: I think we should call this simply `Parser API`
```suggestion
Version `2.0.0` of package introduced a lot of breaking changes, including changing the API of the returned parsed document (parser uses [Parser API](https://github.com/asyncapi/parser-api)). Due to the fact that a large part of the AsyncAPI tooling ecosystem uses a Parser with the old API and rewriting the tool for the new one can be time-consuming and difficult, the package exposes the `migrateToOldAPI()` function to convert new API to old one:
```

magicmatatjahu: so old one how we will call?

smoya: You are calling it "OldAPI" within your code 😅 .

README.md
```diff
@@ -300,11 +324,27 @@ For that, the Parser supports the ability to stringify a parsed AsyncAPI documen
...
+
+```js
+import { Parser, migrateToOldAPI } from '@asyncapi/parser';
+
+const parser = new Parser();
+const { document } = parser.parse(...);
+const oldAsyncAPIDocument = migrateToOldAPI(document);
```
smoya, line 342: I do have a question: Would it make sense to, instead of having an isolated function here that requires an already parsed document, just create another parse function? E.g. `parser.parseWithOldAPI(...)`. It feels more user-friendly to me. Are there any technical cons?

magicmatatjahu: Maybe it is more user friendly, but this method will add all the old API to the final bundle for browser applications. With a separate method we can import the old api whenever we need it and not when it is optional (we will broke treeshaking support)

smoya: Fair enough and well explained 👍 👍

src/utils.ts
```diff
@@ -42,6 +43,12 @@ export function hasWarningDiagnostic(diagnostics: ISpectralDiagnostic[]): boolea
   return diagnostics.some(diagnostic => diagnostic.severity === DiagnosticSeverity.Warning);
 }
 
+export function setExtension(id: string, value: any, model: BaseModel): void {
+  id = id.startsWith('x-') ? id : `x-${id}`;
```
smoya, line 47: We are adding a technical limitation here: extensions with the shape `x-x-whatever` are being transformed to `x-whatever`, i.e. `x-x-wing`. I know it might seem a remote edge case, but it is still valid. What if we just assume the user should provide the full extension name (including the `x-` prefix`) and we validate it is happening? Still it might happen the user sets `x-wing` expecting `x-x-wing` as extension, but at least there is no limitation.

test/custom-operations/anonymous-naming.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+            operationId: 'operation',
+            message: {
+              payload: {}
+            }
+          }
+        }
+      }
+    });
+
+    expect(document?.messages()).toHaveLength(1);
```
smoya, line 26: Isn't like this? I'm confused now 😅 
```suggestion
    expect(document?.messages().all()).toHaveLength(1);
```

magicmatatjahu: Collection class extends Array, so `messages()` returns exactly array of data (with additional methods).

src/utils.ts
magicmatatjahu: Yes 😄

src/old-api/iterator.ts
```diff
@@ -0,0 +1,272 @@
+import type { AsyncAPIDocument } from './asyncapi';
```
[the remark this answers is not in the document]
magicmatatjahu: Done 👌🏼

README.md
magicmatatjahu: Done 👌🏼

smoya (approved):
LGTM! 🚀🌔 Great addition!

PR #611, opened 2022-09-14
refactor: add migration guide

magicmatatjahu:
**Description**

Add migration guide `v1` -> `v2`.

Some sections will have to be updated before release.

cc @smoya @jonaslagoni @derberg @fmvilas 

**Related issue(s)**
Part of #481 
Part of #482


docs/migration-guide-v2.md
```diff
@@ -0,0 +1,214 @@
+# Parser v1 to v2 Migration Guide
+
+## General
+
```
jonaslagoni, line 4: I am missing a bit of context why I am reading a list here.

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,214 @@
+# Parser v1 to v2 Migration Guide
```
jonaslagoni, line 1: Maybe it makes sense to start of with "why v2" 🤔

jonaslagoni (changes requested):
I would paste the migration guide into Grammarly to fix small grammatically errors, there is quite a few and don't want to point them out 😆

docs/migration-guide-v2.md
magicmatatjahu: Maybe we will change that header to the `tl;dr`? 😆 Because it's for that.

jonaslagoni: Would make sense yes 😄

docs/migration-guide-v2.md
magicmatatjahu: Overall, the first point of the General indicates why v2 - because a lot of breaking changes. Do we really need some description for that?

jonaslagoni: I guess not no 😄

jonaslagoni:
> Some sections will have to be updated before release.

If we should not review and merge it yet, do you mind converting it to a draft? 🙂

smoya (approved):
LGTM! 🚀🌔

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,276 @@
...
+              exampleDate:
+                type: string
+                format: date-time
+`);
+
+if (document) {
+  // stringify function returns string type
+  const stringifiedDocument = stringify(document);
+  // unstringify function returns new AsyncAPIDocument instance
+  const unstringifiedDocument = unstringify(stringifiedDocument);
```
fmvilas, line 259: Wait 🤔 Isn't `unstringify` just the `.parse()` method? Why is it provided as a separate function?

And why isn't `.stringify()` a method of the parsed document instead? I mean, why not `document.stringify()`?

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,276 @@
...
+- Validation is powered by [Spectral](https://github.com/stoplightio/spectral). This allows better validation, checking not only syntax errors, but also informing about good practices, such as defining `operationID` field etc. In addition, Spectral uses a different references resolver, which causes changes in defining custom resolvers.
+- AsyncAPI document validation (without parsing) is available via a separate method - `.validate()`.
+- The function `.parse()` returns parsed AsyncAPi `document` and possible `diagnostics` based on validation configuration.  
+- The function `parseFromUrl()` has been deprecated in favour of `fromURL` standalone function.
+- The Custom Schema Parser interface has been changed with an additional function `validate()`.
+- Stringify and unstringify the parsed document is now a separate functionality.
+- Stringify and unstringify the parsed document is now a separate functionality.
+- Package is fully [tree-shakable](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking).
```
fmvilas, line 13:
```suggestion
```

This is already described in the Typescript and ESM section.

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+### New Intent API
+
+The old Document API based 1:1 on AsyncAPI Specification has been rewritten to support the new so-called [Intent API](https://github.com/asyncapi/parser-api), which makes it easier to operate on the document and retrieve collections of needed objects from the document. Due to the fact that the new API does not stick firmly to AsyncAPI Specification, it allows for better integration with future versions of AsyncAPI. User does not need to know the new API against the new version of AsyncAPI - all methods will remain the same with the inclusion of new/deprecated fields. Also new API has been enhanced with some additional functionality like retrieving the JSON Path of the object using the `.jsonPath()` method etc.
+
+### EcmaScript Modules
```
fmvilas, line 19:
```suggestion
### TypeScript and ESM
```

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+The source code of the package was rewritten from CJS (CommonJS) to ESM (EcmaScript) modules. This allows the fully [tree-shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking) and much better integration in browser-type environments like SPA applications. In addition, the package is published with:
```
fmvilas, line 21:
```suggestion
The source code of the package has been rewritten to ESM (EcmaScript modules). This enables [tree-shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking) and much better integration in browser-based environments. In addition, the package exposes a `cjs` directory containing the CommonJS version of the parser. For compatibility with older environments.
```

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+- `esm` (ESM modules) folder mainly for browser applications and newer versions of NodeJS, to support fully [tree-shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking) 
+- `cjs` (CJS modules) folder for old versions of NodeJS
```
fmvilas, line 24: You don't need that now :)

```suggestion
```

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,276 @@
...
+- Package source code has been rewritten to the TypeScript alongside with rewritten from CJS (CommonJS) to ESM (EcmaScript) modules.
+- The functionality of the Parser was wrapped with the class. This means that from now on it's possible to have a dozen other Parser instances with different configurations.
+- Validation is powered by [Spectral](https://github.com/stoplightio/spectral). This allows better validation, checking not only syntax errors, but also informing about good practices, such as defining `operationID` field etc. In addition, Spectral uses a different references resolver, which causes changes in defining custom resolvers.
+- AsyncAPI document validation (without parsing) is available via a separate method - `.validate()`.
+- The function `.parse()` returns parsed AsyncAPi `document` and possible `diagnostics` based on validation configuration.  
+- The function `parseFromUrl()` has been deprecated in favour of `fromURL` standalone function.
+- The Custom Schema Parser interface has been changed with an additional function `validate()`.
+- Stringify and unstringify the parsed document is now a separate functionality.
```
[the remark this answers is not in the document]
magicmatatjahu: Stringify and unstringify functionalities aren't related to the ref resolvers.

docs/migration-guide-v2.md
magicmatatjahu: > Wait 🤔 Isn't unstringify just the .parse() method? Why is it provided as a separate function?

If you put stringified document to the `.parse()` method, you will fire `unstringify()` function, don't worry, but then you need bundle all code related to the Parser, even when you don't need that, e.g. our html-template - `unstringify` function is for such a situation, to reduce bundled size of app.

> And why isn't .stringify() a method of the parsed document instead? I mean, why not document.stringify()?

To treeshake not needed functionality as much as possibility. If you will add `stringify` method to the AsyncAPIDocument class then this code will be always bundled in final code (even if you won't use that method).

docs/migration-guide-v2.md
magicmatatjahu: Just because some lib is written with ESM does not mean it is fully tree-shakable :) it is better to list it here so that people do not have to look in other subsections.

magicmatatjahu: Ok, I changed second point in the list to the:

> Source code has been rewritten in TypeScript and ESM. This makes the package fully [tree-shakable](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking). [Read more](#typescript-and-esm).

What do you think?

docs/migration-guide-v2.md
magicmatatjahu: Aaa ok, I read another comment.

jonaslagoni (approved):

docs/migration-guide-v2.md
fmvilas: Alright 👍 That makes sense. Thanks for the explanation!

docs/migration-guide-v2.md
fmvilas: Yeah, that's much better. Thanks!

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,276 @@
+# Parser v1 to v2 Migration Guide
+
+- The document's AsyncAPI API has been rewritten from old to new supporting the [Intent API](https://github.com/asyncapi/parser-api) - there is a function that converts the new API to the old one for backward compatibility.
```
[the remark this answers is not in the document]
fmvilas: What about this suggestion, @magicmatatjahu? I think it improves the readability of the first point.

docs/migration-guide-v2.md
```diff
@@ -0,0 +1,284 @@
...
+If there is a need to convert an existing instance of the old API to the new one, the library provides the `convertToNewAPI` function:
+
+```js
+import { Parser, convertToNewAPI } from '@asyncapi/parser';
+
+const newAsyncAPIDocument = convertToNewAPI(oldDocument);
+```
+
+> **Warning**
+> Due to the fact that the new Parser validates the docuemnt more restrictively, it may be that a parsed document in the old way will not work 100% in accordance with the new API.
```
fmvilas, line 284:
```suggestion
> Due to the fact that the new Parser validates the document more restrictively, it may be that a parsed document in the old way will not work 100% in accordance with the new API.
```

fmvilas (changes requested):
Left some comments

docs/migration-guide-v2.md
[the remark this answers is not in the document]
magicmatatjahu: I forgot about this 😄 Applied.

fmvilas (dismissed):
Just change the title of the PR to `docs:`. I think it's more appropriate.

smoya:
@magicmatatjahu you need to change the base branch to be `master` now since the Parser has been merged.

smoya:
@magicmatatjahu After changing base branch to `master`, now it has conflicts. I think the best is to just reset --hard to the HEAD of `master` and do cherry-pick of the commit(s) that adds the migration guide.

jonaslagoni:
This really should have been resolved before we release v2 of the parser.

jonaslagoni:
Re-creating

PR #612, opened 2022-09-15
refactor: implement schemas and all* methods in AsyncAPIDocument class

magicmatatjahu:
**Description**

Implement `schemas`, `allChannels`, `allServers`, `allOperations`, `allMessages`, `allSchemas` methods in the AsyncAPIDocument class.

cc @smoya 

**Related issue(s)**
Part of #481 
Part of #482

### DIFF
--- a/src/models/v2/asyncapi.ts
+++ b/src/models/v2/asyncapi.ts
@@ -18,7 +18,9 @@ import { tilde } from '../../utils';
 import type { AsyncAPIDocumentInterface } from '../asyncapi';
 import type { InfoInterface } from '../info';
 import type { ServersInterface } from '../servers';
+import type { ServerInterface } from '../server';
 import type { ChannelsInterface } from '../channels';
+import type { ChannelInterface } from '../channel';
 import type { ComponentsInterface } from '../components';
 import type { OperationsInterface } from '../operations';
 import type { OperationInterface } from '../operation';
@@ -66,27 +68,20 @@ export class AsyncAPIDocument extends BaseModel<v2.AsyncAPIObject> implements As
 
   operations(): OperationsInterface {
     const operations: OperationInterface[] = [];
-    this.channels().forEach(channel => operations.push(...channel.operations().all()));
+    this.channels().forEach(channel => operations.push(...channel.operations()));
     return new Operations(operations);
   }
 
   messages(): MessagesInterface {
     const messages: MessageInterface[] = [];
-    this.operations().forEach(operation => messages.push(...operation.messages().all()));
+    this.operations().forEach(operation => operation.messages().forEach(message => (
+      !messages.some(m => m.json() === message.json()) && messages.push(message)
+    )));
     return new Messages(messages);
   }
 
   schemas(): SchemasInterface {
-    const schemas: Set<SchemaInterface> = new Set();
-    function callback(schema: SchemaInterface) {
-      if (!schemas.has(schema.json())) {
-        schemas.add(schema);
-      }
-    }
-    // return only schemas used in channels ("active" schemas)
-    const toIterate = Object.values(SchemaTypesToIterate).filter(s => s !== SchemaTypesToIterate.Components);
-    traverseAsyncApiDocument(this, callback, toIterate);
-    return new Schemas(Array.from(schemas));
+    return this.__schemas(false);
   }
 
   securitySchemes(): SecuritySchemesInterface {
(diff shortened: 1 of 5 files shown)


src/models/v2/asyncapi.ts
```diff
@@ -76,7 +77,16 @@ export class AsyncAPIDocument extends BaseModel<v2.AsyncAPIObject> implements As
...
 
   schemas(): SchemasInterface {
-    return new Schemas([]);
+    const schemas: Set<SchemaInterface> = new Set();
+    function callback(schema: SchemaInterface) {
+      if (!schemas.has(schema.json())) {
+        schemas.add(schema);
+      }
+    }
+    // return only schemas used in channels ("active" schemas)
```
smoya, line 86: This is raising a hidden behaviour I didn't think when designing the API. Are we sure this is the behavior all users will need? Once i have the "active" schemas, how do I know the ones are not active?

I think we should rather return **all schemas** and add a filter in the `Schemas` collection such as `FilterByActive()` and `FilterByInactive` or `FilterByInUse()`/`FilterByNotInUse()` or whatever.

magicmatatjahu: If someone will need also inactive schemas can write:

```ts
    const schemas: Set<SchemaInterface> = new Set();
    function callback(schema: SchemaInterface) {
      if (!schemas.has(schema.json())) {
        schemas.add(schema);
      }
    }
    traverseAsyncApiDocument(this, callback);
    const schemas = new Schemas(Array.from(schemas));
```

smoya: Sure they can but I don't think we should take this decision on return only the ones used by channels but rather all of them. Then, thanks to filters, the could filter after all.

magicmatatjahu: 🤷🏼

smoya: We had a chat on Slack about it. As it ended up making sense, I'm gonna create those filters not only for `Schemas` but also for `Channels`, `Operations`, `Messages`, and `Servers`, also a PR to `parser-api` reflecting the change.
Can you @magicmatatjahu then make this method to return all schemas?

smoya: https://github.com/asyncapi/parser-js/pull/619

jonaslagoni (approved):

src/models/v2/asyncapi.ts
smoya: @magicmatatjahu would you have some time to review the PR mentioned in my previous component, so we can unblock this? 🙏

smoya: I think you can adapt the code of this PR to consider https://github.com/asyncapi/parser-js/pull/619 is gonna happen so we can merge this right? It is not really blocked in fact.

magicmatatjahu: https://github.com/asyncapi/parser-js/pull/619#discussion_r985692933

src/models/v2/components.ts
```diff
@@ -76,9 +76,7 @@ export class Components extends BaseModel<v2.ComponentsObject> implements Compon
   }
 
   operations(): OperationsInterface {
-    const operations: OperationInterface[] = [];
-    this.channels().forEach(channel => operations.push(...channel.operations().all()));
-    return new Operations(operations);
+    return new Operations([]);
```
magicmatatjahu, line 79: We don't have `operations` in `components` in v2, so we should return empty collection.

src/models/v2/asyncapi.ts
```diff
@@ -91,7 +96,60 @@ export class AsyncAPIDocument extends BaseModel<v2.AsyncAPIObject> implements As
     return this.createModel(Components, this._json.components || {}, { pointer: '/components' });
   }
 
+  allServers(): ServersInterface {
```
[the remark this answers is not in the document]
magicmatatjahu: In your example you will have two servers, `servers.myServer` and `components.servers.anotherServer`. We only compare the data of given object and we base on that filtering of these same "references".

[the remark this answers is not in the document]
smoya: So we compare the references and not the value of the json object?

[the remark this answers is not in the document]
magicmatatjahu: yep, these same references == these same JSON objects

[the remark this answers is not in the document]
magicmatatjahu: You can check tests for `allSchemas` or `allMessages` :)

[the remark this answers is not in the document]
smoya: Then, all good! Thanks for the explanation.

smoya (approved):
Awesome work @magicmatatjahu ! LGTM!

PR #615, opened 2022-09-15
refactor: export cjs and esm modules

magicmatatjahu:
**Description**

Export `cjs` and `esm` (with TS typings) modules from package. It's needed to packages which will consume parser-js. `ESM` is for website and treeshakable apps and `CJS` is for NodeJS app (but never versions can use ESM).

cc @smoya 

**Related issue(s)**
Part of #481 
Part of #482

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,4 +4,6 @@ node_modules
 /docs
 /coverage
 /lib
+/esm
+/cjs
 /browser
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,6 @@ node_modules
 /docs
 /coverage
 /lib
+/esm
+/cjs
 /browser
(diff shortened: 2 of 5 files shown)


jonaslagoni (approved):
Is there a reason why you don't use the same behavior as we have with other TS projects i.e. having it under `lib/cjs` etc 🤔 ?

Not that's it's important but you know, consistency is great 😆

magicmatatjahu:
> Is there a reason why you don't use the same behavior as we have with other TS projects i.e. having it under lib/cjs etc 🤔 ?

Tbh our other projects had bad decision to put everything to lib, separate `esm` and `cjs` folders is "better" for me and I saw that solution in another projects. Also by this decision we can easily support https://nodejs.org/api/packages.html#package-entry-points (in the future) but I checked and that feature doesn't work very well with current TS entry points and in old NodeJS versions. So yeah, wait 6months and we can enable it :)

PR #622, opened 2022-09-20
refactor: add tests for old api

magicmatatjahu:
**Description**

- add tests for old api
- fix found bugs in old and new api

cc @smoya 

**Related issue(s)**
Part of #481 
Part of #482


src/iterator.ts
```diff
@@ -39,7 +39,7 @@ export enum SchemaTypesToIterate {
 
 export type TraverseOptions = {
   callback: TraverseCallback
-  schemaTypesToIterate: SchemaTypesToIterate[]
+  schemaTypesToIterate: Array<`${SchemaTypesToIterate}`>;
```
smoya, line 42: I'm lost on this 😆 Is this the preferable way? Or it was a bug?

magicmatatjahu: It's an interpolation of enum values as union. Feature in the TS 4.1, so you can put `schemaTypesToIterate` as array of normal string (but with values from enum) or `SchemaTypesToIterate` signature for given enum value.

smoya (approved):
LGTM! 🚀🌔

PR #643, opened 2022-10-06
fix: change schema.uid() method to schema.id()

smoya:
**Description**

This PR renames the `schema.uid()` method to become `schema.id()` to be consistent with all the other models (which are using `id()` as well).

The change in the `parser-api` can be seen [here](https://github.com/asyncapi/parser-api/pull/71/commits/0ae8f98405ed49e26bb46b7d73eea5efd51f1d4f).


magicmatatjahu (approved):
🚀

PR #654, opened 2022-10-13
chore: add support for asyncapi 3.0.0 operations and channels

fmvilas:
This PR contains the following changes:

* Move all the collections from `v2` directory to the `models` directory because their implementation is always the same. Therefore we can reuse them for `v3` and beyond. Check out 0791adca4f6187946b990eec764b6f56d4e055af for more details.
* Adds v3 models to work with v3 operations and channels.
* Adds `npm run dev` so it recompiles on file change.

I split all the changes into meaningful commits to make the review more manageable.


src/models/base.ts
```diff
@@ -3,7 +3,7 @@ import type { DetailedAsyncAPI } from '../types';
 
 export interface ModelMetadata {
   asyncapi: DetailedAsyncAPI;
-  pointer: string;
+  pointer?: string;
```
fmvilas, line 6: I found some cases in which `pointer` will be undefined and had to do this. Happy to hear about alternatives.

src/models/channel-parameters.ts
```diff
@@ -1,4 +1,10 @@
...
 
-export type ChannelParametersInterface = Collection<ChannelParameterInterface>
\ No newline at end of file
+export type ChannelParametersInterface = Collection<ChannelParameterInterface>;
+
+export class ChannelParameters extends Collection<ChannelParameterInterface> implements ChannelParametersInterface {
+  override get(id: string): ChannelParameterInterface | undefined {
+    return this.collections.find(parameter => parameter.id() === id);
+  }
+}
```
fmvilas, line 10: You'll find a ton of these too. I found that these collection implementations are always the same so I'm moving them from v2 to the models' directory. Check out the first commit of this PR to see all the related changes.

src/models/operation-trait.ts
```diff
@@ -1,11 +1,10 @@
...
 import type { BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface } from './mixins';
-import { SecurityRequirements } from './v2/security-requirements';
+import { SecurityRequirements } from './security-requirements';
 
 export interface OperationTraitInterface extends BaseModel, BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface {
-  id(): string;
-  hasOperationId(): boolean;
-  operationId(): string | undefined;
+  id(): string | undefined;
+  hasId(): boolean;
```
fmvilas, line 7: I changed this because I think it's confusing. An operation shouldn't have an `id` and an `operationId`. The `operationId` is the `id` of the operation 😅 So yeah, changed it and changed tests too.

fmvilas:
BTW, this PR is not yet done. It's still missing some implementations (challenge you to find it 😄) and, of course, tests. I hate writing tests so if someone in their divine presence wants to help this old weary man, please do it by writing some tests.

jonaslagoni:
> I hate writing tests so if someone in their divine presence wants to help this old weary man

If you really hate writing tests, I can only suggest you try writing them first before the implementation. That way they don't become a chore but a positive experience 😄 At least in theory 😛

src/models/operation-trait.ts
smoya: Makes sense 👍 . Would you mind adding such a change into https://github.com/asyncapi/parser-api/pull/71 ?

fmvilas:
> > I hate writing tests so if someone in their divine presence wants to help this old weary man
> 
> If you really hate writing tests, I can only suggest you try writing them first before the implementation. That way they don't become a chore but a positive experience 😄 At least in theory 😛

Dude, I've been around for a while. Don't preach the TDD shit at me, please ✋ 😝

src/models/v3/operation.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  channels(): ChannelsInterface {
+    const channels: ChannelInterface[] = [];
+    const pointer = this._json.channel.address ? tilde(`/channels/${this._json.channel.address}`) : undefined;
+    channels.push(
+      this.createModel(Channel, this._json.channel, { id: this._json.channel['x-parser-id'], address: this._json.channel.address, pointer })
+    );
+    return new Channels(channels);
+  }
```
magicmatatjahu, line 76: Why are you doing it this way? Since we define with `$ref` channel in the operation, we only need to pass it to the collection without additional operations. Only problem is to retrieve proper pointer:

```suggestion
  channels(): ChannelsInterface {
    const channels = this._meta?.asyncapi?.parsed?.channels || {};
    let pointer: string = '';
    let channelId: string = '';
    for (const [channelName, channel] of Object.entries(channels)) {
      if (channel === this._json.channel) {
        channelId = channelName;
        pointer = `/channels/${channelName}`;
      }
    }
    const channel = this.createModel(Channel, this._json.channel, { id: channelId, pointer })
    return new Channels(channel);
  }
```

If there is anything you don't understand then let me know :)

Or I don't understand something.

magicmatatjahu (changes requested):
Good! Do not forget to refer to my comments and please add unit tests for new v3 models - for channel and operation will be enough.

EDIT: I will write some tests 😄

package.json
```diff
@@ -26,6 +26,7 @@
     "README.md"
   ],
   "scripts": {
+    "dev": "tsc --watch",
```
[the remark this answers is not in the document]
fmvilas: That's not for Jest but for development. I'm not running tests while developing because I'm not doing TDD 😄

src/models/operation-trait.ts
fmvilas: I don't think I understand 😅 Just to clarify, in v3 `operationId` is not mandatory, it just doesn't exist. The id comes from the key in the `operations` object.

magicmatatjahu: we have also `components.operations`, therefore `id` and `operationId` exist together.

package.json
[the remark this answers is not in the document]
magicmatatjahu: So why you need `--watch` for tsc? IDE itself should throw TS errors to you.

src/models/v3/operation.ts
fmvilas: Did you try this implementation? I have a couple of concerns with this:

1. You're only checking for existence in `this._meta?.asyncapi?.parsed?.channels` but the reality is that a `$ref` can be pointing anywhere, even another file. Actually, most probably another file in many cases.
2. `if (channel === this._json.channel) {` is this strict equality gonna work? With `===` the two have to be the same object instance in memory, right? Or am I missing something?

package.json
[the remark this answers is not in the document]
fmvilas: Yes, but only in the files (tabs) that I have open. This is throwing me errors in any file, even if not open.

src/models/base.ts
fmvilas: https://github.com/asyncapi/parser-js/pull/654/files#diff-d7ef7ffb6795ee82f8ee52244fb91a58a582e6b65a273d0ec162b0d04ae4c340R71-R73

src/models/operation-trait.ts
fmvilas: Yeah, something to check in this case with v2.

fmvilas: For sure, what doesn't make sense is to have `operation.id` and `operation.operationId`. They're the same thing with different names. BTW, the same will happen with `messageId` eventually.

src/models/v3/channel.ts
```diff
@@ -0,0 +1,90 @@
...
+    // ['publish', 'subscribe'].forEach(operationAction => {
+    //   const id =  this._json[operationAction as 'publish' | 'subscribe'] && (this._json[operationAction as 'publish' | 'subscribe'] as v2.OperationObject).operationId || `${this.meta().id  }_${  operationAction}`;
+    //   if (this._json[operationAction as 'publish' | 'subscribe']) {
+    //     operations.push(
+    //       this.createModel(Operation, this._json[operationAction as 'publish' | 'subscribe'] as v2.OperationObject, { id, action: operationAction as OperationAction, pointer: `${this._meta.pointer}/${operationAction}` }),
+    //     );
+    //   }
+    // });
+    return new Operations(operations);
+  }
```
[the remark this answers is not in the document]
fmvilas: Yes 😄 https://github.com/asyncapi/parser-js/pull/654#issuecomment-1277843988

Now seriously, just time. I wanted to share something as soon as possible for review because this PR was getting too big already.

src/models/v3/operation.ts
magicmatatjahu: Ok, I pushed changes to the PR in latest commit - I know that there are quite a lot of changes, but we had a lot of missed methods or models didn't implement correctly interfaces. I will write unit tests and fix sonarcloud issues, don't worry. If you wanna, you can check implementation and ask if you have questions.

src/models/operation-trait.ts
magicmatatjahu: from v2 spec point of view it makes sense and that's the problem 😄

src/models/v3/operation-trait.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  isSend(): boolean {
+    return this.action() === 'send';
+  }
+
+  isReceive(): boolean {
+    return this.action() === 'receive';
+  }
+
+  channels(): ChannelsInterface {
```
magicmatatjahu, line 65: Not that function returns empty collection, because we should discuss https://github.com/asyncapi/spec/issues/856 problem. Also in current implementation that function creates circular references between import:

```
OperationTrait -> Channel -> Operation -> OperationTrait
```

I will try to fix it if we will agree what about `channel` and `action` in Operation Trait Object.

magicmatatjahu (approved):
@fmvilas Ok, I wrote tests and also fixed all interfaces and corresponding models for v2 and v3 versions. I think we can merge it and fix all errors (especially sonar issues) in next PRs.

Also I changed title of PR to the `chore`, because:
- we don't validate new version of spec
- we don't have all implemented models

so package now is "broken". I hope that you understand :) Of course if you have any questions about implementation, why etc, feel free.

cc @smoya @jonaslagoni @derberg

src/models/operation-trait.ts
fmvilas: I don't think it does, even from v2 point of view. `operationId` was there to mean the `id` of the operation. If it's present, it should be the operation identifier. If it's not, ok, we can generate one if you want, but it was never meant to be an "extra" id.

fmvilas:
@magicmatatjahu what do you mean that the package is broken now? Will it fail if I try to simply import it? Or you mean that not everything is implemented?

src/models/operation-trait.ts
magicmatatjahu: Hmm 🤔 I have a problem in general because in the v3 version people will be able to define operations in components and how to distinguish this identifier in `components.operations[x]` from `operationId` - that's why we have these two methods.`id()` method is more of a method used to filter operations by identifier (used as key in the `components.operations[x]` and `operations[x]`), but... we'll see how to figure it out and `id()` should return `operationId`. Can we fix it in next PRs? We need also to fix that in `parser-api` and remove `operationId()` method.

To clarify: I wrote about `components.operations[x]` and `operations[x]`, because in `parser-api` we have method `allOperations()` which should return all operations defined in given AsyncAPI document.

magicmatatjahu:
For me it is also not a big problem to release, but why would anyone need a non-working model? Also simple validation is needed, because the `parser-api` itself is written in such a way that we have sure that the spec is valid.

magicmatatjahu:
@fmvilas So what? Can we merge it?

[abongoblob](https://emoji.gg/emoji/9509_abongoblob)

fmvilas:
> For me it is also not a big problem to release, but why would anyone need a non-working model? Also simple validation is needed, because the `parser-api` itself is written in such a way that we have sure that the spec is valid.

Nobody, except us, is using this version of the parser. It's not just that Parser is incomplete but v3 is also incomplete. It's ok, we're not gonna release it anytime soon 😄 

So yeah, I approve it but I created this PR so I'm not sure I should be approving it 😅

jonaslagoni (approved):
Did not see anything jumping out, only smaller formatting issues, but since we have no code formatting configuration it's not something that can be changed later 😄 

*pssst, smaller PRs preferred 😉*

magicmatatjahu:
@jonaslagoni Thanks!

src/models/operation-trait.ts
smoya: I realized no one updated the parser-api with any change regarding this.

smoya: All is fixed and there is no need to add anything else.

PR #656, opened 2022-10-19
feat: export `AsyncAPISchema` and `v2` types

smoya:
**Description**

This PR exports some types that are required by other packages. In particular, by [this avro-schema-parser PR](https://github.com/asyncapi/avro-schema-parser/pull/168).

**Related issue(s)**
https://github.com/asyncapi/avro-schema-parser/issues/166

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,6 +17,6 @@ export type { ValidateOptions, ValidateOutput } from './validate';
 export type { ParseOptions, ParseOutput } from './parse';
 export type { StringifyOptions } from './stringify';
 export type { ValidateSchemaInput, ParseSchemaInput, SchemaParser } from './schema-parser';
-export type { v2 as specTypesV2 } from './spec-types';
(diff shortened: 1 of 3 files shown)


magicmatatjahu:
@smoya Thanks for PR! I pushed some changes.
- I know that you need `AsyncAPiSchemaObject` and `...Definition` but we should export on the root only `AsyncAPiSchemaObject` type. Only `Avro` needs `...Definition` so you will have to import it by `specTypesV2.AsyncAPISchemaDefinition`
- I also exported `v2` types.

smoya:
> @smoya Thanks for PR! I pushed some changes.
> 
> * I know that you need `AsyncAPiSchemaObject` and `...Definition` but we should export on the root only `AsyncAPiSchemaObject` type. Only `Avro` needs `...Definition` so you will have to import it by `specTypesV2.AsyncAPISchemaDefinition`

I would love to understand the reason! 🙏

magicmatatjahu (approved):

PR #663, opened 2022-10-26
fix: change metadata of models and fix minor things

magicmatatjahu:
**Description**

- add `input` to the `asyncapi` metadata of models.
- `source` in model metadata now points to the source of file/document, not to the input.
- to boolean values we cannot assign things like `x-parser-schema-id` so I will add proper if in `setExtension` function.
- catch global errors from `validate` and `parse` functions.
- remove `version()` method from `Extension` model - we don't have clarified it in spec, so we shouldn't add that in current api.
- revert some changes in the `additionalItems` and `additionalProperties` methods from old Schema model - we havw to consistent with old logic.
- add unit tests.

cc @smoya


src/old-api/schema.ts
```diff
@@ -116,20 +116,15 @@ export class Schema extends SpecificationExtensionsModel<v2.AsyncAPISchemaObject
...
     return this.__createChild(additionalProperties);
   }
 
   additionalItems() {
-    if (typeof this._json === 'boolean') return this._json;
     const additionalItems = this.__get('additionalItems');
-    if (typeof additionalItems === 'boolean') return additionalItems;
-    if (additionalItems === undefined) return true;
-    if (additionalItems === null) return false;
+    if (additionalItems === undefined || additionalItems === null) return;
```
magicmatatjahu, line 127: We have to be consistent with implementation in v1 parser - current logic in `next-major` has breaking change in `additionalItems` and `additionalProperties` methods and some templates won't work.

magicmatatjahu: https://github.com/asyncapi/parser-js/blob/master/lib/models/schema.js#L221
https://github.com/asyncapi/parser-js/blob/master/lib/models/schema.js#L231

src/models/extension.ts
```diff
@@ -2,6 +2,5 @@ import type { BaseModel } from './base';
 
 export interface ExtensionInterface<T = any> extends BaseModel {
   id(): string;
-  version(): string;
```
smoya, line 5: Please ensure this change is proposed in https://github.com/asyncapi/parser-api/pull/71

magicmatatjahu: Done https://github.com/asyncapi/parser-api/pull/71#discussion_r1006894610 :)

src/utils.ts
```diff
@@ -86,6 +90,22 @@ export function toJSONPathArray(jsonPath: string): Array<string | number> {
...
+        message: `${message}. Name: ${err.name}, message: ${err.message}, stack: ${err.stack}`,
+        path: [],
+        severity: DiagnosticSeverity.Error,
+        range,
+      }
+    ];
+  }
+  return [];
+}
+
```
smoya, line 108: Super minor but readability improvement in my opinion:

```suggestion
export function createUncaghtDiagnostic(err: unknown, message: string, document?: Document): Diagnostic[] {
  if (!err instanceof Error) {
    return [];
  }
  
  const range: Diagnostic['range'] = document ? document.getRangeForJsonPath([]) as Diagnostic['range'] : Document.DEFAULT_RANGE;
  return [
    {
      code: 'uncaught-error',
      message: `${message}. Name: ${err.name}, message: ${err.message}, stack: ${err.stack}`,
      path: [],
      severity: DiagnosticSeverity.Error,
      range,
    }
  ];  
}
```

smoya (approved):
LGTM!! Just one very minor, feel free to discard

magicmatatjahu:
@smoya Thanks!

PR #669, opened 2022-11-07
fix: add missed methods to the SecurityScheme model

magicmatatjahu:
**Description**

Add missed methods to the `SecurityScheme` model - `name()`.

I don't know if we should have methods like `hasName`, `hasScheme` for every optional fields. What do you think?

Blocks https://github.com/asyncapi/markdown-template/pull/312

cc @smoya


jonaslagoni (approved):

smoya:
> I don't know if we should have methods like `hasName`, `hasScheme` for every optional fields. What do you think?

I think we should. Not only because of consistency but making life easier. 👍

smoya (approved):
LGTM! 🚀🌔

PR #671, opened 2022-11-08
fix: handle complex circular reference cases

### LINKED ISSUE #599: Support all edge cases with circular references
With switching to the Spectral, we found that we do not support all cases with circular references (or, more precisely, one particular case). Spectral has its own `ref-resolver` that does not resolve circular references like `json-schema-ref-parser`, but only leaves the `$ref` field intact at the point when such a reference is encountered. I described that problem in one of [issue](https://github.com/asyncapi/parser-js/pull/591#discussion_r964497593):

> Circular references you can only have with Schema Objects, so `'/components/channels/test'.includes('/channels/test')` is not a good example.
>
> You can create circular reference only when you point to the schema which is already used in parent path(s), so schema like:
>
> ```yaml
> type: object
> properties:
>   deepProperty:
>     type: object
>     properties:
>       circular:
>         $ref: '#/properties/deepProperty'
> ```
>
> is circular in `properties/deepProperty/properties/circular` path. That case is handled by first `if` in mentioned function
>
> Another case is related to Spectral and how it resolves circular references. We can have two two schemas, main one and external > (from file):
>
> ```yaml
> type: object
> properties:
>   fileProperty:
>     $ref: './fromfile.yaml'
>     
> # fromfile.yaml
> type: object
> properties:
>   deepProperty:
>     type: object
>     properties:
>       circular:
>         $ref: '#/properties/deepProperty'
> ```
>
> but in this case the circular reference after dereferencing has `$ref` field still with `'#/properties/deepProperty'` value, so it means that if we will check path we will have `'#/properties/deepProperty'` from file pov, not from main file pov, like `'#/properties/fileProperty/properties/deepProperty'` and due to this fact we need to also check (by second if) the subpath. I hope it's clear now. Above logic is dictated by Spectral "problems".

it's resolved in our parser side, but one case is very hard to resolve like:

```yaml
type: object
properties:
  fileProperty:
    $ref: './fromfile#/properties/circular1'
        
# fromfile.yaml
type: object
properties:
  circular1:
    type: object
    properties:
      circular:
        $ref: '#/properties/circular2'
  circular2:
    type: object
    properties:
      circular:
        $ref: '#/properties/circular1'
```

and we don't have any additional data (metadata?) to resolve such a reference. We should switch to the `json-schema-ref-resolver` in Spectral or fix that problem in Spectral's `ref-resolver` (but that resolver is deprecated). We can also wait for https://github.com/stoplightio/spectral/issues/1054

magicmatatjahu:
**Description**

Support edge circular reference cases:
- resolve circular refs to appropriate resolved object in JSON tree.
- update `SchemaV2` and `SchemaOld` models.
- remove unnecessary functions.
- update unit test.

**Related issue(s)**
Resolves #599


jonaslagoni (approved):
🔥

PR #674, opened 2022-11-09
fix: remove action and channel fields from Operation Trait Object

magicmatatjahu:
**Description**

Remove `action` and `channel` fields from Operation Trait Object.

**Related issue(s)**
https://github.com/asyncapi/spec/issues/856

### DIFF
--- a/src/models/operation-trait.ts
+++ b/src/models/operation-trait.ts
@@ -1,17 +1,11 @@
 import type { BaseModel } from './base';
-import type { ChannelsInterface } from './channels';
-import type { OperationAction } from './operation';
 import type { SecurityRequirements } from './security-requirements';
 import type { BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface } from './mixins';
 
 export interface OperationTraitInterface extends BaseModel, BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface, ExternalDocumentationMixinInterface, TagsMixinInterface {
-  id(): string | undefined;
   hasId(): boolean;
-  action(): OperationAction | undefined;
-  isSend(): boolean;
-  isReceive(): boolean;
+  id(): string | undefined;
   hasSummary(): boolean;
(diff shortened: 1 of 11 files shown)


src/models/operation.ts
```diff
@@ -1,14 +1,18 @@
...
 import type { OperationTraitInterface } from './operation-trait';
+import type { ChannelsInterface } from './channels';
 import type { ServersInterface } from './servers';
+import type { MessagesInterface } from './messages';
 
 export type OperationAction = 'send' | 'receive' | 'publish' | 'subscribe';
 
 export interface OperationInterface extends BaseModel, OperationTraitInterface {
-  action(): OperationAction;
+  action(): OperationAction | undefined;
```
jonaslagoni, line 11: Can operations really have undefined action? 🤔

magicmatatjahu: Yeah, right, I will fix that - it's a leftover when we could define action in Operation Trait.

magicmatatjahu: Ok, done :) Could you check again?

src/models/operation-trait.ts
```diff
@@ -1,17 +1,11 @@
...
-  id(): string | undefined;
   hasId(): boolean;
-  action(): OperationAction | undefined;
-  isSend(): boolean;
-  isReceive(): boolean;
+  id(): string | undefined;
   hasSummary(): boolean;
   summary(): string | undefined;
   security(): SecurityRequirements[];
-  channels(): ChannelsInterface;
```
jonaslagoni, line 16: Is there a reason why channels arent part of traits?

magicmatatjahu: https://github.com/asyncapi/spec/pull/860 we had PR for that in spec. There is also a link to the issue for it :)

magicmatatjahu: Maybe in the future we add such a information in ParserJS to check in which channels (where operations are assigned) given trait is applied, but not now :)

magicmatatjahu: You can even create PR in `parser-api` for such a method(s) :)

PR #677, opened 2022-11-22
chore: add missed metadata methods to the core models

magicmatatjahu:
**Description**

- add `title()`, `hasTitle()`, `summary()`, `hasSummary()`, `tags()` methods to the core models: `Channel`, `Server`, `Operation`.
- add `CoreMixins` interface and abstract `CoreModel` class to use inheritance.
- add unit tests for new methods.

**Related issue(s)**
Part of https://github.com/asyncapi/spec/issues/795
Part of https://github.com/asyncapi/spec/issues/750


fmvilas (approved):
I'm happy to accept this PR but I don't agree with this kind of "optimizations". This is one of these situations where DRY doesn't apply well and it's going to bite us further down the track.

You're grouping summary and description with unrelated stuff like externalDocs and bindings. All in a single place just to avoid repetition. It's all good now but code evolves over time and what's common today may not be common tomorrow and this PR is hiding all these "common" things making it harder to spot.

Just an opinion though. Time will tell 😄

magicmatatjahu:
Looking at the speed of change in the spec it will be in about 3 years 😆

PR #687, opened 2022-12-08
feat: add isOldAsyncAPIDocument helper

magicmatatjahu:
**Description**

- add `x-parser-api-version` to have info about parser-api used in AsyncAPI model
- `isOldAsyncAPIDocument` helper function to check if document use old api
- improve `isAsyncAPIDocument`
- add unit tests

cc @jonaslagoni as you want :)


jonaslagoni (commented):
Do you want to add a converter as well? I.e. from old parser document to new version? The code you send works:

```ts

        // Is from old parser
        const parsedJSON = input.json();
        const detailed = createDetailedAsyncAPI(parsedJSON, parsedJSON);
        doc = createAsyncAPIDocument(detailed);
```

derberg:
one cent from me, please update readme where we document all custom parser extensions

magicmatatjahu:
@jonaslagoni I can add but it won't work in 100% cases.

src/document.ts
```diff
@@ -32,7 +34,25 @@ export function toAsyncAPIDocument(maybeDoc: unknown): AsyncAPIDocumentInterface
...
+  if (!maybeDoc) {
+    return false;
+  }
+  if (maybeDoc instanceof AsyncAPIDocumentV2 || maybeDoc instanceof AsyncAPIDocumentV3) {
+    return true;
+  }
+  if (maybeDoc && typeof (maybeDoc as AsyncAPIDocumentInterface).json === 'function') {
+    const versionOfParserAPI = (maybeDoc as AsyncAPIDocumentInterface).json()[xParserApiVersion];
+    return versionOfParserAPI === 1;
+  }
```
jonaslagoni, line 46: You dont want to reuse `isOldAsyncAPIDocument` here? 🤔

jonaslagoni (commented):
Just one comment, but other then that it LGTM 👍

src/document.ts
magicmatatjahu: why?

magicmatatjahu: We cannot reuse because we need to know if `maybeDoc` has `json()` method.

derberg (approved):

PR #692, opened 2023-01-24
test: add less restricted test for memory usage

magicmatatjahu:
**Description**

Test locally passes, but problem is in CI side, so I will update test to be less restricted. Hopefully it should fix problem.


magicmatatjahu:
@jonaslagoni We need that to go forward with 2.6.0 release

jonaslagoni (approved):

magicmatatjahu:
@jonaslagoni Thanks! 🤞🏼

PR #694, opened 2023-01-24
test: skip memory usage test

magicmatatjahu:
**Description**

Skip memory usage test due to problems. We don't want block the `2.6.0` release. We should check that after spec release what's going wrong.


magicmatatjahu:
@jonaslagoni https://github.com/asyncapi/parser-js/issues/695

PR #701, opened 2023-01-30
ci: add cli to ignore repos in bump workflow

magicmatatjahu:
**Description**

As in title: add cli to ignore repos in bump workflow.

**Related issue(s)**
See https://github.com/asyncapi/cli/pull/376#issuecomment-1408733804

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -11,7 +11,7 @@ on:
   # This means we cannot extract easily latest version and have a risk that package is not yet on npm
   push:
     branches:
-      - master
+      - next-major
 
 jobs:
   bump-in-dependent-projects:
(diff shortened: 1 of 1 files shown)


derberg:
@magicmatatjahu changes in this PR should be done in the other one, and changes from the other one in this one

What is the plan? is RC supposed to be updated in all the AsyncAPI projects or only cli, modelina, studio?

magicmatatjahu:
> changes in this PR should be done in the other one, and changes from the other one in this one

Why? One PR should release `next-major` version in some repos, in another repos we should still bump `parser-js` from master.

derberg:
- then in `next-major` make sure workflow reacts on `next-major` and ignores repos that do not use 2.0 (like `html-template`)
- then in `master` make sure workflow reacts on `master` and ignores repos that use 2.0 (like `cli`)

magicmatatjahu:
@derberg Done, please check :)

derberg (approved):

PR #702, opened 2023-01-30
ci: add cli to ignore repos in bump workflow

magicmatatjahu:
**Description**

As in title: add cli to ignore repos in bump workflow.

**Related issue(s)**
See https://github.com/asyncapi/cli/pull/376#issuecomment-1408733804


derberg (commented):
what about others like modelina, or studio? I guess we do not want to bump `parser-js` master version anywhere or are there exceptions?

magicmatatjahu:
@derberg Done, please check :)

derberg (approved):

PR #711, opened 2023-02-01
chore: update release branch

AceTheCreator:
Update branch with new changes from Master.

cc @derberg


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/parser",
-  "version": "1.17.2",
+  "version": "1.17.3",
```
derberg, line 3: can you remove it, we never change version in any package manually

derberg (approved):

PR #714, opened 2023-02-03
fix: message-examples rule and check unused security schemes in 2.x.x

magicmatatjahu:
**Description**

- fix `asyncapi2-message-examples` rule and check unused security schemes in `2.x.x` document (new rule)
- update unit tests

cc @jonaslagoni


src/ruleset/v2/functions/messageExamples.ts
```diff
@@ -6,6 +6,23 @@ import type { IFunctionResult, RulesetFunctionContext } from '@stoplight/spectra
...
+      schema = {};
+    }
+  } else if (typeof schema === 'boolean') { // spectral cannot handle boolean schemas
+    if (schema === true) {
+      schema = {}; // everything
+    } else {
+      schema = { not: {} }; // nothing
+    }
+  }
+  return schema;
```
jonaslagoni, line 23: Wtf 😆

magicmatatjahu: You only can pass object as schema, yeah weird but it's easier to handle that in this way rather than contribution :P

src/ruleset/functions/unusedComponent.ts
```diff
@@ -22,6 +23,11 @@ export const unusedComponent = createRulesetFunction<{ components: Record<string
 
     const results: IFunctionResult[] = [];
     Object.keys(components).forEach(componentType => {
+      // if component type is `securitySchemes` and we operate on AsyncAPI 2.x.x skip validation
```
jonaslagoni, line 26: Can you add the `why` here as well? Just to clarify why it is we skip it.

jonaslagoni (changes requested):
Just one comment 👍

src/ruleset/functions/unusedComponent.ts
magicmatatjahu: Added :)

jonaslagoni (approved):
👍

Issue #405, opened 2021-11-10
References are bundled before validated


jonaslagoni:
#### Describe the bug
Currently, the parser bundles any references together with the provided AsyncAPI document before it's validated against the [JSON Schema document for AsyncAPI](https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.2.0.json).

This results in invalid AsyncAPI documents being valid.

#### How to Reproduce
Given the following AsyncAPI document: 
```yaml
asyncapi: '2.2.0'
info:
  title: Test overriding dereferenced objects 
  version: '1.0.0'
servers:
  test:
    $ref: 'testServer.yaml'
```
Together with the server file:
```yaml
url: ws://mycompany.com/ws
protocol: ws
```

The parser accepts the AsyncAPI document as being valid because we bundle the reference together with the document, before it is validated.

The spec however does not allow that servers to be referenced: https://www.asyncapi.com/docs/specifications/v2.2.0#serversObject

NOTE: You cannot try this in the studio because it automatically validates beyond what the parser does, so the studio catches such mistakes.

jonaslagoni:
Not quite sure about this, but solving this **might** also solve #404, as additional properties would not be allowed in most cases.

NektariosFifes:
@jonaslagoni hey i think i will look this up

smoya:
@jonaslagoni could you remove the `stale` label? Thanks!

magicmatatjahu:
Will be fixed by https://github.com/stoplightio/spectral/pull/2262

jonaslagoni:
Closing, as it was changed in https://github.com/asyncapi/parser-js/pull/720

PR #720, opened 2023-02-07
feat: validate unparsed AsyncAPI document

magicmatatjahu:
**Description**

Validate unparsed AsyncAPI document (before `$ref`s resolving).

### DIFF
--- a/test/ruleset/rules/asyncapi-document-unresolved.spec.ts
+++ b/test/ruleset/rules/asyncapi-document-unresolved.spec.ts
@@ -1,90 +1,85 @@
-// import { testRule, DiagnosticSeverity } from '../tester';
+import { testRule, DiagnosticSeverity } from '../tester';
 
-describe('asyncapi-document-unresolved', () => {
-  it.todo('add tests after fixing https://github.com/asyncapi/spec-json-schemas/issues/296');
-});
+testRule('asyncapi-document-unresolved', [
+  {
+    name: 'valid case',
+    document: {
+      asyncapi: '2.0.0',
+      info: {
+        title: 'Valid AsyncApi document',
+        version: '1.0',
+      },
+      channels: {
+        someChannel: {
+          publish: {
+            message: {
+              $ref: '#/components/messages/someMessage',
+            },
+          },
+        },
+      },
+      components: {
+        messages: {
+          someMessage: {},
+        },
+      },
+    },
+    errors: [],
+  },
 
-// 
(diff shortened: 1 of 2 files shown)


magicmatatjahu:
@jonaslagoni Could you check?

test/ruleset/rules/asyncapi-document-unresolved.spec.ts
```diff
@@ -1,90 +1,85 @@
...
+      },
+      components: {
+        'x-operations': {
+          someOperation: {},
+        },
+      },
+    },
+    errors: [
+      {
+        message: 'Referencing in this place is not allowed',
```
jonaslagoni, line 79: Why is this needed? Is it not caught by the spec-schema?

test/ruleset/rules/asyncapi-document-unresolved.spec.ts
```diff
@@ -1,90 +1,85 @@
...
```
jonaslagoni, line 53: Why is this needed? Is it not caught by the spec-schema?

jonaslagoni (changes requested):
I guess I dont understand why we don't change the parsing logic to validate before resolving references, and then again after? This fix looks like an incorrect approach to take 🤔 Or is there some underlying reason?

Related to https://github.com/asyncapi/parser-js/issues/405

magicmatatjahu:
@jonaslagoni 

> I guess I dont understand why we don't change the parsing logic to validate before resolving references, and then again after? This fix looks like an incorrect approach to take 🤔 Or is there some underlying reason?

Spectral resolver (and other json-schema resolvers) doesn't validate the json-schema but only resolve it. Also AJV try to validate the json-schema based on json-schema and that's the problem. After referencing we don't know if `$ref` in given place can be located and/or given ref has valid object (for given location, e.g. message etc.). So we exactly validate two things: unresolved AsyncAPI document with refs and with resolved refs (even that deep). I don't know how to make it in different way. That PR is related to the https://github.com/asyncapi/parser-js/issues/405

jonaslagoni:
Ahh shit, validation happens through spectral now, yea then you cant do a simple double validation once before refs are resolved and once after...

magicmatatjahu:
@jonaslagoni 

> Ahh shit, validation happens through spectral now, yea then you can't do a simple double validation once before refs are resolved and once after...

this is what I do 😄

jonaslagoni (approved):

PR #723, opened 2023-02-09
fix: update models with latest parser-api

magicmatatjahu:
**Description**

Update models with latest `parser-api`:
- make `location` in CorrelationId required
- add missed methods to the OAuthFlow object
- update Operation and OperationTrait objects
- add tests

**Related issue(s)**
See https://github.com/asyncapi/parser-api/blob/master/docs/v1.md
and https://github.com/asyncapi/parser-api/pull/86

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,14321 @@
 {
   "name": "@asyncapi/parser",
   "version": "2.0.0-next-major.13",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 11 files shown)


jonaslagoni (commented):
Why the package-lock change? 🤔

magicmatatjahu:
@jonaslagoni `v1` -> `v2` - we have problem because workflows always install deps with npm v6 (with v1 package-lock).

jonaslagoni (approved):

PR #726, opened 2023-02-13
feat: use spec json schema version 5

derberg:
Related to https://github.com/asyncapi/spec-json-schemas/pull/298

### DIFF
--- a/lib/asyncapiSchemaFormatParser.js
+++ b/lib/asyncapiSchemaFormatParser.js
@@ -68,7 +68,7 @@ function getMimeTypes() {
 function getValidator(version) {
   let validate = ajv.getSchema(version);
   if (!validate) {
-    const payloadSchema = preparePayloadSchema(asyncapi[String(version)], version);
+    const payloadSchema = preparePayloadSchema(asyncapi[String(version)]);
     ajv.addSchema(payloadSchema, version);
     validate = ajv.getSchema(version);
   }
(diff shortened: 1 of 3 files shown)


lib/asyncapiSchemaFormatParser.js
```diff
@@ -81,17 +81,14 @@ function getValidator(version) {
...
  */
-function preparePayloadSchema(asyncapiSchema, version) {
-  const payloadSchema = `http://asyncapi.com/definitions/${version}/schema.json`;
+function preparePayloadSchema(asyncapiSchema) {
   const definitions = asyncapiSchema.definitions;
-  // Remove the meta schemas because it is already present within Ajv, and it's not possible to add duplicate schemas.
-  delete definitions['http://json-schema.org/draft-07/schema'];
-  delete definitions['http://json-schema.org/draft-04/schema'];
+
+  //we return $ref and "definitions" and not only "definitions.schema" because some properties of "schema" refer to other definitions, like json schema one
```
derberg, line 89: adding a comment for the future as I forgot why we did it like this in the first place and was like: "how the hell does it work this way, I will change to definitions.schema" 😄

lib/asyncapiSchemaFormatParser.js
```diff
@@ -81,17 +81,14 @@ function getValidator(version) {
...
- * @param {Object} version AsyncAPI version.
  * @returns {Object} valid JSON Schema document describing format of AsyncAPI-valid schema for message payload
  */
-function preparePayloadSchema(asyncapiSchema, version) {
-  const payloadSchema = `http://asyncapi.com/definitions/${version}/schema.json`;
+function preparePayloadSchema(asyncapiSchema) {
   const definitions = asyncapiSchema.definitions;
-  // Remove the meta schemas because it is already present within Ajv, and it's not possible to add duplicate schemas.
-  delete definitions['http://json-schema.org/draft-07/schema'];
-  delete definitions['http://json-schema.org/draft-04/schema'];
```
derberg, line 92: no need to remove it as we do not use `http://json-schema.org/draft-07/schema` in $id any longer

smoya:
We would need to duplicate the changes in `next-major` branch.

derberg:
closing as it is no longer relevant

this pr addresses first implementation in  https://github.com/asyncapi/spec-json-schemas/pull/298 that was breaking current schemas. Now there are alternative schemas added

Issue #740, opened 2023-03-30
Parser 2.0 does not validate channel name


ivangsa:
#### Describe the bug
Parser 2.0 has a bug and do not validate channel name

#### How to Reproduce
This was reported here
https://github.com/asyncapi/vs-asyncapi-preview/issues/166

#### Expected behavior
It return a validation error like version 1.x: 

```6.TestChannel?Content-Type=application/json;charset=utf-8 channel contains invalid name with url query parameters: ?Content-Type=application/json;charset=utf-8```

derberg:
my guess is that https://github.com/asyncapi/parser-js/blob/master/lib/customValidators.js#L424 is not reflected in spectral rules for Parser 2 RC

magicmatatjahu:
ParserJS v2 validates it, but it's treated as warning, not error - when you put the mentioned AsyncAPI document you will see this:

![image](https://user-images.githubusercontent.com/20404945/228857583-89a4902c-6cf7-4525-8fe5-b41999510b00.png)

derberg:
Yeah, but that is an error, spec says there `SHALL NOT` be query params. Parser 1 throws error, we should not have regression here. In case of custom validators from version 1 I remember that there was only once case where instead of error we should show warning

magicmatatjahu:
So it's only a changing the severity kind of given rule, in this place - https://github.com/asyncapi/parser-js/blob/next-major/src/ruleset/v2/ruleset.ts#L290

derberg:
@magicmatatjahu should it also be move under `v2CoreRuleset` ?

magicmatatjahu:
Can be moved.

derberg:
What needs to be done to solve:

Add `severity: 'error'` to https://github.com/asyncapi/parser-js/blob/next-major/src/ruleset/v2/ruleset.ts#L290 and move it under `v2CoreRuleset`

/gfi ts

smoya:
A PR with the fix has been created https://github.com/asyncapi/parser-js/pull/747

PR #747, opened 2023-04-13
fix: change severity for channel name validation rule

smoya:
**Description**

Fixes https://github.com/asyncapi/parser-js/issues/740.

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/740


derberg (approved):
Thanks a lot! 🍺

PR #778, opened 2023-06-01
fix: skip message object example validation for non default schemaFormats

### LINKED ISSUE #772: message examples invalid against avro payload
#### Describe the bug
When supplying an avro payload for a message, and supplying examples that match that payload the examples cause a validation error:
```
schema is invalid: data/type must be equal to one of the allowed values, data/type must be array, data/type must match a schema in anyOf
```
- If I comment out the examples the async api validates correctly and generates an example that looks exactly like the one supplied. 
- If I comment out the payload and leave the examples the async api validates correctly
- If I switch the avro payload to a json payload everything validates correctly 

#### How to Reproduce
Use this example api:
```yaml
asyncapi: 2.6.0
info:
  description: Orders which have been placed on our webstore
  title: Orders
  version: 1.0.0
tags:
  - name: Webstore
  - name: Product
channels:
  Orders:
    subscribe:
      message:
        schemaFormat: "application/vnd.apache.avro;version=1.9.0"
        payload: {"type":"record","name":"WindRecordings","fields":[{"name":"direction","type":{"type":"enum","name":"directionEnum","symbols":["North","East","South","West"]}},{"name":"speed","type":"string"}]}
        examples: [
          {
            payload: {
              "direction": "North",
              "speed": "18"
            }
          }
        ]
```

Screenshot 2023-05-17 at 14 01 40

#### Expected behaviour
I would expect that the parser would validate the examples against the avro schema and my async api would be valid.

#### Notes
I tried to look through the code to find where the issue was coming from, and my best guess is that https://github.com/asyncapi/parser-js/blob/20be09d64077d4933cc093cf3d5d87d4ea17d99c/src/ruleset/v2/functions/messageExamples.ts#L79 is not taking into consideration that the schema may be avro. Given that this may actually be a feature request, or I may be completely wrong

chrispatmore:
**Description**
Small change to skip the validation of message example objects when a `schemaFormat` is
set. This allows the parser to validate schemas with non default schemaFormats that are also
providing examples.

This does mean if anyone manually sets `schemaFormat` to the default the examples will not
be checked. Also it does not validate the examples against the schema for non default formats
so they may be incompatible 

**Related issue(s)**
Fixes #772

### DIFF
--- a/src/ruleset/v2/ruleset.ts
+++ b/src/ruleset/v2/ruleset.ts
@@ -116,24 +116,24 @@ export const v2CoreRuleset = {
      * Message Object rules
      */
     'asyncapi2-message-examples': {
-      description: 'Examples of message object should follow by "payload" and "headers" schemas.',
+      description: 'Examples of message object should validate againt the "payload" and "headers" schemas.',
       message: '{{error}}',
       severity: 'error',
       recommended: true,
       given: [
         // messages
-        '$.channels.*.[publish,subscribe].message',
-        '$.channels.*.[publish,subscribe].message.oneOf.*',
-        '$.components.channels.*.[publish,subscribe].message',
-        '$.components.channels.*.[publish,subscribe].message.oneOf.*',
-        '$.components.messages.*',
+        '$.channels.*.[publish,subscribe].[?(@property === \'message\' && @.schemaFormat === void 0)]',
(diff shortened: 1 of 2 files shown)


chrispatmore:
FYI raised the fix @jonaslagoni

jonaslagoni:
Sorry @chrispatmore, too many notifications, thanks for the ping 👍

jonaslagoni (approved):
LGTM 👍

Issue #781, opened 2023-06-06
Validate default schema format examples


chrispatmore:
#### Reason/Context
As an extension to #772 instead of not validating non default schema format examples. the parser should validate that the example is valid w.r.t to the schema regardless of format. This gives increased confidence that the complete document is valid.

#### Description
To do this a new rule will have to be introduced which like the `src/schema-parser/spectral-rule-v2.ts` rule will take in the parser instance. This will be used to parse the schema in the provided format to the default format, which can then be used to validate the examples against the schema.

The old rule can renamed to something like `asyncapi2-message-examples-default` and the new rule would be `asyncapi2-message-examples` like the `asyncapi2-schemas` rule and equivalent `asyncapi2-schemas-default` rule there at the moment.

This would be a breaking change for anyone that had an invalid example for a schema provided with the non default schema format. However it was only possible to do that after #772 was completed, so it is unlikely there will be any people in this situtation at the moment.

I have code ready for this that can be reviewed alongside the issue. It didn't take that long so I won't be at all bothered if the decision is not do this feature

chrispatmore:
@jonaslagoni I have raised this extension feature request as discussed
I can create a PR for this feature now or wait for discussion on whether to do it

chrispatmore:
No I don't think it's required, will raise it

PR #782, opened 2023-06-07
feat: validate message object examples with spectral schemas

chrispatmore:
validate message object examples against the schemas in the
object by parsing the schemas using spectral before validating
this proves that the examples are valid.

**Related issue(s)**
Contributes to: #781

Signed-off-by: Chris Patmore <chrism.patmore@btinternet.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "webapi-parser": "^0.5.0"
   },
   "devDependencies": {
+    "@asyncapi/avro-schema-parser": "^3.0.1",
     "@jest/types": "^29.0.2",
     "@swc/core": "^1.2.248",
     "@swc/jest": "^0.2.22",
--- a/src/ruleset/v2/functions/messageExamples-spectral-rule-v2.ts
+++ b/src/ruleset/v2/functions/messageExamples-spectral-rule-v2.ts
@@ -1,14 +1,13 @@
 import { RuleDefinition, createRulesetFunction } from '@stoplight/spectral-core';
-import { schema as schemaFn } from '@stoplight/spectral-functions';
 
 import type { JsonPath } from '@stoplight/types';
-import type { IFunctionResult, RulesetFunctionContext } from '@stoplight/spectral-core';
-import type { JSONSchema7 } from 'json-schema';
+import type { IFunctionResult } from '@stoplight/spectral-core';
 import type { v2 } from 'spec-types';
 import type { Parser } from 'parser';
 import type { DetailedAsyncAPI } from 'types';
 import { ParseSchemaInput, getDefaultSchemaFormat, getSchemaFormat, parseSchema } from '../../../schema-parser';
 import { createDetailedAsyncAPI } from '../../../utils';
+import { getMessageExamples, validate } from './messageExamples';
 
 export function asyncApi2MessageExamplesParserRule(parser: Parser): RuleDefinition {
   return {
@@ -136,38 +135,4 @@ async function parseExampleSchema(parser: Parser, schema: unknown, type: 'payloa
     };
     return {path, schema: undefined, errors: [error]};
   }
-}
-
-function getMessageExamples(message: v2.MessageObject): Array<{ path: JsonPath; value: v2.MessageExampleObject }> {
-  if (!Array.isArray(message.examples)) {
-    return [];
-  }
-  return (
-    message.examples.map((example, index) => {
-      return {
-        path: ['examples', index],
-        value: example,
-      };
-    }) ?? []
-  );
-}
-
-function validate(
-  value: unknown,
-  path: JsonPath,
-  type: 'payload' | 'headers',
-  schema: unknown,
-  ctx: RulesetFunctionContext,
-): ReturnType<typeof schemaFn> {
-  return schemaFn(
-    value,
-    {
-      allErrors: true,
-      schema: schema as JSONSchema7,
-    },
-    {
-      ...ctx,
-      path: [...ctx.path, ...path, type],
-    },
-  );
 }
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "webapi-parser": "^0.5.0"
   },
   "devDependencies": {
+    "@asyncapi/avro-schema-parser": "^3.0.1",
```
chrispatmore, line 63: Added so I can test that the rule against a non default schema Format, there may be another way to do this I am not sure?

src/ruleset/v2/ruleset.ts
```diff
@@ -238,6 +239,7 @@ export const v2SchemasRuleset = (parser: Parser) => {
           },
         },
       },
+      'asyncapi2-message-examples-custom-format': asyncApi2MessageExamplesParserRule(parser),
```
chrispatmore, line 242: Added the rule in but might not want to do this yet. If not, need a pointer on how to get the tests to run without being in the ruleset

chrispatmore (commented):
Couple of comments / questions for others

chrispatmore:
There is code duplication w.r.t to the old rule. I can deal with this after review / thoughts on what's there

chrispatmore:
has duplication which I can sort out with other comments. and is out of date with base now I think, but any comments on the code are appreciated @jonaslagoni

jonaslagoni:
I dont think it changed, I just need to find some time to review it ASAP 😅

chrispatmore:
This will probably need some slight changes / merge conflict resolutions from https://github.com/asyncapi/parser-js/pull/786

jonaslagoni:
@chrispatmore fixed the conflicts, gonna review this once I get back from the gym 🙂

jonaslagoni (changes requested):
@chrispatmore would it make sense to add the rule ONLY for when schemaFormat has been defined? i.e. that is the difference between them?

chrispatmore:
Let me take a look. Forcing me to play around with json path expressions again 😉

chrispatmore:
That code smell is false positive, if I remove the `as any` it won't build

chrispatmore:
How did removing code increase duplication...

chrispatmore:
@jonaslagoni thoughts on duplication in the tests?

package.json
jonaslagoni: Dont think so no 😄

test/ruleset/rules/v2/asyncapi2-message-examples.spec.ts
```diff
@@ -345,12 +345,24 @@ testRule('asyncapi2-message-examples', [
...
                 ]
               },
+              headers: {
+                type: 'record',
+                name: 'TestHeader',
+                fields: [
+                  {name: 'someKey', type: 'string'},
+                  {name: 'someOtherKey', type: 'string'}
+                ]
+              },
```
jonaslagoni, line 355: What was this change for? 🤔

test/ruleset/rules/v2/asyncapi2-message-examples.spec.ts
```diff
@@ -378,19 +390,32 @@ testRule('asyncapi2-message-examples', [
                   {name: 'speed', type: 'string'}
                 ]
               },
+              headers: {
```
jonaslagoni, line 393: Pretty sure headers cannot be avro 🤔 At least not in v2

src/ruleset/v2/functions/messageExamples-spectral-rule-v2.ts
```diff
@@ -0,0 +1,138 @@
...
+import type { v2 } from 'spec-types';
+import type { Parser } from 'parser';
+import type { DetailedAsyncAPI } from 'types';
+import { ParseSchemaInput, getDefaultSchemaFormat, getSchemaFormat, parseSchema } from '../../../schema-parser';
+import { createDetailedAsyncAPI } from '../../../utils';
+import { getMessageExamples, validate } from './messageExamples';
+
+export function asyncApi2MessageExamplesParserRule(parser: Parser): RuleDefinition {
+  return {
+    description: 'Examples of message object should validate against the "payload" and "headers" schemas.',
```
jonaslagoni, line 14: We agree this rule now only are `Examples of message object when using explicit schema format should validate correctly against payloads` right? 🤔

test/ruleset/rules/v2/asyncapi2-message-examples.spec.ts
chrispatmore: I think I added this when the same tests were being used for both rules, to ensure no change in behaviour, less relevant now, can remove

test/ruleset/rules/v2/asyncapi2-message-examples.spec.ts
chrispatmore: No I added the headers to the original tests (schemaFormat not set case) so that It added confidence that header behaviour was maintained across both versions of the rule (before we decided to split the rules)
I can revert this.

Though the new info that headers cannot be v2 means I can simplify the new v2 rule!

chrispatmore:
No worries @jonaslagoni not in any particular rush. Have adapted to your comments.

The headers cant be avro was interesting, because my rule validating the headers examples against an avro header schema was perfectly happy with it!

Have removed it regardless as, as you say it's not allowed in the spec atm

I expect we will still have quite high duplication in the test spect

chrispatmore:
Hey @jonaslagoni , I am back from a little break, what are we thinking here?

jonaslagoni:
Thanks for the patience @chrispatmore, sorry it took so long 🙏

chrispatmore:
Thanks @jonaslagoni no worries !

PR #798, opened 2023-07-13
Feat/remove schema parsers

smoya:
**Description**

This PR finishes an old cleanup task that, for some reason, got lost during some merge/update before releasing v2.
This task is about removing the several schema parsers from this code base as now they all leave back in their own repository. See https://github.com/asyncapi/parser-js/issues/482#issuecomment-1225503177

Technically, this could be considered as a BC since we are removing code others could be using. However, we never exposed that, not even in the docs, where we ask people to install from each package. See https://github.com/asyncapi/parser-js#official-supported-custom-schema-parsers.
I would love to release this not as v3.0.0 but just v2.1.0, but I'm open-minded. 

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/482#issuecomment-1225503177


smoya:
Removing. Brain was hacked beacuse I used a wrong branch when navigating.

PR #814, opened 2023-07-24
feat: add support for multi-format schema

smoya:
**Description**
This PR adds the required changes for supporting [Multi-format schemas](https://github.com/asyncapi/spec/issues/622).
The API changes are in this`parser-api` PR https://github.com/asyncapi/parser-api/pull/86 waiting to be reviewed.

Replaces https://github.com/asyncapi/parser-js/pull/785

**Related issue(s)**
https://github.com/asyncapi/spec/issues/622

### DIFF
--- a/src/models/schema.ts
+++ b/src/models/schema.ts
@@ -49,6 +49,7 @@ export interface SchemaInterface extends BaseModel<v2.AsyncAPISchemaObject>, Ext
   propertyNames(): SchemaInterface | undefined;
   readOnly(): boolean | undefined;
   required(): Array<string> | undefined;
+  schemaFormat(): string
   then(): SchemaInterface | undefined;
   title(): string | undefined;
   type(): string | Array<string> | undefined;
--- a/src/models/v2/message-trait.ts
+++ b/src/models/v2/message-trait.ts
@@ -5,7 +5,6 @@ import { MessageExample } from './message-example';
 import { Schema } from './schema';
 
 import { xParserMessageName } from '../../constants';
-import { getDefaultSchemaFormat } from '../../schema-parser';
 import { bindings, hasDescription, description, extensions, hasExternalDocs, externalDocs, tags } from './mixins';
 
 import type { BindingsInterface } from '../bindings';
@@ -18,10 +17,15 @@ import type { SchemaInterface } from '../schema';
 import type { TagsInterface } from '../tags';
 
 import type { v2 } from '../../spec-types';
+import { getDefaultSchemaFormat } from '../../schema-parser';
 
 export class MessageTrait<J extends v2.MessageTraitObject = v2.MessageTraitObject> extends BaseModel<J, { id: string }> implements MessageTraitInterface {
   id(): string {
     return this.messageId() || this._meta.id || this.json(xParserMessageName) as string;
+  }
+
+  hasSchemaFormat(): boolean {
+    return true;
   }
 
   schemaFormat(): string {
--- a/src/models/v2/message-trait.ts
+++ b/src/models/v2/message-trait.ts
@@ -6,6 +6,7 @@ import { Schema } from './schema';
 
 import { xParserMessageName } from '../../constants';
 import { bindings, hasDescription, description, extensions, hasExternalDocs, externalDocs, tags } from './mixins';
+import { getDefaultSchemaFormat } from '../../schema-parser';
 
 import type { BindingsInterface } from '../bindings';
 import type { CorrelationIdInterface } from '../correlation-id';
@@ -15,8 +16,6 @@ import type { MessageExamplesInterface } from '../message-examples';
 import type { MessageTraitInterface } from '../message-trait';
 import type { SchemaInterface } from '../schema';
 import type { TagsInterface } from '../tags';
-
-import { getDefaultSchemaFormat } from '../../schema-parser';
 
 import type { v2 } from '../../spec-types';
 
--- a/src/models/v2/message-trait.ts
+++ b/src/models/v2/message-trait.ts
@@ -25,10 +25,10 @@ export class MessageTrait<J extends v2.MessageTraitObject = v2.MessageTraitObjec
   }
 
   hasSchemaFormat(): boolean {
-    return true;
+    return this.schemaFormat() !== undefined;
   }
 
-  schemaFormat(): string {
+  schemaFormat(): string | undefined {
     return this._json.schemaFormat || getDefaultSchemaFormat(this._meta.asyncapi.semver.version);
   }
 
--- a/src/models/v2/message-trait.ts
+++ b/src/models/v2/message-trait.ts
@@ -16,8 +16,9 @@ import type { MessageTraitInterface } from '../message-trait';
 import type { SchemaInterface } from '../schema';
 import type { TagsInterface } from '../tags';
 
-import type { v2 } from '../../spec-types';
 import { getDefaultSchemaFormat } from '../../schema-parser';
+
+import type { v2 } from '../../spec-types';
 
 export class MessageTrait<J extends v2.MessageTraitObject = v2.MessageTraitObject> extends BaseModel<J, { id: string }> implements MessageTraitInterface {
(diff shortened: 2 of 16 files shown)


src/models/v2/message-trait.ts
```diff
@@ -18,12 +17,17 @@ import type { SchemaInterface } from '../schema';
 import type { TagsInterface } from '../tags';
 
 import type { v2 } from '../../spec-types';
+import { getDefaultSchemaFormat } from '../../schema-parser';
```
magicmatatjahu, line 20: Please move that import above the type imports, here and in other files :) Only for consistency in source code.

src/models/v2/schema.ts
```diff
@@ -8,8 +8,10 @@ import type { ExternalDocumentationInterface } from '../external-docs';
 import type { SchemaInterface } from '../schema';
 
 import type { v2 } from '../../spec-types';
+import { getDefaultSchemaFormat } from '../../schema-parser';
+import { DetailedAsyncAPI } from 'types';
 
-export class Schema extends BaseModel<v2.AsyncAPISchemaObject, { id?: string, parent?: Schema }> implements SchemaInterface {
+export class Schema extends BaseModel<v2.AsyncAPISchemaObject, { id?: string, asyncapi?: DetailedAsyncAPI, parent?: Schema, schemaFormat?: string }> implements SchemaInterface {
```
magicmatatjahu, line 14: `DetailedAsyncAPI` you have in `_meta` - it has all models https://github.com/asyncapi/parser-js/blob/master/src/models/base.ts#L12

src/models/v2/schema.ts
```diff
@@ -267,6 +269,10 @@ export class Schema extends BaseModel<v2.AsyncAPISchemaObject, { id?: string, pa
     return this._json.required;
   }
 
+  schemaFormat(): string {
+    return this._meta.schemaFormat || getDefaultSchemaFormat(this._meta.asyncapi.semver.version);
+  }
```
magicmatatjahu, line 274: We can even has `schemaFormat` method to recognize if user define schemaFormat field or not (default one as fallback). If is defined, it should return true, otherwise false.

magicmatatjahu (commented):
I won't to request changes to not block you, but I have some comments. Very well! :)

src/models/v2/schema.ts
smoya: Why this would be needed? I see this is simple enough isn't?

src/models/v2/schema.ts
smoya: fixed

src/models/v2/message-trait.ts
smoya: Fixed

src/models/v2/schema.ts
magicmatatjahu: How you can recognize if someone define schema? 😄 I mean, it's not needed, we can add it later, but I think there can be cases to support that.

smoya: > We can even has schemaFormat method to recognize if user defined schemaFormat field or not (default one as fallback). If is defined, it should return true, otherwise false.

But you suggest to add a `hasSchemaFormat()` method?

magicmatatjahu: Yes :)

smoya: We had a discussion about this in the other PR and we realized it wasn't necessary. That's why the signature of `schemaFormat()` method returns always a `string` because we assume there is always a format needed even though the user didn't explicitly added.
I can't find any use case for supporting what you suggest here.

jonaslagoni: Dont really see a reason to check whether a schemaFormat is defaulting 🤨 I would probably say let a feature request decide if that's needed 😄 

As you said, we can easily add it later 👍

src/models/v2/message-trait.ts
jonaslagoni: Pretty sure this is not solved? as I understand what @magicmatatjahu asks for is that `import { getDefaultSchemaFormat } from '../../schema-parser';` should be above line 10, i.e. the `import type { BindingsInterface } from '../bindings';`. right? 🤔

test/models/v2/message.spec.ts
```diff
@@ -26,20 +26,6 @@ describe('Message model', function() {
     });
   });
 
-  describe('.schemaFormat()', function() {
```
jonaslagoni, line 29: Why remove the tests when we still have the functions? 🤔

test/models/v3/message.spec.ts
```diff
@@ -26,20 +20,6 @@ describe('Message model', function() {
     });
   });
 
-  describe('.schemaFormat()', function() {
```
jonaslagoni, line 29: Why remove the tests when we still have the functions? 🤔

src/models/schema.ts
```diff
@@ -49,6 +49,7 @@ export interface SchemaInterface extends BaseModel<v2.AsyncAPISchemaObject>, Ext
   propertyNames(): SchemaInterface | undefined;
   readOnly(): boolean | undefined;
   required(): Array<string> | undefined;
+  schemaFormat(): string
```
jonaslagoni, line 52: Didnt think about it until now, but it's so similar to `$schema` just with different values 🙃

jonaslagoni (changes requested):
Small comments 👍

src/models/v2/message-trait.ts
magicmatatjahu: exactly, above all type imports

src/models/schema.ts
magicmatatjahu: $schema is metadata of `JSON Schema`. These two fields should not be "merged".

jonaslagoni: Agree 😄 

But it's just who very similar values in some cases and in others not at all, weird combo to have.

jonaslagoni: Feels like a hack, but no way around it when we transform everything to JSON Schema 😆

smoya: Not a big issue imho. We could create a model that inherits from Schema but includes that method, but at the end is the same.

test/models/v2/message.spec.ts
smoya: Reverted. good catch.

test/models/v3/message.spec.ts
smoya: The reality is that message **should not** have the `schemaFormat` method but only the MessageTrait. See https://github.com/asyncapi/parser-api/pull/86/files#diff-ac35ac6e0b0ad3629802a0e04076aa18e94da614f5392e1164dcaacdfe3ee198L157

test/models/v2/message.spec.ts
smoya: The reality is that message **should not** have the `schemaFormat` method but only the MessageTrait. See https://github.com/asyncapi/parser-api/pull/86/files#diff-ac35ac6e0b0ad3629802a0e04076aa18e94da614f5392e1164dcaacdfe3ee198L157

smoya: I think i will need to change the way inheritance is being made atm...

smoya: After thinking again and again, makes sense to keep `schemaFormat` + add `hasSchemaFormat` to `messageObject` as well so we reduce the BC impact. I applied the changes + changes in parser-api. See https://github.com/asyncapi/parser-api/pull/86#issuecomment-1651893775

smoya: So this test method is back now

smoya:
@magicmatatjahu @jonaslagoni this is now ready for review again 🙏

jonaslagoni (approved):
LGTM, @magicmatatjahu do you want to have a look again?

jonaslagoni:
Merging, if you have anything @magicmatatjahu it can be addressed afterwards 🙂

Issue #815, opened 2023-07-25
Implement new simplified channel parameters


smoya:
Apply the changes introduced by https://github.com/asyncapi/spec/pull/935

Regarding the changes in Parser-JS, we don’t really need to do a lot because we can keep the current API -> `schema(): Schema | undefined`
However, things like `parameter.schema().example()` will now require to return the value from the `parameter.example()` which is basically the parent level.

In that way, we don’t need to add a breaking change in [the parser-api](https://github.com/asyncapi/parser-api/pull/86). Internally, we can create a new object called `simplifiedSchema` or similar, that inherits the behaviour from the `Schema` object, but only “allowing” the enum, default, and examples methods, logging a warning message when using the others. 

Remember that including any non-supported field in the parameter will make any ASyncAPI v3 document to become invalid, so it is not expected developers call any other method from the schema (but still, logging a warning message its a good idea)

PR #816, opened 2023-07-25
feat: add simplified channel parameter

jonaslagoni:
**Description**
This PR adds the support for the simplified channel parameters by utilizing a custom implementation of Schema that just looks in the parent parameter object for the values instead of under a `.schema` property that no longer exist for v3.

**Related issue(s)**
Fixes https://github.com/asyncapi/parser-js/issues/815


smoya:
Would you consider throwing warning log lines for those methods that users are unexpected to use?

jonaslagoni:
@smoya I considered it, but I could not justify it with any reason. Cause it's not like the functions are deprecated or anything 🤷 They just wont return anything if paired with v3 documents, but will with v2.

So not sure who the logging is for?

magicmatatjahu:
@jonaslagoni I don't know if it was discussed but you can reuse v3.SchemaModel but passing as `json` object the serialized schema object from new parameter shape, I mean map fields like type etc:

```
{
  type: "string"
  description: parameter.description
}
```

etc.

Creating new object for this is not a good idea.

jonaslagoni:
@magicmatatjahu you right, thats way cleaner, let me refactor it.

jonaslagoni:
Done 👍

src/models/v3/channel-parameter.ts
```diff
@@ -7,7 +7,7 @@ import type { SchemaInterface } from '../schema';
 import type { ExtensionsInterface } from '../extensions';
 
 import type { v3 } from '../../spec-types';
-import { ParameterSchema } from './channel-parameter-schema';
+import { Schema } from './schema';
```
magicmatatjahu, line 10: Import in wrong place 😄

magicmatatjahu (commented):
Only one comment, rest is good :)

src/models/v3/channel-parameter.ts
jonaslagoni: Done 👍

src/models/v3/channel-parameter.ts
```diff
@@ -15,12 +13,21 @@ export class ChannelParameter extends BaseModel<v3.ParameterObject, { id: string
...
+    );
   }
 
   schema(): SchemaInterface | undefined {
-    if (!this._json.schema) return undefined;
-    return this.createModel(Schema, this._json.schema, { pointer: `${this._meta.pointer}/schema` });
+    if (!this.hasSchema()) return undefined;
+    return this.createModel(Schema, {
+      type: 'string', 
+      enum: this._json.enum,
```
magicmatatjahu, line 27:
```suggestion
      description: this._json.description,
      enum: this._json.enum, 
```

magicmatatjahu (changes requested):

jonaslagoni:
@magicmatatjahu do you have a bit of OCD by any chance? 😄

magicmatatjahu (approved):

PR #821, opened 2023-08-01
feat: match Spectral formats based on schemas found in @asyncapi/specs pkg

smoya:
**Description**

This PR removes the need for creating new Spectral ruleset formats matching new AsyncAPI versions. 
This translates to the removal of the requirement of making changes every time we add a new version into `@asyncapi/specs` package. It might be some exceptions in the case we, for example, add a new major spec version, where rules won't apply and manual changes (and expected I would say) will need to be done.

This PR enables adding support for v3 of the spec in a **very** easy way (in a next PR). In fact, this allows a v3 doc to be parsed but not validated.

cc @jonaslagoni @magicmatatjahu @Amzani @fmvilas

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -19,4 +19,4 @@ export const EXTENSION_REGEX = /^x-[\w\d.\-_]+$/;
 
 // Only >=2.0.0 versions are supported
 export const specVersions = Object.keys(specs.schemas);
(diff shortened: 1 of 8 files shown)


smoya:
Changing base to `master` but in a new PR.

PR #825, opened 2023-08-01
feat: enable AsyncAPI v3 validation

smoya:
Tests are failing. Read comments below.

### DIFF
--- a/src/ruleset/formats.ts
+++ b/src/ruleset/formats.ts
@@ -9,10 +9,6 @@ export class Formats extends Map<string, Format> {
     return new Formats([...this.entries()].filter(element => {return majorsToInclude.includes(element[0].split('.')[0]);}));
   }
 
-  excludeByMajorVersions(versionsToExclude: string[]): Formats {
-    return new Formats([...this.entries()].filter(element => {return !versionsToExclude.includes(element[0].split('.')[0]);}));
-  }
-
   excludeByVersions(versionsToExclude: string[]): Formats {
     return new Formats([...this.entries()].filter(element => {return !versionsToExclude.includes(element[0]);}));
   }
--- a/src/ruleset/formats.ts
+++ b/src/ruleset/formats.ts
@@ -9,6 +9,10 @@ export class Formats extends Map<string, Format> {
     return new Formats([...this.entries()].filter(element => {return majorsToInclude.includes(element[0].split('.')[0]);}));
   }
 
+  excludeByMajorVersions(versionsToExclude: string[]): Formats {
+    return new Formats([...this.entries()].filter(element => {return !versionsToExclude.includes(element[0].split('.')[0]);}));
+  }
+
   excludeByVersions(versionsToExclude: string[]): Formats {
     return new Formats([...this.entries()].filter(element => {return !versionsToExclude.includes(element[0]);}));
   }
--- a/test/ruleset/formats.spec.ts
+++ b/test/ruleset/formats.spec.ts
@@ -79,15 +79,4 @@ describe('AsyncAPIFormats collection', () => {
       expect(filteredFormats.find(version)).toBeFalsy();
     });
   });
-
-  it('Excludes by major version', () => {
-    const excludedMajorVersions = ['3'];
-    const previousLenght = AsyncAPIFormats.formats().length;
-    const filteredFormats = AsyncAPIFormats.excludeByMajorVersions(excludedMajorVersions);
-
-    expect(filteredFormats.size).toEqual(previousLenght - excludedMajorVersions.length);
-    filteredFormats.forEach((_, version) => {
-      expect(excludedMajorVersions).not.toContain(String(getSemver(version).major));
-    });
-  });
 });
--- a/test/ruleset/formats.spec.ts
+++ b/test/ruleset/formats.spec.ts
@@ -79,4 +79,15 @@ describe('AsyncAPIFormats collection', () => {
       expect(filteredFormats.find(version)).toBeFalsy();
     });
   });
+
+  it('Excludes by major version', () => {
+    const excludedMajorVersions = ['3'];
+    const previousLenght = AsyncAPIFormats.formats().length;
+    const filteredFormats = AsyncAPIFormats.excludeByMajorVersions(excludedMajorVersions);
+
+    expect(filteredFormats.size).toEqual(previousLenght - excludedMajorVersions.length);
+    filteredFormats.forEach((_, version) => {
+      expect(excludedMajorVersions).not.toContain(String(getSemver(version).major));
+    });
+  });
 });
--- a/test/ruleset/rules/asyncapi-document-resolved.spec.ts
+++ b/test/ruleset/rules/asyncapi-document-resolved.spec.ts
@@ -15,7 +15,7 @@ testRule('asyncapi-document-resolved', [
   },
 
   {
-    name: 'invalid case (channels property is missing)',
+    name: 'invalid case for 2.X.X (channels property is missing)',
     document: {
       asyncapi: '2.0.0',
       info: {
@@ -32,7 +32,7 @@ testRule('asyncapi-document-resolved', [
   },
 
   {
-    name: 'valid case (case when other errors should also occur but we filter them out)',
+    name: 'valid case for 2.X.X (case when other $ref errors should also occur but we filter them out)',
     document: {
       asyncapi: '2.0.0',
       info: {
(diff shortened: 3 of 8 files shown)


smoya:
Validation fails with the following errors:

```json
[
   {
      "code":"asyncapi-document-resolved",
      "message":"'someOperation1' property must match exactly one schema in oneOf",
      "path":[
         "operations",
         "someOperation1"
      ],
      "range":{
         "end":{
            "character":46,
            "line":20
         },
         "start":{
            "character":21,
            "line":10
         }
      },
      "severity":0
   },
   {
      "code":"asyncapi-document-resolved",
      "message":"'someOperation2' property must match exactly one schema in oneOf",
      "path":[
         "operations",
         "someOperation2"
      ],
      "range":{
         "end":{
            "character":46,
            "line":35
         },
         "start":{
            "character":21,
            "line":24
         }
      },
      "severity":0
   }
]
```
The validation rule that fails is the `asyncapi-document-resolved`, which checks if the AsyncAPI document is valid (using JSON Schema) **after resolving references**.
Tests are failing due to the fact `channel` property under `Operation` one should be just a reference to a channel and not a channel object directly, and the rule applies after dereferencing the channel object, which becomes an invalid document since the expected channel value should be a reference.

Thanks @jonaslagoni for helping (and actually finding) that out.

The error messages shown above are also wrong. We do some error filtering as stated in https://github.com/asyncapi/parser-js/blob/master/src/ruleset/functions/documentStructure.ts#L20, and it seems we are doing too much and, in this case, we skip the real error, which can be found in the list of errors prior to the filtering:

```json
[
  {
    "instancePath": "/operations/someOperation1",
    "schemaPath": "#/required",
    "keyword": "required",
    "params": {
      "missingProperty": "$ref"
    },
    "message": "must have required property '$ref'"
  },
  {
    "instancePath": "/operations/someOperation1/channel",
    "schemaPath": "#/required",
    "keyword": "required",
    "params": {
      "missingProperty": "$ref"
    },
    "message": "must have required property '$ref'"
  },
  {
    "instancePath": "/operations/someOperation1",
    "schemaPath": "#/additionalProperties/oneOf",
    "keyword": "oneOf",
    "params": {
      "passingSchemas": null
    },
    "message": "must match exactly one schema in oneOf"
  },
  {
    "instancePath": "/operations/someOperation2",
    "schemaPath": "#/required",
    "keyword": "required",
    "params": {
      "missingProperty": "$ref"
    },
    "message": "must have required property '$ref'"
  },
  {
    "instancePath": "/operations/someOperation2/channel",
    "schemaPath": "#/required",
    "keyword": "required",
    "params": {
      "missingProperty": "$ref"
    },
    "message": "must have required property '$ref'"
  },
  {
    "instancePath": "/operations/someOperation2",
    "schemaPath": "#/additionalProperties/oneOf",
    "keyword": "oneOf",
    "params": {
      "passingSchemas": null
    },
    "message": "must match exactly one schema in oneOf"
  }
]
```

The expected error should be the ones with an `instancePath` pointing to the operation channel, where the message says must have required property '$ref'.

smoya:
Based on my prior comment, this is what we need to do next in this PR:

1. To fix the bug in the error messages filter, so the error that stays is the one that points to each operation.channel field.
2. To skip validation of operation.channel for v3 in the `asyncapi-document-resolved` (still will happen in `asyncapi-document-unresolved`).

Alternatives for the second point will be to either allow by JSON Schema to define `operation.channel` as object aswell, which will then incompatible with the current spec, or rather to keep different JSON Schema files, one for pre and another for post reference resolution. I think this is also discarded as it is a lot of work and effort.

src/ruleset/formats.ts
```diff
@@ -9,6 +9,10 @@ export class Formats extends Map<string, Format> {
     return new Formats([...this.entries()].filter(element => {return majorsToInclude.includes(element[0].split('.')[0]);}));
   }
 
+  excludeByMajorVersions(versionsToExclude: string[]): Formats {
```
smoya, line 12: This is not needed anymore. My idea was to quickly disable the rule that fails for v3 and work on it later on. But we do have a solution (temporary?) in place now.

smoya:
I’m thinking that we should not go in the direction of manipulating the schema for the operation.channel in the mid term. Having to manipulate the schema from now on is not either good nor maintainable IMHO.

The alternative is to remove the `asyncapi-document-resolved` rule forever and just relay on the ``asyncapi-document-unresolved`. The point is that we are filtering errors [here](https://github.com/smoya/parser-js/blob/feat/enableV3Validation/src/ruleset/functions/documentStructure.ts#L86-L90), so we would need to return them and not filter. What is the real implication of it?

WDYT @magicmatatjahu @jonaslagoni

magicmatatjahu:
@smoya Issues are fixed :)

smoya:
> @smoya Issues are fixed :)

Thanks for the help @magicmatatjahu 🙌 🙌 !!

I still have [this concern](https://github.com/asyncapi/parser-js/pull/825#issuecomment-1667682844) and I would love to know your opinion as you are the one with most knowledge on this code.

I would like to have this PR merged but to keep iterating after so we go for a better and maintainable alternative.

cc @jonaslagoni @Amzani

smoya:
>  There is also a idea to allow inline objects in spec itself - but then 3.1.0 is needed.

Yeah, this is something I want to bring up and discuss with the community.

> `asyncapi-document-resolved` is very needed because we should also check the correctness of imported objects from external files/https (external refs). Without this rule how to check if something imported is ok?

Yeah, I agree there is no better alternative and, for sure, this validation is an important addition on top of the basic schema validation.

Maybe we could work on an automation, for every time we find a property with **only** a reference to a `Reference Object` , to modify that schema to allow the object as well (oneOf or just right the definition object as you did).

src/ruleset/formats.ts
jonaslagoni: If it's no longer needed, please remove it 😄

test/ruleset/formats.spec.ts
```diff
@@ -79,4 +79,15 @@ describe('AsyncAPIFormats collection', () => {
...
+
+  it('Excludes by major version', () => {
+    const excludedMajorVersions = ['3'];
+    const previousLenght = AsyncAPIFormats.formats().length;
+    const filteredFormats = AsyncAPIFormats.excludeByMajorVersions(excludedMajorVersions);
+
+    expect(filteredFormats.size).toEqual(previousLenght - excludedMajorVersions.length);
+    filteredFormats.forEach((_, version) => {
+      expect(excludedMajorVersions).not.toContain(String(getSemver(version).major));
+    });
```
jonaslagoni, line 91: This can be removed as well

test/ruleset/rules/asyncapi-document-resolved.spec.ts
```diff
@@ -309,4 +309,89 @@ testRule('asyncapi-document-resolved', [
...
+    errors: [
+      {
+        message: '"info" property must have required property "version"',
+        severity: DiagnosticSeverity.Error,
+      },
+    ],
+  },
+
+  {
+    name: 'valid case for 3.X.X (case when other $ref errors should also occur but we filter them out)',
```
jonaslagoni, line 370: Are these names on cases still valid?

jonaslagoni (changes requested):
As I see it there is one additional approach that is unmentioned, and that is we could make the reference resolver ignore the specific path i.e. 
- channelObject.properties.servers.items.$ref
- operationSchema.properties.channel.$ref
- operationSchema.properties.messages.items.$ref
- operationReplySchema.properties.channel.$ref
- operationReplySchema.properties.messages.items.$ref
Just for that specific rule, but it might be hard to pull of. There are also some more unknowns such as can the library we use do this? (I know https://apitools.dev/json-schema-ref-parser/docs/options.html can) Can we still ensure the refs are correct and point to a valid location?

So yea, I think your current solution makes sense 👍 

@smoya have you added enough test cases to test for all those paths and that they are caught?
> The alternative is to remove the asyncapi-document-resolved rule forever and just relay on the asyncapi-document-unresolved. The point is that we are filtering errors [here](https://github.com/smoya/parser-js/blob/feat/enableV3Validation/src/ruleset/functions/documentStructure.ts#L86-L90), so we would need to return them and not filter. What is the real implication of it?

[What @magicmatatjahu said](https://github.com/asyncapi/parser-js/pull/825#issuecomment-1669148409) 👍 

> There is also a idea to allow inline objects in spec itself - but then 3.1.0 is needed.

I don't see this as making much sense from a spec point of view, so it might be really hard to do just because reference resolvement becomes hard to do 😄

src/ruleset/formats.ts
smoya: Yup. Removed!

test/ruleset/formats.spec.ts
smoya: Done

test/ruleset/rules/asyncapi-document-resolved.spec.ts
magicmatatjahu: Spectral works in two modes - with schema resolved and unresolved (with untouched refs). Even if we wrote a ref in a place then Spectral would not throw an error (because it operates on resolved schema). This is more of a test for checking if rule isn't "broking" resolved/unresolved spectral behaviour, just in case for us :)

magicmatatjahu: If you want you can change the name of the test because it should have a different name, I can't remember why I called it that.

jonaslagoni (approved):
👍

PR #849, opened 2023-08-18
fix: set proper parser-api version

smoya:
**Description**

This PR fixes the parser-api version in use from `1` to `2.0.0`. It moves the value to a constant set in the same file as `AsyncAPIDocumentInterface` is declared.


jonaslagoni (approved):
Is this breaking? 🤔

smoya:
> Is this breaking? 🤔

Well, the breaking change is **already** in `next-major-spec` branch because it implements parser-api v2.

jonaslagoni:
@smoya 👍

PR #861, opened 2023-09-28
fix: isAsyncAPIDocument not recognizing correct documents

jonaslagoni:
**Description**
This PR fixes a problem where `isAsyncAPIDocument` cannot recognize parsed documents which means a lot of support functions are not working correctly.

### DIFF
--- a/src/models/asyncapi.ts
+++ b/src/models/asyncapi.ts
@@ -12,7 +12,7 @@ import type { ServersInterface } from './servers';
 import type { v2, v3 } from '../spec-types';
 
 // https://github.com/asyncapi/parser-api/releases/tag/v2.0.0
-export const ParserAPIVersion = '2';
+export const ParserAPIVersion = 2;
 
 export interface AsyncAPIDocumentInterface extends BaseModel<v2.AsyncAPIObject | v3.AsyncAPIObject>, ExtensionsMixinInterface {
   version(): string;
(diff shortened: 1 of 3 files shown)


src/models/asyncapi.ts
```diff
@@ -12,7 +12,7 @@ import type { ServersInterface } from './servers';
 import type { v2, v3 } from '../spec-types';
 
 // https://github.com/asyncapi/parser-api/releases/tag/v2.0.0
-export const ParserAPIVersion = '2.0.0';
+export const ParserAPIVersion = '2';
```
smoya, line 15: Why changing this back to the major? Having that granularity of a full semver, we will always know which parser-api version was supported when parsing. 
If you want to compare by major, you can `ParserAPIVersion.split('.')[0]`

smoya: Thinking deeply, I'm not sure if has a useful use case. If you don't see one (as I guess), we should also change the multi-parser-js library (I'm more than happy to accept a PR)

jonaslagoni: I dont imidiately see a reason to have the full number there, also that would mean we would need to change it for each change we do, not really maintainable 🤔

jonaslagoni: The multi parser implementation would still work right?

smoya (approved):
LGTM! 🚀🌔

Issue #859, opened 2023-09-07
[v3] Implement custom schema parsing


smoya:
#### Describe the bug
I found out the following commented code https://github.com/asyncapi/parser-js/blob/next-major-spec/src/custom-operations/index.ts#L44-L47.

That means the call to `parse()` on custom schema parsers (Avro, OpenAPI, etc) is not being called for v3. That means schemas with those schema formats are never parsed. It also means users and tools won't be able to access to the transformed schema (we transform all of them to AsyncAPI Schema).

#### Expected behavior
The call to `parse` should be implemented.

cc @magicmatatjahu

magicmatatjahu:
It wasn't implemented because we had to wait to final version of schemas (exactly multiple schemas) in v3

smoya:
This is then completely critical and requires to be implemented for the next release, otherwise schemas with a schema format different than AsyncAPI one won't be parsed, meaning raw schemas will be in place.

jonaslagoni:
/progress 90 finished and ready to be reviewed

jonaslagoni:
Done in https://github.com/asyncapi/parser-js/pull/864

PR #864, opened 2023-10-03
fix: schema parsing for v3 

jonaslagoni:
**Description**
This PR fixes a few things:

- Fixes schemas are not parsed when having different schema formats
- Fixes v3 message trait with headers was using incorrect types as they can now be multiformat schemas.

**Related issue(s)**
Fixes https://github.com/asyncapi/parser-js/issues/859


src/custom-operations/parse-schema.ts
```diff
@@ -22,6 +22,14 @@ const customSchemasPathsV2 = [
   '$.components.messages.*',
 ];
 
+const customSchemasPathsV3 = [
+  // operations
+  '$.channels.*.messages.*.payload',
+  '$.components.channels.*.messages.*.payload',
+  '$.components.messages.*.payload',
+  '$.components.schemas.*',
+];
```
smoya, line 31: Just wondering if we should not include headers and examples.

```suggestion
const customSchemasPathsV3 = [
  // channels
  '$.channels.*.messages.*.payload',
  '$.channels.*.messages.*.headers',
  '$.components.channels.*.messages.*.payload',
  '$.components.channels.*.messages.*.headers',
  // operations
  '$.operations.*.messages.*.payload',
  '$.operations.*.messages.*.headers',
  '$.components.operations.*.messages.*.payload',
  '$.components.operations.*.messages.*.headers',
  // messages
  '$.components.messages.*.payload',
  '$.components.messages.*.headers.*',
  '$.components.messages.*.examples.*.headers',
  '$.components.messages.*.examples.*.payload',
  // schemas 
  '$.components.schemas.*',
];
```

src/models/v3/schema.ts
```diff
@@ -47,7 +47,7 @@ export class Schema extends BaseModel<v3.MultiFormatSchemaObject, { id?: string,
 
   $schema(): string {
     if (typeof this._schemaObject === 'boolean') return 'http://json-schema.org/draft-07/schema#';
-    return this._schemaObject.$schema || 'http://json-schema.org/draft-07/schema#';
+    return this._schemaObject.$schema ?? 'http://json-schema.org/draft-07/schema#';
```
smoya, line 50: Is there an explicit case where you define the `$schema` field but with null or empty string? Just out of curiosity.

src/custom-operations/parse-schema.ts
jonaslagoni: But does any of these need custom schema parsing? Headers, you right 👍 Examples, those are actual values not schemas 🤔 

Also `'$.operations.*.messages.*.payload',` is not even a thing right? Unless references are resolved the schema parsing?

src/models/v3/schema.ts
jonaslagoni: No, at least not to my knowledge 🙂

smoya: So then why changing this?

jonaslagoni: Because sonar likes it better 🤷

src/custom-operations/parse-schema.ts
smoya: Resolution happens **after** applying custom operations (inclulding schema parsing), right?

jonaslagoni: It did not 😄 

Adapted the paths and tests accordingly.

smoya (approved):
LGTM!🙌

PR #868, opened 2023-10-06
fix: channel parameter not always returning schema

jonaslagoni:
**Description**
One thing that I found weird while working on https://github.com/asyncapi/markdown-template/pull/360 was that parameter schemas could be undefined with v3 🤔 

Cause it should always at least have the schema `{type: "string"}` and then the rest is optional.


test/models/v3/channel-parameter.spec.ts
```diff
@@ -60,10 +60,10 @@ describe('ChannelParameter model', function() {
       expect(d.hasSchema()).toEqual(true);
     });
 
-    it('should return false when there is no value', function() {
+    it('should return true when there is no value', function() {
       const doc = serializeInput<v3.ParameterObject>({});
       const d = new ChannelParameter(doc);
-      expect(d.hasSchema()).toEqual(false);
+      expect(d.hasSchema()).toEqual(true);
```
fmvilas, line 66: I agree with the other change but this one is highly debatable. "hasSchema" is meant to tell you if the document has the schema specified, therefore it should be false. However, even though it's false, you should get a default schema in the "schema()" call. These are two different things IMHO.

fmvilas (changes requested):

test/models/v3/channel-parameter.spec.ts
jonaslagoni: But for v3, you will always have a schema (underneath) as the default behavior is that the parameters are of `type: string` 🤔 Regardless of what else you define in your parameter.

For v2 it makes sense.

smoya: As @fmvilas mentioned, hasSchema is there for answering: Did the user set the schema?

jonaslagoni: But for v3 the schema is always set, even when an empty parameter is defined.

fmvilas: What I mean is if the user has set the `schema` key. `hasSchema` answers that question. If the `schema` key is there, then `hasSchema` is true, otherwise it's false. Now, the `schema()` call will always give you a schema because it has a default. Does that make sense?

fmvilas: Alright, discussed offline and this is fine. On v3, there's no `schema` keyword anymore so it should always be true.

smoya: > But for v3, you will always have a schema (underneath) as the default behavior is that the parameters are of `type: string` 🤔 Regardless of what else you define in your parameter.
> 
> For v2 it makes sense.

Oh yeah its true😅

PR #871, opened 2023-10-09
chore: bumped spec version to latest

AceTheCreator:
This PR bump next major for the parser with the latest spec package

cc @smoya


jonaslagoni (approved):
Lets use feat to trigger a release 🙂

smoya (approved):

PR #889, opened 2023-11-03
feat!: remove messageId()

fmvilas:
**Description**
Removes the `messageId()` function from the Message and MessageTrait objects. Keeps the same behavior on v2 but it doesn't look for `messageId` on v3. Now we only have one method to grab the id of a message and it's `.id()`.

Also, now `.hasMessageId()` now always returns `false` for v3.

**Related issue(s)**
https://github.com/asyncapi/spec/issues/978

### DIFF
--- a/src/models/v3/message-trait.ts
+++ b/src/models/v3/message-trait.ts
@@ -15,11 +15,11 @@ import type { v3 } from '../../spec-types';
 
 export class MessageTrait<J extends v3.MessageTraitObject = v3.MessageTraitObject> extends CoreModel<J, { id: string }> implements MessageTraitInterface {
   id(): string {
-    return this.messageId() || this._meta.id || this.extensions().get(xParserMessageName)?.value<string>() as string;
+    return this._meta.id || this.extensions().get(xParserMessageName)?.value<string>() as string;
   }
 
   hasMessageId(): boolean {
-    return !!this._json.messageId;
+    return false;
   }
 
   hasSchemaFormat(): boolean {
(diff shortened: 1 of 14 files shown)


src/models/v3/message-trait.ts
```diff
@@ -15,11 +15,11 @@ import type { v3 } from '../../spec-types';
...
 export class MessageTrait<J extends v3.MessageTraitObject = v3.MessageTraitObject> extends CoreModel<J, { id: string }> implements MessageTraitInterface {
   id(): string {
-    return this.messageId() || this._meta.id || this.extensions().get(xParserMessageName)?.value<string>() as string;
+    return this._meta.id || this.extensions().get(xParserMessageName)?.value<string>() as string;
   }
 
   hasMessageId(): boolean {
-    return !!this._json.messageId;
+    return false;
   }
```
jonaslagoni, line 23: Why is this function not removed? 🤔

jonaslagoni (commented):
Just one question, same as in https://github.com/asyncapi/parser-api/pull/102#discussion_r1382421659

src/models/v3/message-trait.ts
fmvilas: Because we want to keep the same API regardless of the AsyncAPI version, right? We need to keep the API stable independently of the version you're using.

smoya (approved):
LGTM! 🚀🌔

src/models/v3/message-trait.ts
jonaslagoni: Solved in https://github.com/asyncapi/parser-api/pull/102#discussion_r1382421659

jonaslagoni (approved):
👌

PR #892, opened 2023-11-07
fix: change ParserAPIVersion constant to 3

smoya:
**Description**

After recent changes made in https://github.com/asyncapi/parser-js/pull/889, this latest version of the Parser only supports Parser-API v3 instead of v2.

This PR fixes that.

**Related issue(s)**
https://github.com/asyncapi/parser-js/pull/889

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ The following table shows a compatibility matrix between this parser, and the [P
 | Parser-JS | Parser-API                                                            | Spec 2.x | Spec 3.x |
 |-----------|-----------------------------------------------------------------------|----------|----------|
 | 2.x       | [1.x](https://github.com/asyncapi/parser-api/blob/v1.0.0/docs/v1.md)  | ✓        |          |
-| 3.x       | [2.x](https://github.com/asyncapi/parser-api/blob/v2.0.0/docs/api.md) | ✓        | ✓        |
+| 3.x       | [3.x](https://github.com/asyncapi/parser-api/blob/v3.0.0/docs/api.md) | ✓        | ✓        |
 
 - `✓` Fully supported version.
 - `-` The AsyncAPI Spec version has features the Parser-JS can't use but the rest are fully supported.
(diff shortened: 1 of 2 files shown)


fmvilas:
Good catch, thanks! So, since you're here already, would you mind also changing this table? https://github.com/asyncapi/parser-js/blob/master/README.md#api-documentation

smoya:
> Good catch, thanks! So, since you're here already, would you mind also changing this table? https://github.com/asyncapi/parser-js/blob/master/README.md#api-documentation

Done! 🚀

README.md
```diff
@@ -23,26 +23,26 @@ Updated bundle for the browser is always attached to the GitHub Release.
...
-  * [Example with performing actions on HTTP source](#example-with-performing-actions-on-http-source)
-  * [Example with performing actions on file source](#example-with-performing-actions-on-file-source)
-  * [Example with stringify and unstringify parsed document](#example-with-stringify-and-unstringify-parsed-document)
+  - [Example with parsing](#example-with-parsing)
+  - [Example with validation](#example-with-validation)
+  - [Example using Avro schemas](#example-using-avro-schemas)
+  - [Example using OpenAPI schemas](#example-using-openapi-schemas)
+  - [Example using RAML data types](#example-using-raml-data-types)
+  - [Example with performing actions on HTTP source](#example-with-performing-actions-on-http-source)
+  - [Example with performing actions on file source](#example-with-performing-actions-on-file-source)
```
smoya, line 32: Ouch, all those changes.... VS Code made them. Reverting them...

README.md
```diff
@@ -23,26 +23,26 @@ Updated bundle for the browser is always attached to the GitHub Release.
...
+  - [Example with stringify and unstringify parsed document](#example-with-stringify-and-unstringify-parsed-document)
```
fmvilas, line 33: Someone has a prettifier installed 😄

fmvilas (dismissed):

README.md
smoya: Fixed

fmvilas (approved):

Issue #873, opened 2023-10-10
Message ID doesn't exist on reply level


AceTheCreator:
When using the request-reply pattern, and you're trying to retrieve the reply() messages, the parser is returning the wrong message ID for some reason. 

Here's what I mean by wrong messages. In the example below, I expect the parser to return the messageID of the messages available in the reply object, but instead, all I get is the index(`integer`) of the message as ID

```
operations:
    requestCosting:
      action: receive
      channel: 
        $ref: '#/channels/costingRequest'
      reply:
        channel: 
          $ref: '#/channels/costingResponse'
        messages:
          - $ref: "#/components/messages/costingResponse"
```

cc @smoya @fmvilas

smoya:
@AceTheCreator could you please provide a code snippet of what you are executing and the output? Thank you!

smoya:
Can this be somehow related to the issue under https://github.com/asyncapi/parser-js/issues/856 @jonaslagoni ?

jonaslagoni:
> The 0 returned by id is because you in this case have an array of messages where this one is the first index.
undefined is returned because, well, the message object does not contain a messageId property.

Whether this behavior is the right and expected behavior, we will have to chime in @magicmatatjahu 😄

jonaslagoni:
The problem then is the way references are currently being handled, is NOT to replace the reference in operations to channels and messages with the already resolved object that we created for the instance in the channel. 

We simply don't have access to the key in the key/value pair that is being referenced. Therefore the ONLY possible solution we can do is this with the tooling we have at hand.

Otherwise, we have to create our own reference tooling that can resolve partial references and manually control what instance should be returned.

derberg:
Just kind reminder that we knew that [using $ref everywhere will have a side effect that it will he harder to get the id of the element](https://github.com/asyncapi/spec/issues/829) and that we will have to [hook into dereferencing process](https://github.com/asyncapi/spec/issues/663#issuecomment-1163050817). So instead of waiting for or working on completely new tool that handles references, parser should grab id from the reference, and put in some extension like `x-parser-refId` or whatever

smoya:
> So instead of waiting for or working on completely new tool that handles references, parser should grab id from the reference, and put in some extension like `x-parser-refId` or whatever

[...]

PR #895, opened 2023-11-10
fix: use JSON Pointer last element as ID for reply.channel and reply.messages[]

smoya:
**Description**

Fixes https://github.com/asyncapi/parser-js/issues/873

This PR puts on practice one of the solutions we might opt in for getting Id's on referenced objects via the last element of their JSON Pointer. In this case, we are doing this for the `reply.channel` and `reply.messages[]`, since in the Operation Reply those are referenced via a JSON Pointer.

I'm also fixing a bug where `reply.messages` were expected to be a map instead of an array.

**Related issue(s)**
Fixes https://github.com/asyncapi/parser-js/issues/873

### DIFF
--- a/test/resolver.spec.ts
+++ b/test/resolver.spec.ts
@@ -1,7 +1,84 @@
-import { AsyncAPIDocumentV2 } from '../src/models';
+import { AsyncAPIDocumentV2, AsyncAPIDocumentV3 } from '../src/models';
 import { Parser } from '../src/parser';
 
 describe('custom resolver', function() {
+  it('Reply channel + messages[] Ids come from JSON Pointer', async function() {
+    const parser = new Parser();
+
+    const documentRaw = `{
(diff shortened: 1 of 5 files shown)


test/resolver.spec.ts
```diff
@@ -1,7 +1,84 @@
-import { AsyncAPIDocumentV2 } from '../src/models';
+import { AsyncAPIDocumentV2, AsyncAPIDocumentV3 } from '../src/models';
 import { Parser } from '../src/parser';
 
 describe('custom resolver', function() {
+  it('Reply channel + messages[] Ids come from JSON Pointer', async function() {
+    const parser = new Parser();
+
+    const documentRaw = `{
```
smoya, line 8: This doc is the one @AceTheCreator shared in https://github.com/asyncapi/parser-js/issues/873.
We could remove it or adapt it, but we need something to test this anyway.

Issue #896, opened 2023-11-13
document.allSchemas() returns "duplicated" schemas


smoya:
#### Describe the bug
`document.allSchemas()` method returns duplicated schemas when a schema is referenced from a component. 
In order to compare if two schemas are equal, we compare against their JSON representation (`object.json()`).

I compared the JSON string between duplicated and the only thing that changes is the pointer: one refers to the component section, the other to the place where the reference is.

#### How to Reproduce

Given the following doc:
```yaml
asyncapi: '2.6.0'
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    subscribe:
      message:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
```

Parse the document and print all schemas:

```ts
import { Parser, fromFile, SchemaInterface } from '@asyncapi/parser';

const parser = new Parser();
const parse = async () => {
    const { document } = await fromFile(parser, './asyncapi.yaml').parse();
    document?.allSchemas().forEach((schema: SchemaInterface) => {
      console.log(JSON.stringify(schema), '\n');
    });
};

parse();
```

This prints the following 6 schemas:

<details>
  <summary>Open</summary>

  ```js
[...]

PR #897, opened 2023-11-14
fix: doc.allSchemas() skips now duplicates

smoya:
**Description**

Fixes https://github.com/asyncapi/parser-js/issues/896

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/896

### DIFF
--- a/src/models/utils.ts
+++ b/src/models/utils.ts
@@ -1,5 +1,9 @@
 import type { BaseModel, ModelMetadata } from './base';
 import type { DetailedAsyncAPI } from '../types';
+import { SchemaInterface } from './schema';
+import { SchemaTypesToIterate, traverseAsyncApiDocument } from '../iterator';
+import { AsyncAPIDocumentInterface } from './asyncapi';
+import { SchemasInterface } from 'models';
 
 export interface Constructor<T> extends Function {
   new (...any: any[]): T;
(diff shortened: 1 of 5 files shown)


src/models/v3/asyncapi.ts
```diff
@@ -26,12 +26,12 @@ import type { MessageInterface } from '../message';
 import type { ComponentsInterface } from '../components';
 import type { SecuritySchemesInterface } from '../security-schemes';
 import type { ExtensionsInterface } from '../extensions';
-import type { SchemaInterface } from '../schema';
 import type { SchemasInterface } from '../schemas';
+import type { OperationInterface } from '../operation';
+import type { ChannelInterface } from '../channel';
+import type { ServerInterface } from '../server';
```
smoya, line 32: Just reordering and tiding up imports

src/models/utils.ts
```diff
@@ -12,3 +16,24 @@ export type InferModelMetadata<T> = T extends BaseModel<infer _, infer M> ? M :
 export function createModel<T extends BaseModel>(Model: Constructor<T>, value: InferModelData<T>, meta: Omit<ModelMetadata, 'asyncapi'> & { asyncapi?: DetailedAsyncAPI } & InferModelMetadata<T>, parent?: BaseModel): T {
   return new Model(value, { ...meta, asyncapi: meta.asyncapi || parent?.meta().asyncapi });
 }
+
+export function schemasFromDocument<T extends SchemasInterface>(document: AsyncAPIDocumentInterface, SchemasModel: Constructor<T>, includeComponents: boolean): T {
```
smoya, line 20: As the logic is exactly the same for both v2 and v3 models, I moved this to this common file.

test/models/v3/asyncapi.spec.ts
```diff
@@ -213,6 +213,13 @@ describe('AsyncAPIDocument model', function() {
       const d = new AsyncAPIDocument(doc);
       expect(d.allSchemas()).toBeInstanceOf(Schemas);
     });
+    it('should return a collection of schemas (with schemas from components) without duplicates', function() {
+      const sharedMessage = { payload: {} };
```
smoya, line 217: Remember, we compare by "reference" and not by value.

PR #902, opened 2023-11-16
fix: wrong parserAPI number being checked

jonaslagoni:
**Description**
This PR fixes that we hardcoded a number to check the parser API when checking whether a document is from this parser version or not. This causes the check to fail, even though it was correct.

**Related issue(s)**
Related to https://github.com/asyncapi/html-template/issues/456


smoya (approved):
Nice catch! LGTM

PR #908, opened 2023-11-20
fix: fix import statement

KhudaDad414:
with  the latest version of parser in `next-major-spec` I am getting this error:
https://github.com/asyncapi/glee/actions/runs/6929361068/job/18846911426?pr=589


smoya (approved):
Good catch! Thanks for the fix 🙌

PR #913, opened 2023-11-28
feat: add Spectral rules to validate required operation channe and channel servers field

smoya:
**Description**

Part of https://github.com/asyncapi/spec/issues/991

Despite the name of the branch, this PR adds two new Spectral rules (packed all together to minimize review efforts and to release asap).

This PR adds [the last two new  Spectral rules](https://github.com/asyncapi/spec/issues/991#issuecomment-1825700075) related with https://github.com/asyncapi/spec/issues/991.

1. Checks that the `channel` field of a required operation (under root) points to a required (under root) channel and not to an optional (under components) channel, no matter if from the same document or from an external.
2. Checks that the `servers` field of a required channel (under root) points to a subset of required (under root) servers.

**Related issue(s)**
https://github.com/asyncapi/spec/issues/991


src/ruleset/v3/ruleset.ts
```diff
@@ -24,5 +25,18 @@ export const v3CoreRuleset = {
         function: operationMessagesUnambiguity,
       },
     },
+    'asyncapi3-required-operation-channel-unambiguity': {
+      description: 'Required operation (under root channels) "channel" must reference to a required channel (under root channels).',
```
smoya, line 29: Moving the suggestion made by @jonaslagoni in the previous (closed) [PR](https://github.com/smoya/parser-js/pull/5#discussion_r1406717397).

```suggestion
      description: 'Operation channel's must reference to a required channel  (under root channels).'',
```

smoya: But this only applies to those operations defined as required (under root). So I still think we should clarify this.
I agree my sentence is not so good though.

smoya: Actually, I would even include those clarifications in the final error message the user will see, which is this one https://github.com/asyncapi/parser-js/pull/913/files#diff-e325fe55145146570971d2b1efd54035bbe2b005e557575279284a6f792c747bR35

fmvilas: Two things:
1. Don't apply this suggestion as is because that won't be valid JS. See the conflict with single quotes.
2. Let's make it explicit everywhere. No need to use optional and required words. It could be something like `The "channel" field of an operation under the root "operations" object must always reference a channel under the root "channels" object.`. Same for https://github.com/asyncapi/parser-js/pull/913/files#diff-e325fe55145146570971d2b1efd54035bbe2b005e557575279284a6f792c747bR35.

smoya: > The "channel" field of an operation under the root "operations" object must always reference a channel under the root "channels" object.

I like it

src/ruleset/v3/ruleset.ts
```diff
@@ -24,5 +26,19 @@ export const v3CoreRuleset = {
...
+      description: 'The "channel" field of an operation under the root "operations" object must always reference a channel under the root "channels" object.',
+      severity: 'error',
+      recommended: true,
+      resolved: false, // We use the JSON pointer to match the channel.
+      given: '$.operations.*',      
+      then: {
+        field: 'channel.$ref',
+        function: pattern,
+        functionOptions: {
+          match: '#\\/channels\\/', // If doesn't match, rule fails.
```
smoya, line 39: @fmvilas @jonaslagoni I simplified this by using this `pattern` function Spectral gives us. It allows to match by a regex applied to a field, in this case `channel.$ref`. 
In consequence, I removed the function I created in favor of this.

fmvilas: Much better 👍

fmvilas (approved):

fmvilas (approved):
😄

smoya:
> I'm sorry but I can't approve until I see "Spectral rules master" added as a skill in your LinkedIn.

Oops, something might went wrong. Can you please check again?

smoya:
@jonaslagoni @fmvilas FYI I've added to this PR also another rule that checks that the servers field of a required channel (under root) points to a subset of required (under root) servers.

src/ruleset/v3/ruleset.ts
```diff
@@ -24,5 +25,37 @@ export const v3CoreRuleset = {
...
+        },
+      },
+    },
+    
+    /**
+     * Channel Object rules
+     */
+    'asyncapi3-required-channel-servers-unambiguity': {
+      description: 'The "servers" field of a channel under the root "channels" object must always reference to a subset of the servers under the root "servers" object.',
```
fmvilas, line 47:
```suggestion
      description: 'The "servers" field of a channel under the root "channels" object must always reference a subset of the servers under the root "servers" object.',
```

fmvilas (changes requested):
Just this typo.

src/ruleset/v3/ruleset.ts
smoya: fixed

fmvilas (approved):

PR #916, opened 2023-11-30
Add new Spectral rule to check for null addresses in channels associated with operations defining reply

### LINKED ISSUE #875: Parser do not validate and throw error when `parameters` are provided but address is null
Like in the subject.

Document:
```yml
asyncapi: 3.0.0
info:
  title: Account Service
  version: 1.0.0
channels:
  userSignedup:
    address: user/signedup
    parameters:
      test:
    messages:
      UserSignedUp:
operations:
  sendUserSignedup:
    action: send
    channel:
    messages:
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
```

Gmin2:
Fixes #875


Min2who:
@smoya I guess i have understand the issue wrongly can you give me a brief idea of how can i test this issue locally

smoya:
Please reopen if applies

PR #937, opened 2024-01-10
fix!: expose all functions for browser

jonaslagoni:
**Description**
Right now it is not possible to use utility functions such as `fromURL` in browsers. This PR solves this by removing the default part of the export, and instead export the whole namespace: https://webpack.js.org/configuration/output/#outputlibraryexport

**Related issue(s)**
Related to https://github.com/asyncapi/EDAVisualiser/pull/54


jonaslagoni:
Still need to update docs as soon as we agree on an approach.

webpack_new.config.js
```diff
@@ -0,0 +1,48 @@
...
+  target: 'web',
+  mode: 'production',
+
+  output: {
+    path: path.resolve(__dirname, 'browser_new'),
+    filename: 'index.js',
+    globalObject: '(typeof self !== \'undefined\' ? self : this)',
+    library: {
+      name: 'AsyncAPIParser',
+      type: 'umd'
```
smoya, line 15: Is the lack of `export: 'default'` here making exporting just everything instead of the default export (which if I understand correctly is just `Parser`) ?

jonaslagoni: Yes exactly

smoya: Cool! Then, do we wanna expose every single method? Not sure. Then the package size will increment but not sure if this is really a critical aspect. Worth to check.

jonaslagoni: I cannot figure out how to partially expose the functionality i.e. only parser and fromURL...

> Cool! Then, do we wanna expose every single method? Not sure. Then the package size will increment but not sure if this is really a critical aspect. Worth to check.

I don't care 😄 All I care about is that we expose the necessary functionality.

jonaslagoni: If we want to minimize the size we should split the exposed libraries into individual packages instead of bundling everything.

smoya: Fair enough. +1 in my side to go in that direction. Would then you move the `browser_new` content  stuff to the `browser` dir right?

smoya: ping @jonaslagoni

jonaslagoni: Soo, we just have a single folder called `browser` and make a breaking change? 🤔

smoya: Why breaking change? 🤔

jonaslagoni: Because you will no longer be able to use the same import statement i.e. `import Parser from '@asyncapi/parser/browser';` will be no more. 

Instead you have to do `import {Parser} from '@asyncapi/parser/browser';` if I recall correctly

smoya: Ping @jonaslagoni

jonaslagoni: Done

PR #948, opened 2024-02-16
fix: iterating between operations, channels and messages

jonaslagoni:
**Description**
Once before I implemented this functionality incorrectly. this PR fixes that operations, channels, and messages could not be iterated between each other. On top of that, the functionality did not work when references were local files, this has also been fixed. 

**Related issue(s)**
Raised on slack: https://asyncapi.slack.com/archives/CQVJXFNQL/p1707760039073239

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -55,8 +55,6 @@ export async function parse(parser: Parser, spectral: Spectral, asyncapi: Input,
     } else {
       loadedObj = asyncapi;
     }
-    // Apply unique ids before resolving references
-    applyUniqueIds(loadedObj);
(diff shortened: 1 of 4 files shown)


src/parse.ts
```diff
@@ -71,6 +69,9 @@ export async function parse(parser: Parser, spectral: Spectral, asyncapi: Input,
   
     // unfreeze the object - Spectral makes resolved document "freezed" 
     const validatedDoc = copy(validated as Record<string, any>);
+
+    // Apply unique ids which are used as part of iterating between channels <-> operations <-> messages
+    applyUniqueIds(validatedDoc);
```
smoya, line 74: Executed now after validation.What is the wrong thing in previous implementation?

smoya (commented):
LGTM! 🚀🌔

smoya (approved):

PR #969, opened 2024-03-23
feat: add all{components} functions

KhudaDad414:
**Description**
- this PR adds the functions that have been proposed in https://github.com/asyncapi/parser-api/issues/109:


src/models/asyncapi.ts
```diff
@@ -31,4 +38,15 @@ export interface AsyncAPIDocumentInterface extends BaseModel<v2.AsyncAPIObject |
...
   allMessages(): MessagesInterface;
   allSchemas(): SchemasInterface;
+  allSecuritySchemes(): SecuritySchemesInterface;
+  allServerVariables(): ServerVariablesInterface;
+  allParameters(): ChannelParametersInterface;
+  allCorrelationIds(): CorrelationIdsInterface;
+  allTags(): TagsInterface;
+  allOperationTraits(): OperationTraitsInterface;
+  allMessageTraits(): MessageTraitsInterface;
+  allServerBindings(): BindingsInterface;
```
smoya, line 48: If it doesn't have a clear use case, please drop it. Same with any other method if it is the case.

Context: https://github.com/asyncapi/parser-api/issues/109#issuecomment-2015374655

src/models/asyncapi.ts
```diff
@@ -31,4 +38,15 @@ export interface AsyncAPIDocumentInterface extends BaseModel<v2.AsyncAPIObject |
   allOperations(): OperationsInterface;
   allMessages(): MessagesInterface;
   allSchemas(): SchemasInterface;
+  allSecuritySchemes(): SecuritySchemesInterface;
+  allServerVariables(): ServerVariablesInterface;
+  allParameters(): ChannelParametersInterface;
+  allCorrelationIds(): CorrelationIdsInterface;
+  allTags(): TagsInterface;
```
smoya, line 45: You are missing `allReplies` https://github.com/asyncapi/parser-api/pull/111/files#diff-9eddf4dc51bf6a0125ca7fb094468ad284112270385c41cebce4f8f0a29620abR23

src/models/v3/asyncapi.ts
```diff
@@ -30,10 +32,172 @@ import type { SchemasInterface } from '../schemas';
...
+import type { ServerVariablesInterface } from 'models/server-variables';
+import type { TagsInterface } from 'models/tags';
+import type { Tag } from './tag';
+import type { TagInterface } from 'models/tag';
 
 import type { v3 } from '../../spec-types';
 
 export class AsyncAPIDocument extends BaseModel<v3.AsyncAPIObject> implements AsyncAPIDocumentInterface {
+  allSecuritySchemes(): SecuritySchemesInterface {
+    return this.securitySchemes();
```
smoya, line 49: I see two issues here:

1. If `allSecuritySchemes` does the same as `securitySchemes`, then this is redundant.
2. I believe `securitySchemes` method implementation is wrong because it is only returning Security Schemes from `components` but instead would need to return from, for example, Operations in the root document.

cc @jonaslagoni please double 👀 just in case I'm delirious.

src/models/asyncapi.ts
KhudaDad414: @smoya I'm a little confused about where to add the `allReplies` function. If I add it here, then v2 also needs to implement it. But v2 doesn't have replies. I see three possible approaches:

 1) Remove `allReplies` from `parser-api` since it can't be implemented for v2.
 2) Add it here and return an empty array for the v2 implementation.
 3) Add `allReplies` to v3 only.

 I would appreciate some insight on this.

src/models/v3/asyncapi.ts
KhudaDad414: > 1. If `allSecuritySchemes` does the same as `securitySchemes`, then this is redundant.

I agree. it has only been added for the sake of completeness.

> 2. I believe `securitySchemes` method [implementation](https://github.com/asyncapi/parser-js/blob/1ac0843347318c3095ae0872c55dd6a80dd44415/src/models/v3/asyncapi.ts#L260-L266) is wrong ATM because it is only returning Security Schemes from `components` but instead would need to return from, for example, Operations in the root document.

the problem is that when I try to access the security scheme of an specific `operation` or `operationTrait` it returns `SecurityRequirements[]` instead of `SecuritySchemes`. 

not sure if this is a bug or I am missing something.

KhudaDad414: @smoya pingy pongy.

Issue #996, opened 2024-05-24
[BUG] Parsing yields error due to correlation between a Channel and Operation


g-radam:
### Describe the bug.

The parser yields an error due to some correlation between a Channel `address` being null and its associated Operation not having `traits: []` defined. Either changing the address to be not-null, OR, adding a `traits: []` field to the operation will suppress the error.

Note that Operation Traits are NOT required by the spec, so being forced to add a `traits: []` field to an operation due to a channel address being null is quite erroneous. I have zero insight into where such an issue could be stemming from.

### Expected behavior

Traits can be omitted from operations without an error, regardless of if the associated channel has a null address.

### Screenshots

[Studio link to the error](https://studio.asyncapi.com/?base64=YXN5bmNhcGk6IDMuMC4wCmluZm86CiAgdGl0bGU6IE5leHVzIFNlcnZlciBBUEkKICB2ZXJzaW9uOiAxLjAuMAoKY2hhbm5lbHM6CiAgRXhjaGFuZ2U6CiAgICBhZGRyZXNzOiBudWxsCiAgICBtZXNzYWdlczoKICAgICAgRm9vRXZlbnQ6CiAgICAgICAgbmFtZTogRm9vRXZlbnQKICAgICAgICB0aXRsZTogVGVuYW50IENyZWF0ZWQKICAgICAgICBjb250ZW50VHlwZTogYXBwbGljYXRpb24vanNvbgogICAgICAgIHBheWxvYWQ6CiAgICAgICAgICB0eXBlOiBzdHJpbmcKCm9wZXJhdGlvbnM6CiAgc2VuZFRlbmFudENyZWF0ZWQ6CiAgICB0aXRsZTogU2VuZCB0ZW5hbnQgY3JlYXRlZCBldmVudCB0byBjbGllbnQKICAgIGFjdGlvbjogc2VuZAogICAgY2hhbm5lbDoKICAgICAgJHJlZjogJyMvY2hhbm5lbHMvRXhjaGFuZ2UnCiAgICAjIHRyYWl0OiBbXQogICAgbWVzc2FnZXM6CiAgICAgIC0gJHJlZjogJyMvY2hhbm5lbHMvRXhjaGFuZ2UvbWVzc2FnZXMvRm9vRXZlbnQn)

```
1:1	Error thrown during AsyncAPI document parsing.
Name: TypeError, message: Cannot read properties of null (reading 'traits'),
[...]

PR #999, opened 2024-06-03
fix: remove unexpected support for traits at Channel Object level

smoya:
**Description**

Fixes https://github.com/asyncapi/parser-js/issues/996

The current AsyncAPI spec (v3.0.0) support traits only at:

- [Message Object level](https://www.asyncapi.com/docs/reference/specification/v3.0.0#messageTraitObject)
- [Operation Object level](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationTraitObject)

Channels traits are not supported, however for some reason we added such support in our Parser code. 
This was causing a bug (linked above).

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/996

### DIFF
--- a/src/custom-operations/apply-traits.ts
+++ b/src/custom-operations/apply-traits.ts
@@ -51,11 +51,9 @@ function applyTraitsToObjectV2(value: Record<string, unknown>) {
 const v3TraitPaths = [
   // operations
   '$.operations.*',
-  '$.operations.*.channel.*',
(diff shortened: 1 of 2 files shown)


jonaslagoni (approved):

smoya:
We can't merge this PR until https://github.com/asyncapi/parser-js/issues/1001 gets solved. Working on it.

Issue #1028, opened 2024-06-19
Changeset Add Release Action using Changeset


ayushnau:
### Why do we need this improvement?

Currently, we are using Semantic Release for our versioning and package publishing. However, as we transition to a monorepo setup, we require the capability to push to multiple GitHub and npm registries. Semantic Release does not support this functionality out of the box, necessitating the integration of Changeset for our release process.

### How will this change help?

By incorporating Changeset as our release tool, we can streamline the management and publishing process across multiple npm and GitHub registries. This will enable us to handle monorepos more efficiently and maintain better control over our package distribution.

### How could it be implemented/designed?

Add GitHub Action File:

    Create a GitHub Action workflow file in your repository (e.g., .github/workflows/release.yml).
    Configure the workflow to trigger on relevant events (e.g., push, pull request, etc.).
    Define the steps to install dependencies, run tests, and execute Changeset for releasing.

Update package.json:

    Add the necessary Changeset dependencies to your package.json file.
    Include scripts for managing Changeset operations (e.g., release, bump, etc.).

### 🚧 Breaking changes

Yes

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

PR #1008, opened 2024-06-09
feat: introduce the turborepo to the parser.js

ayushnau:
Changes Added.

    Turborepo Added.
    Converted the parser repo to app, inside the turborepo.
    Added the required scripts of the turborepo to the root package.json file.

Related issue(s)

Related to: https://github.com/asyncapi/parser-js/issues/963

PR #1029, opened 2024-06-19
feat: add GitHub action for releasing via Changeset

ayushnau:
Changes Added.

    Added Release github Action using Changeset. 
Related issue(s)

Related to: https://github.com/asyncapi/parser-js/issues/1028


ayushnau:
@smoya please review and merge this pr.

package.json
```diff
@@ -38,7 +38,10 @@
     "generate:readme:toc": "markdown-toc -i \"README.md\"",
     "generate:assets": "npm run build && npm run generate:readme:toc",
     "bump:version": "npm --no-git-tag-version --allow-same-version version $VERSION",
-    "prepublishOnly": "npm run generate:assets"
+    "prepublishOnly": "npm run generate:assets",
+    "changeset": "changeset",
+    "version-packages": "changeset version",
+    "publish-packages": "turbo run build && changeset publish"
```
derberg, line 44: instead of adding those to package.json did you consider just using `npx` and package directly in release config?

in past with semantic-release we resigned from using npm commands if they were not needed during regular development. So semantic-release we used directly like `npx semantic-release@19.0.4` so we had less things to maintain, it was easier to reuse across repos, and dependabot was not bugging us with new versions bump

.github/workflows/release.yml
```diff
@@ -0,0 +1,137 @@
+# This action is centrally managed in https://github.com/asyncapi/.github/
+# Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in the above-mentioned repo
+
+# It does magic only if there is a package.json file in the root of the project
+name: Release - if Node project
```
derberg, line 5:
```suggestion
name: Release with changeset
```

can you please name file like `release-with-changeset.yml` as might be that this release will work also in our generator and then also other monorepos, so we will maintain it in `.github` and we will do it with `release-with-changeset.yml` name - so once you opt in in future for file from `.github`, it will just be overwritten

ayushnau: ok sure changed the name as required.

package.json
ayushnau: ok now added.

.github/workflows/release-with-changeset.yml
```diff
@@ -0,0 +1,141 @@
+# This action is centrally managed in https://github.com/asyncapi/.github/
```
smoya, line 1: This is not centrally managed yet. Please remove the comment. 

```suggestion
```

.github/workflows/release-with-changeset.yml
```diff
@@ -0,0 +1,141 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: "${{ steps.lockversion.outputs.version }}"
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Install dependencies
+        shell: bash
+        run: npm ci
+     - if: steps.packagejson.outputs.exists == 'true'
+        name: Install Changesets
+        run: npm install --save-dev @changesets/cli@2.26.1
```
derberg, line 111: is it still needed if in the end you call below `npx` ?

.github/workflows/release-with-changeset.yml
```diff
@@ -0,0 +1,141 @@
...
+
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Create Release Pull Request or Release
+        uses: changesets/action@v1
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Publish to any of NPM, Github, and Docker Hub
+        id: release
+        with:
+          publish: npx turbo run build && npx changeset publish
+          version: npx changeset version
```
derberg, line 121: if you do not specify the version, npx will always pick up latest - which in future will kick maintainers big time

.github/workflows/release-with-changeset.yml
```diff
@@ -0,0 +1,141 @@
+# This action is centrally managed in https://github.com/asyncapi/.github/
+# Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in the above-mentioned repo
+
+# It does magic only if there is a package.json file in the root of the project
+name: Release with changeset
```
smoya, line 5:
```suggestion
name: Release with Changesets
```

derberg (commented):
you can revert package-lock changes as well

.github/workflows/release-with-changeset.yml
```diff
@@ -0,0 +1,141 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      # The below lines are not enough to have release supported for these branches
+      # Make sure the configuration of the `semantic-release` package mentions these branches
```
smoya, line 12: We do not use `semantic-release` anymore so this comment should be removed.

```suggestion
```

.github/workflows/release-with-changeset.yml
```diff
@@ -0,0 +1,141 @@
...
+      - next-spec
+      - next-major
+      - next-major-spec
+      - beta
+      - alpha
+      - next
```
smoya, line 18: Do we need anything else to be configured (perhaps in changeset config file?) to allow those branches to be released as well? How does it work?

smoya:
I don't fully understand why we created this in a separate PR. The fact we move to changesets is purely linked to the monorepo needs. If something goes wrong, I expect to revert just one commit and not few.

.github/workflows/release-with-changeset.yml
smoya: I found out this in their docs: https://github.com/changesets/changesets/blob/main/docs/prereleases.md

It should be done manually it seems. I guess in the workflow, detecting we are not in master branch and run `npx changeset pre enter {tag}` then. More investigation is needed. 

We can work on it in a second iteration.






[12:08](https://asyncapi.slack.com/archives/D03CPF9EXQF/p1718878116286129)
Seems like npx changeset pre enter {tag}

.github/workflows/release.yml
ayushnau: ok changed to the changesets.

.github/workflows/release-with-changeset.yml
ayushnau: done.

.github/workflows/release-with-changeset.yml
ayushnau: ok sure will do that.

.github/workflows/release-with-changeset.yml
ayushnau: ok removed we should add this like we do for semantic release.

.github/workflows/release-with-changeset.yml
ayushnau: ok now added the version.

smoya:
@ayushnau I understand this PR can be closed as you moved the changes to https://github.com/asyncapi/parser-js/pull/1008 ?

Issue #1017, opened 2024-06-14
Create the rule `asyncapi3-server-variables` for the v3 core ruleset


smoya:
Part of https://github.com/asyncapi/parser-js/issues/1016 (included in Bounty).

Add the following rule for the v3 core ruleset:

Based on `asyncapi2-server-variables`: Server variables must be defined and there must be no redundant variables.

PR #1030, opened 2024-06-27
feat: added the rules for server variables in v3

ibishal:
**Description**

- Added the ruleset for server variable in v3

**Related issue(s)**
#1017 

image


ibishal:
cc @smoya

Issue #963, opened 2024-03-14
Integrate smoya/multi-parser-js into asyncapi/parser-js


smoya:
## Context

[@smoya/multi-parser](https://github.com/smoya/multi-parser-js) is a library that allows users to parse an AsyncAPI document and get an output model based on a desired [Parser-API](https://github.com/asyncapi/parser-api) version. Additionally, it allows to convert models from one [Parser-API](https://github.com/asyncapi/parser-api) version to another.

This tool is required by some AsyncAPI tools that use the Parser-JS and have to send the parsed document model to another library, which could be compatible with a different [Parser-API](https://github.com/asyncapi/parser-api) version than the one used in that particular version of the Parser-JS.
You have several examples of usage here:

- CLI, to send the right parsed model to Modelina when Modelina didn't have support for Parser-JS v3: https://github.com/asyncapi/cli/blob/master/src/commands/generate/models.ts#L186C29-L186C77
- Modelina, to ensure any AsyncAPI model input uses the right [Parser-API](https://github.com/asyncapi/parser-api) version, so it can accept any [Parser-API](https://github.com/asyncapi/parser-api) v1, v2, v3... model as input: https://github.com/asyncapi/modelina/blob/master/src/processors/AsyncAPIInputProcessor.ts#L68-L69
- Modelina, to instantiate a Parser compatible with a particular [Parser-API](https://github.com/asyncapi/parser-api) version: https://github.com/asyncapi/modelina/blob/master/src/processors/AsyncAPIInputProcessor.ts#L72

###  1. 👤 Ownership 
[@smoya/multi-parser](https://github.com/smoya/multi-parser-js) is a library hosted in a GH repository under my particular ownership (@smoya). Not a big issue per se, but I don't wanna be the only maintainer of such a tool. I thought on donating this to AsyncAPI org (TSC would need to vote first) but the reality is that **won't be needed**, see next point:

### 2. 🔃 Release is not in sync with Parser-JS
Every new Parser-JS release is not triggering an update on the [@smoya/multi-parser](https://github.com/smoya/multi-parser-js). I agree I could configure a bot like [Renovate](https://www.mend.io/renovate/) (Dependabot doesn't allow NPM package aliases, and this package makes use of them) to try update those.
But still we have two issue with that situation:

[...]

PR #1036, opened 2024-07-04
feat: add multiparser to the parserjs monorepo repository.

ayushnau:
Changes Added.
 
    Added the Multiparser to the turborepo. 
    Added the required scripts of the turborepo to the root package.json file.

Related issue(s)

Related to: https://github.com/asyncapi/parser-js/issues/963


smoya:
@ayushnau Please fix the conflicts 🙏

smoya:
We are in the last stages, and this Bounty Issue might get completed anywhere from several hours to several days.

smoya (changes requested):

.github/workflows/release-with-changesets.yml
```diff
@@ -113,7 +113,7 @@ jobs:
         with:
           version: npx -p @changesets/cli changeset version
           commit: version packages
-          title: "chore: version packages"
+          title: "feat: version packages"
```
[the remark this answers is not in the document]
smoya: In fact, I would say this fits better:

```suggestion
          title: "chore(release): version packages"
```

aeworxet:
@asyncapi/bounty_team

packages/multi-parser/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "bump:version": "npm --no-git-tag-version --allow-same-version version $VERSION",
+    "prepublishOnly": "npm run generate:assets"
+  },
+  "dependencies": {
+    "@asyncapi/avro-schema-parser": "^3.0.3",
+    "@asyncapi/openapi-schema-parser": "^3.0.4",
+    "@asyncapi/protobuf-schema-parser": "^3.0.0",
+    "@asyncapi/raml-dt-schema-parser": "^4.0.4",
+    "parserapiv1": "npm:@asyncapi/parser@^2.1.0",
+    "parserapiv2": "npm:@asyncapi/parser@3.0.0-next-major-spec.8"
```
smoya, line 46: Where is the `parserapiv3` ?

ayushnau: yes this is shifted to line number 65. as "@asyncapi/parser": "file:../parser"
which will take the latest version as a  @asyncapi/parser

AyushNautiyalDeveloper: Actually I was thinking as the parser is in the local now will not need to keep it as dependency. but now i am thinking though the test are passing successfully it should be in dependency only. anyway updated the added it back.

ayushnau: Actually I was thinking as the parser is in the local now will not need to keep it as dependency. but now i am thinking though the test are passing successfully it should be in dependency only. anyway updated the added it back.

ayushnau:
@smoya  can you merge this. now.

smoya:
> @smoya  can you merge this. now. 

LOL i hope it is not an order 😂. Will review and merge when i have a chance. I'm out on holidays, and sick at the same time.

packages/multi-parser/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "prepublishOnly": "npm run generate:assets"
+  },
+  "dependencies": {
+    "@asyncapi/avro-schema-parser": "^3.0.3",
+    "@asyncapi/openapi-schema-parser": "^3.0.4",
+    "@asyncapi/protobuf-schema-parser": "^3.0.0",
+    "@asyncapi/raml-dt-schema-parser": "^4.0.4",
+    "parserapiv1": "npm:@asyncapi/parser@^2.1.0",
+    "parserapiv2": "npm:@asyncapi/parser@3.0.0-next-major-spec.8",
+    "@asyncapi/parser": "file:../parser"
```
smoya, line 48: Is there any reason for changing the dependency name? If there isnt, please do not modify the original code from https://github.com/smoya/multi-parser-js/blob/19750c914386d4417c9e17b13584634132805e27/package.json#L47

smoya (changes requested):
First review round

packages/multi-parser/package.json
ayushnau: Previously, we used parserv3, but now that the repositories are merged and multiparser is using the latest version, it makes sense to refer to it by the correct name. If anyone needs to know the version, they can simply check the version of parser.

smoya: The reason the version was explicit is exactly because we wanted it to be explicit for each version

ayushnau:
> How would new releases of the `@asyncapi/parser` affect this multi-parser package? How shall we do releases now if we want to sync the dependency of the `@asyncapi/parser` in `@asyncapi/multi-parser` 

1. Now whenever @asyncapi/parser  will be updated. New md file will be created in the .changeset folder staging the package and the level of change. 
2.  Now as the @asyncapi/multiparser is the directly using the local @asyncapi/parser. The dependency for the multiparser will be updated. 
3. Now while creating the md I am suggesting to add the @asyncapi/multiparser along with the parser as the dependency for that file is also changed. both package will be release together. 
 ```js
---
"@asyncapi/parser": minor
"@asyncapi/multi-parser": patch


changes.
```

**Comment by smoya** [comment]
> > How would new releases of the `@asyncapi/parser` affect this multi-parser package? How shall we do releases now if we want to sync the dependency of the `@asyncapi/parser` in `@asyncapi/multi-parser`
> 
> 1. Now whenever @asyncapi/parser  will be updated. New md file will be created in the .changeset folder stating the package and the level of change.
> 2. Now as the @asyncapi/multiparser is the directly using the local @asyncapi/parser. The dependency for the multiparser will be updated.
> 3. Now while creating the md I am suggesting to add the @asyncapi/multiparser along with the parser, as the dependency for multiparser is also changed. both package will be release together.
> 
> ```js
> ---
> "@asyncapi/parser": minor
> "@asyncapi/multi-parser": patch
> 
> 
> changes.
> ```
> 
> yes that would mean we dont need to do anything else to sync the parser with multiparser it will be automatically synced
> 
> @smoya

I guess it works by now. We could work on some automation in the future for this. 👍

File: turbo.json [hunk 4]
```diff
@@ -1,12 +1,16 @@
...
-    "test": { "cache": false },
-    "test:unit": {},
-    "test:integration": {},
-    "test:browser": {},
-    "lint": {},
-    "lint:fix": {},
-    "build": {}
+   "test": {"cache": false},
+   "prepublishOnly": {"cache": false},
+    "test:unit": {"cache": false},
```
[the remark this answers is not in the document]
smoya: ping

packages/multi-parser/package.json
smoya: ping @ayushnau

ayushnau:
will be giving update on this by on weekend. currently cant take it.

turbo.json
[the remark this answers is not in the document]
ayushnau: added the correct indentation.

packages/multi-parser/package.json
ayushnau: ok fixed this.

ayushnau:
@smoya updated this please check.

.changeset/wise-laws-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@asyncapi/multi-parser": minor
```
smoya, line 2: Remember we also need to release the `@asyncapi/parser` package as previous release failed.

BTW, I'm really concerned about the scenario where releasing a new version of the `@asyncapi/parser` and the fact the `@asyncapi/multi-parser` package will need to be released as well but with an up-to-date `@asyncapi/parser` dependency. 

You mentioned:

> yes that would mean we dont need to do anything else to sync the parser with multiparser it will be automatically synced
> Source: https://github.com/asyncapi/parser-js/pull/1036#issuecomment-2225374094

Because the release process should need to follow an strict order in this case afaik:

1. First, release the new `@asyncapi/parser` version, so the package is out with a new tag.
2. To update the dependency in the `package.json` in the `@asyncapi/multi-parser`

How that will be done automatically?

turbo.json
```diff
@@ -1,12 +1,16 @@
...
+   "test": {"cache": false},
+   "prepublishOnly": {"cache": false},
+   "test:unit": {"cache": false},
+   "test:integration": {"cache": false},
+   "test:browser": {"cache": false},
+   "lint": {"cache": false},
+   "lint:fix": {"cache": false},
+   "build": {"cache": false , "dependsOn": ["@asyncapi/parser#build"]},
+   "@asyncapi/parser#build":{"cache": false},  
+   "multi-parser:build":{"cache":false, "dependsOn": ["@asyncapi/parser#build"]}
```
smoya, line 13: Stupid question: shouldnt this be named `@asyncapi/multi-parser` as well?

ayushnau: no, the first @asyncapi/parser#build is referring to the name of the package. but the second multiparser is referring to the command in the script.

.changeset/wise-laws-retire.md
ayushnau: in the ci the new tag will be generated for the @asyncapi/parser and also for the @asyncapi/multiparser

smoya: I undesrtand. But to what version of the `@asyncapi/parser` will the `@asyncapi/multi-parser` package point to ?

ayushnau: ok let me clearfy with example. 

current version of @asyncapi/parser is 3.1.0
now as the multiparser is using the @asyncapi/parser like this 
    "parserapiv3": "file:../parser"

so the it will point to @asyncpi/parser 3.1.0

now let suppose we make some chagnes in @asyncapi/parser though we havent pushed the code and released it yet it will point to latest code. only. here we dont need to think about version for parser dependency in multiparser.

smoya: Ok, fair enough. Thanks for clarifying!

smoya: Still this is a valid point:

> Remember we also need to release the @asyncapi/parser package as previous release failed.

ayushnau: this is also done automatically by the changeset action it will check the version that is released in the npm and compare it with the latest version which we have if it is not released it will push the release.

smoya: Ok, I didn't know about that, I thought it was only based on changeset file. Thanks then.

smoya: It seems it didn't happened, isnt? https://github.com/asyncapi/parser-js/releases

ayushnau: it happened for multiparser correctly just not happen for parser because the previous pr should remain opened only.

Issue #936, opened 2023-12-27
Inconsistent `range` Values in Error Diagnostics for Different AsyncAPI File Formats


KhudaDad414:
#### Describe the Bug
When parsing an invalid AsyncAPI document, the parser consistently returns `range` values that seem to correspond to a JSON-formatted document, regardless of the actual file format (YAML or JSON). This inconsistency can lead to confusion, as the error locations might not accurately reflect the structure of the file being parsed.

#### Steps to Reproduce
1. Create two invalid AsyncAPI documents: one in YAML format and the other in JSON format. Ensure that the errors in both documents are structurally similar but occur at different lines or positions due to format differences.
2. Parse both files using the AsyncAPI parser.
3. Compare the `diagnostics.range` values returned for each file.

#### Expected Behavior
The parser should accurately reflect the error locations in the `diagnostics.range` values for both YAML and JSON formats. Since YAML and JSON have different structuring and syntax, errors in similar constructs should appear at different locations in the respective formats. Accurate error reporting is crucial for effective debugging and file validation.

smoya:
Another use case:

[...]

Issue #1012, opened 2024-06-11
[Bounty] A bunch of bugs that can be completed through the Bounty program


smoya:
### Describe the bug
The completion of **all** of the following bugs can be considered as completing a single Bounty Program issue of `medium` complexity level :

- [x] https://github.com/asyncapi/parser-js/issues/980
- [x] https://github.com/asyncapi/parser-js/issues/936

smoya:
> I would like to work on this issue within the Bounty Program.

A pleasure. Assigned to you @aeworxet

aeworxet:
### Bounty Issue's service comment

Text labels: `bounty/2024-Q3`, `bounty/medium`, `bounty/coding`
First assignment to third-party contributors: [2024-06-21 00:00:00 UTC+12:00](https://time.lol/#2024-06-21T00:00:00+12:00)
End Of Life after: [2024-07-31 23:59:59 UTC-12:00](https://time.lol/#2024-07-31T23:59:59-12:00)

@asyncapi/bounty_team

##### The Bounty Program is not a Mentorship Program. The accepted level of Bounty Program Participants is **Middle**/**Senior**.
##### Third-party contributors should [coherently articulate](https://github.com/asyncapi/converter-js/issues/233#issuecomment-2119226365) how they are going to approach the resolution process when expressing a desire to work on this Bounty Issue.

aeworxet:
### Bounty Issue's Timeline
|Complexity Level|Assignment Date (by GitHub)|Start Date (by BP Rules)|End Date (by BP Rules)|Draft PR Submission|Final PR Merge Start|Final PR Merge End|
|----------------|---------------------------|------------------------|----------------------|-------------------|--------------------|------------------|
|Medium          |2024-06-17                 |[2024-07-01](https://time.lol/#2024-07-01T00:00:00+12:00)              |**[2024-08-11](https://time.lol/#2024-08-11T23:59:59-12:00)**        |[2024-07-14](https://time.lol/#2024-07-14T23:59:59-12:00)         |[2024-07-28](https://time.lol/#2024-07-28T23:59:59-12:00)         |**[2024-08-11](https://time.lol/#2024-08-11T23:59:59-12:00)**        |

##### Please note that the dates given represent deadlines, not specific dates, so if the goal is reached sooner, it's better.
[...]

PR #1044, opened 2024-07-30
fix: remove forceful normalization of YAML to JSON

aeworxet:
This PR removes forceful normalization of YAML files to JSON format to fix https://github.com/asyncapi/parser-js/issues/936, which in its turn lays the foundation for reporting correct `range` values in error diagnostics for YAML files in `Studio`.

Related to https://github.com/asyncapi/parser-js/issues/1012

### DIFF
--- a/.changeset/smooth-pumas-deliver.md
+++ b/.changeset/smooth-pumas-deliver.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/parser": patch
+---
+
+fix: remove forceful normalization of YAML to JSON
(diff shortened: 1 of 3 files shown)


aeworxet:
@asyncapi/bounty_team

smoya:
@aeworxet it would be worth adding a comment in https://github.com/asyncapi/parser-js/blob/master/packages/parser/src/validate.ts#L49 mentioning Spectral uses the https://github.com/stoplightio/yaml parser, and basically, JSON is a valid YAML doc, so no matters if you pass YAML or JSON, that YAML parser will be able to parse it.

Additionally, please add two new unit tests in for the validate method testing at least one string input for JSON and another for YAML.

aeworxet:
Updated PR.

aeworxet:
Credits for the pack of changes in `packages/parser/test/parse.spec.ts` go to `eslint --fix`; otherwise, Sonar complained.

smoya (approved):
LGTM! 🚀🌔

PR #1045, opened 2024-08-01
fix: updated the method of importing the parser

ayushnau:
Changes Added.
 
    1. updated the way the nunjucks filter was importing the generator. 

Related issue(s)

Related to: https://github.com/asyncapi/parser-js/issues/963


ayushnau:
@smoya 
Fixed this.

image
image

In your example the error is coming due to asyncapi validation other wise the multiparser package is working fine. 

I have used this package for testing. @ayushnautiyal706020/multi-parser
Also pushed this code here if you want to check https://github.com/ayushnau/asyncapi-parser-example

aeworxet:
@asyncapi/bounty_team

smoya:
I don't get why we need this change. Can you please elaborate on that?

smoya:
> In your example the error is coming due to asyncapi validation other wise the multiparser package is working fine.

Attaching two screenshots running a simple example both with `@smoya/multi-parser` package and then with `@asyncapi/multi-parser` one. See the error on the second.

#### Using `@smoya/multi-parser`
Code_0voe7Odx@2x

#### Using `@asyncapi/multi-parser`
Code_EsUr0mJD@2x

ayushnau:
> I don't get why we need this change. Can you please elaborate on that?

Before I was using **parserv3: file::../parser** to import the parser in multiparser. but doing this way was creating problem in production build. of the multiparser. It turns out that in turborepo for importing internal dependency we need to use 
**@asyncapi/parser : "*"** this only works we are using the name which is the package.json of the package. So we need to use @asyncapi/parser instead of parserv3

ayushnau:
> @ayushnau
> 
> > In your example the error is coming due to asyncapi validation other wise the multiparser package is working fine.
> 
> Attaching two screenshots running a simple example both with `@smoya/multi-parser` package and then with `@asyncapi/multi-parser` one. See the error on the second.
> #### Using `@smoya/multi-parser`
> 
> Code_0voe7Odx@2x
> #### Using `@asyncapi/multi-parser`
> 
> Code_EsUr0mJD@2x

yes I am aware of this. Merging this pr will fix this.

smoya (approved):

smoya:
The prefix title of the pr should be fix.

.changeset/bubble-beach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@asyncapi/multi-parser": patch
+---
+
+Adding the Way the generator is imported in Nunjucks filter.
```
smoya, line 5: Improve the message so it reflects what it fixes.

smoya (changes requested):

.changeset/bubble-beach.md
ayushnau: done.

smoya (approved):

Issue #1016, opened 2024-06-13
Spectral rules for AsyncAPI Spec v3


smoya:
### Context

Spectral rules are the rules that validate AsyncAPI documents. There are different types or levels:
- [Core ruleset](https://github.com/asyncapi/parser-js/blob/master/src/ruleset/ruleset.ts#L10). Basic but required validation. Apply to both v2 and v3 (with some exceptions).
- [Recommended ruleset](https://github.com/asyncapi/parser-js/blob/master/src/ruleset/ruleset.ts#L81). Just like good practices. Those won't create validation errors but warnings. Apply to only v2 documents. Not very clear if we want to add for v3 and forward.
- **Specific** rulesets per AsyncAPI Spec version. Contain particular rulesets that apply only to documents of a particular Spec version.:
  - Core rulesets: Rules that are important and a doc of this Spec version must fulfil.
  - Recommended rulesets: Best practices for this particular Spec version

In the case of **AsyncAPI Spec v3**, the rules that apply ATM are:

- [Core ruleset](https://github.com/asyncapi/parser-js/blob/master/src/ruleset/ruleset.ts#L10):
  - `asyncapi-is-asyncapi` - The input must be a document with a supported version of AsyncAPI.
  - `asyncapi-latest-version` - Checking if the AsyncAPI document is using the latest version.
  - `asyncapi-document-resolved` - Checking if the AsyncAPI document has valid resolved structure.
  - `asyncapi-document-unresolved` - Checking if the AsyncAPI document has valid unresolved structure.
  - `asyncapi-internal` - That rule is internal to extend Spectral functionality for Parser purposes.
- [V3 specific ruleset](https://github.com/asyncapi/parser-js/blob/master/src/ruleset/v3/ruleset.ts):
  - Core rulesets:
    - `asyncapi3-operation-messages-from-referred-channel` - Operation "messages" must be a subset of the messages defined in the channel referenced in this operation.
    - `asyncapi3-required-operation-channel-unambiguity` - The "channel" field of an operation under the root "operations" object must always reference a channel under the root "channels" object.
    - `asyncapi3-required-channel-servers-unambiguity` - The "servers" field of a channel under the root "channels" object must always reference a subset of the servers under the root "servers" object.

### Description

If you check how many specific v2 rules are, you will notice a lot compared to v3. This PR tries to fix that gap between Spec versions and adopt or adapt those rules from v2 that make sense in v3.

[...]

PR #1051, opened 2024-08-11
chore: create rule `asyncapi3-channel-no-query-nor-fragment` for v3 core ruleset

aeworxet:
This PR creates the rule `asyncapi3-channel-no-query-nor-fragment` for the v3 core ruleset.

Resolves https://github.com/asyncapi/parser-js/issues/1018


smoya:
This PR adds a new functionality, so it should be released as, at least, a minor.

It might seem it introduces a breaking change, and somehow, it does since some v3 docs may be invalid now. But nothing [the spec](https://www.asyncapi.com/docs/reference/specification/v3.0.0#channelObject)  doesn't mention, so i would release it as minor.

Anyway, pinging other maintainers @jonaslagoni @magicmatatjahu

aeworxet:
@smoya 
Should I use a minor version for this PR only and create a new minor version for each issue from the https://github.com/asyncapi/parser-js/issues/1016#issue-2350952426 list, or should I use one minor version to annotate all of them (create one PR with all of those rules?)

smoya:
> @smoya 
> Should I use a minor version for this PR only and create a new minor version for each issue from the https://github.com/asyncapi/parser-js/issues/1016#issue-2350952426 list, or should I use one minor version to annotate all of them (create one PR with all of those rules?)

It is up to you. But don't worry about releasing multiple minors in a small period of time. Its all good.

.changeset/strong-garlics-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@asyncapi/parser": minor
+---
+
+chore: create rule `asyncapi3-channel-no-query-nor-fragment` for v3 core ruleset
```
smoya, line 5: It is not a chore but a feature, right?
```suggestion
feat: create rule `asyncapi3-channel-no-query-nor-fragment` for v3 core ruleset
```

smoya (approved):
LGTM! 🚀🌔

Issue #1063, opened 2024-11-17
[BUG] incorrect x-parser-schema-id when having multiple files


DanielRamosAcosta:
### Describe the bug.

I am encountering an issue with my `asyncapi.yml` file, which references schemas located in another file. When I generate the models, the generator creates generic names like `AnonymousSchema_1`, `AnonymousSchema_2`, `AnonymousSchema_3`, and so on for each schema in the referenced file.

After investigating the problem, I discovered that it seems to stem from the parser. Specifically, the `x-parser-schema-id` field in the `AsyncAPIDocumentInterface` object is returning values like `<anonymous-schema-1>`.

### Expected behavior

I would expect the generated models to have the right name.

### Screenshots

![image](https://github.com/user-attachments/assets/0ab6dc35-6061-483c-acaf-5659cf9d08ab)
![image](https://github.com/user-attachments/assets/f01337bc-eb1d-4a11-a49e-8fba2040b045)

### How to Reproduce

I will leave a PR that reproduces the error with tests.
I can try fixing the error, but I need guidance here.

Here is the reproduction: https://github.com/asyncapi/parser-js/pull/1064

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

Yes I am willing to submit a PR!

PR #1064, opened 2024-11-17
fix: external referenced schema names

DanielRamosAcosta:
This PR just adds a test for reproducing the error in #1063 

I would like to fix it, but I'm not sure where is the error, I would need guidance here.


packages/parser/test/parser.spec.ts
```diff
@@ -18,4 +21,20 @@ describe('Parser class', function() {
...
+    const nestedSchemas = fs.readFileSync(source, 'utf8');
+
+    const { document } = await parser.parse(nestedSchemas, { source });
+
+    const channels = (document as AsyncAPIDocumentInterface).channels().all();
+    const onUserRegisteredChannel = channels[0]
+    const onUserLoggedChannel = channels[1]
+    const userRegisteredMessagePayload = onUserRegisteredChannel.messages().all()[0].payload()?.json() as any;
+    const userLoggedMessagePayload = onUserLoggedChannel.messages().all()[0].payload()?.json() as any;
+    expect(userRegisteredMessagePayload?.["x-parser-schema-id"]).toBe("UserRegistered");
```
DanielRamosAcosta, line 37: This fails. The actual value is `<anonymous-schema-1>`

PR #1070, opened 2024-12-03
fix: reply id not accessible

jonaslagoni:
**Description**
The reply.id is not part of meta but the underlying json variable.


jonaslagoni:
Reply does not contain an id

Issue #1020, opened 2024-06-14
Create the rule `asyncapi3-channel-servers` for the v3 core ruleset


smoya:
Part of https://github.com/asyncapi/parser-js/issues/1016 (included in Bounty).

Add the following rule for the v3 core ruleset:

Based on `asyncapi2-channel-servers`: Channel servers must be defined in the "servers" object

PR #1073, opened 2025-01-05
feat: create rule asyncapi3-channel-servers for v3 core ruleset

dishafaujdar:
This PR creates the rule asyncapi3-channel-servers for the v3 core ruleset.

Resolves https://github.com/asyncapi/parser-js/issues/1020

### DIFF
--- a/.changeset/rich-elephants-applaud.md
+++ b/.changeset/rich-elephants-applaud.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/parser": minor
+---
+
+feat: create the rule `asyncapi3-channel-servers` for the v3 rule core ruleset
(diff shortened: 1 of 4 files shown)


dishafaujdar:
The file packages/parser/src/ruleset/v3/functions/channelServers.ts was copied from packages/parser/src/ruleset/v2/functions/channelServers.ts so that it will be duplicated in any case. Should I add some exclusion for SonarCloud?

smoya:
> The file `packages/parser/src/ruleset/v3/functions/channelServers.ts` was copied from `packages/parser/src/ruleset/v2/functions/channelServers.ts`, so it will be duplicated in any case. Should I add some kind of exclusion for SonarCloud?

SonarCloud is properly complaining here about duplication of code. If the code is the same or almost the same, why not sharing code and reuse it in both rules?

PR #1092, opened 2025-04-07
feat!: separate file system operations into a separate file

kennethaasan:
**Description**

This MR tries to separate file system operations into a separate file(s). Please see https://github.com/kennethaasan/parser-js/blob/feat/separate-file-system-operations/docs/migrations/v3-to-v4.md and https://github.com/asyncapi/asyncapi-react/issues/956 why this is a problem.

Also, the `next-major` branch needs to be updated with `master` so that you can see the changes I made easier. 

**Related issue(s)**


kennethaasan:
Closing. See https://github.com/asyncapi/asyncapi-react/pull/1149#issuecomment-2783023460

PR #1094, opened 2025-05-19
chore: remove smoya from CODEOWNERS

smoya:
More at https://github.com/asyncapi/community/pull/1906


jonaslagoni (approved):

derberg:
@jonaslagoni @magicmatatjahu I think something is wrong with Parser and no PR will go through. Weirdly on ubuntu some tests are long running, and eventually canceled by GH. I tried rerun here, and this did not work, which means it is not temporary issue. Also other PRs are affected of course

Issue #1099, opened 2025-07-10
[BUG] Custom Parsers aren't triggered on payloads that are defined though `Operations.Channel`


liam-oldershaw-ctct:
### Describe the bug.

A custom parser isn't triggered on a root document like:
```
operations: 
  FooOperation:
    action: send
    channel:
       $ref: otherFile.yaml#/channels/BarChannel
```

And there is another file that defines the channel and message 
```
channels:
  BarChannel:
    address:
    messages:
      AMessage:
        $ref: '#/components/messages/AMessage'
components:
  messages:
    AMessage:
      payload:
        schemaFormat: ....
        schema: ......
```

In the parsed json that is handled by packages/parser/src/custom-operations/parse-schema.ts#86 (in the variable detailed.parsed) the path to the payload with the custom schema is
`$.operations.FooOperation.channel.messages.AMessage.payload`

which is not covered in packages/parser/src/custom-operations/parse-schema.ts#25

A work around to this is to also re-specify the channel in the root document eg 
````
channels:
  BarChannel:
    $ref: otherFile.yaml#/channels/BarChannel

operations: 
  FooOperation:
    action: send
    channel:
       $ref: #/channels/BarChannel
```` 
but this feels weird as the channel is already fully defined in the other file and I am looking to re-use channels for other definitions

### Expected behavior

I believe customSchemasPathsV3 should also include 
`$.operations.*.channel.messages.*.payload`

### Screenshots

NA

### How to Reproduce

Root AsyncApi file:

```
operations: 
  FooOperation:
    action: send
    channel:
       $ref: otherFile.yaml#/channels/BarChannel
```

Other referenced file:
```
channels:
  BarChannel:
    address:
    messages:
      AMessage:
        $ref: '#/components/messages/AMessage'
components:
  messages:
    AMessage:
      payload:
        schemaFormat: ....
        schema: ......
```

### 🖥️ Device Information [optional]

AsyncApi v3

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

Yes I am willing to submit a PR!

q404365631:
I'd like to work on this! I'll trace why custom parsers don't trigger on payload-defined schemas and submit a PR with a fix + test.

[...]

PR #1100, opened 2025-07-10
Update parse-schema.ts

liam-oldershaw-ctct:
**Description**

- Triggers the custom parser for Messages that are defined inside of the channel of an operation

**Related issue(s)**
Resolves: #1099


fmvilas (commented):
That's a great catch. Appreciate the fix 🙏 Would you mind adding tests to cover this so it won't happen again by mistake?

liam-oldershaw-ctct:
@fmvilas while writing the unit test I've noticed that the schema also doesn't get an anonymous id like the tests above.. So that might be another issue (not sure how important though)

PR #1102, opened 2025-07-29
refactor: remove `node-fetch` dependency

xeho91:
**Description**

- Removed `node-fetch` and `@types/node-fetch` dependencies
- Those dependencies were used only for the types, AFAIK.
  I'm unsure what the minimum Node.js version supported for this package is. I noticed in the package root `>=18`.
  It could create a debate whether this update is a _breaking_ change (requires major upgrade) or not.
  I personally don't think so, given that I solved only the typing issue.

**Related issue(s)**

Package `@types/node-fetch` has a _critical_ vurnerable dependency `form-data`: https://github.com/advisories/GHSA-fjxv-7rqg-78g4


xeho91:
> [!NOTE]
> There's a chance that `browserify-shims` dev dependency is no longer needed

GabDug:
Would love to see this re-opened :)

PR #1123, opened 2026-01-21
fix: validate channel address when parameters are defined

### LINKED ISSUE #875: Parser do not validate and throw error when `parameters` are provided but address is null

abhi-03-kh:
## Description
This PR adds a validation rule to ensure that if a Channel Object has `parameters` defined, the `address` field must not be null.

Fixes #875

## Changes
* **feat:** Added `channel-parameters-require-address` rule to `v3/ruleset.ts`.
* **test:** Added `channel-parameters.spec.ts` to verify the rule catches invalid documents.

## Verification
1. Run the new test:
   ```bash
   npm test -- test/ruleset/rules/v3/channel-parameters.spec.ts

### DIFF
--- a/.changeset/strict-icons-flash.md
+++ b/.changeset/strict-icons-flash.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/parser": patch
+---
+
+fix: validate channel address when parameters are defined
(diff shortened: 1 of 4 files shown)


abhi-03-kh:
@fmvilas @magicmatatjahu  ,  any feedback on this when you have time?

PR #1146, opened 2026-04-13
fix: improve browser/GraalVM compatibility of the parser bundle

### LINKED ISSUE #1131: Request: Browser-compatible bundle for Java/GraalVM integration
### Title
Request: Browser-compatible bundle for Java/GraalVM integration

### Description
I'm working on a Java wrapper for the AsyncAPI parser using GraalVM's JavaScript engine. The current @asyncapi/parser package (v3.6.0) cannot be bundled for use in GraalVM because it depends on Node.js-specific APIs (fetch, Headers, etc.) that aren't available in GraalVM's JavaScript environment.

### Current Situation
- The old bundle-es5.js only supports AsyncAPI 2.4.0
- The new @asyncapi/parser v3.x requires fetch API and browser globals
- When bundled with webpack, it fails with: `TypeError: Cannot set property 'Headers' of undefined`

### Request
Could you provide or help create a browser-compatible bundle that:
1. Supports both AsyncAPI v2.x and v3.x
2. Works in environments without fetch API (or includes polyfills)
3. Can run in GraalVM's JavaScript engine
4. Doesn't depend on Node.js-specific modules

### Use Case
Java applications that need to parse AsyncAPI documents without spawning Node.js processes or using native V8 bindings (which don't support macOS ARM).

### Alternative Solutions Considered
- Using Playwright/Puppeteer (too heavy)
- Using J2V8 (no macOS ARM support)
- Creating custom bundle with webpack (fails due to fetch dependency)

### Environment
- GraalVM JS engine via javax.script
- Target: Java 11+
- Platform: macOS ARM, Linux, Windows

Would appreciate guidance on:
1. How to properly bundle the parser for non-browser/non-Node environments
2. If there's an existing standalone bundle
3. If this use case is supported

armorbreak001:
Fixes #1131

## Problem

The browser bundle (`browser/index.js`) could not be used in GraalVM's JavaScript engine or other non-browser/non-Node environments because:
1. `node-fetch` was bundled inline, which tries to access Node.js globals that don't exist in GraalVM
2. The webpack `globalObject` fallback chain didn't include `globalThis`, causing `TypeError: Cannot set property 'X' of undefined`
3. When no fetch implementation was available, the error message was a cryptic module resolution failure

### 1. `webpack.config.js`
- **Externalize `node-fetch`**: Marked as external (`commonjs2 null`) so it's not bundled. In environments with native fetch (browsers, modern GraalVM), the global fetch API is used instead.
- **Improved `globalObject` fallback**: Added `globalThis` to the fallback chain (`self → globalThis → this`) for broader environment support.

### 2. `from.ts`
- **Better error handling**: When neither global fetch nor node-fetch is available, throw a clear error message explaining what's needed:
  > "No fetch implementation available. In Node.js, install 'node-fetch'. In browsers or GraalVM, ensure the global fetch API is available."

## How to verify

```bash
# Build the browser bundle
cd packages/parser && npm run build:browser

# The bundle should not include node-fetch internals
grep -c "node-fetch" browser/index.js  # should be 0 or minimal

# Test in Node.js (should work as before)
node -e "const p = require('./browser'); console.log(typeof p);"

# For GraalVM: load browser/index.js in GraalVM Context
# and call new AsyncAPIParser().parse('{ asyncapi: "3.0.0", ... }')
```

**Note**: For full GraalVM support, users still need to ensure a fetch polyfill is available if their GraalVM version doesn't include one natively. This change makes that integration straightforward by removing the hard dependency on `node-fetch`.

### DIFF
--- a/packages/multi-parser/package.json
+++ b/packages/multi-parser/package.json
@@ -47,6 +47,9 @@
     "parserapiv1": "npm:@asyncapi/parser@^2.1.0",
     "parserapiv2": "npm:@asyncapi/parser@3.0.0-next-major-spec.8"
   },
+  "overrides": {
+    "jsonpath-plus": "^10.0.7"
+  },
   "devDependencies": {
     "@jest/types": "^29.0.2",
     "@swc/core": "^1.2.248",
--- a/packages/parser/src/custom-operations/parse-schema.ts
+++ b/packages/parser/src/custom-operations/parse-schema.ts
@@ -31,8 +31,12 @@ const customSchemasPathsV3 = [
   // operations
   '$.operations.*.messages.*.payload',
   '$.operations.*.messages.*.headers',
+  '$.operations.*.channel.messages.*.payload',
+  '$.operations.*.channel.messages.*.headers',
   '$.components.operations.*.messages.*.payload',
   '$.components.operations.*.messages.*.headers',
+  '$.components.operations.*.channel.messages.*.payload',
+  '$.components.operations.*.channel.messages.*.headers',
   // messages
   '$.components.messages.*.payload',
   '$.components.messages.*.headers.*',
--- a/packages/parser/src/from.ts
+++ b/packages/parser/src/from.ts
@@ -53,8 +53,18 @@ async function getFetch(): Promise<typeof fetch> {
     return __fetchFn;
   }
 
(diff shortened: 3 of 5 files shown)


armorbreak001:
Closing to free up PR queue — will resubmit if still relevant.

PR #1138, opened 2026-03-25
chore: add new codeowner for the repo

AayushSaini101:
Most of the Maintainers are inactive in the parser. I am already maintaining the CLI project since for a very long period of time, would like to maintain in the future

PR #1147, opened 2026-04-18
fix(parser): add missing title and summary fields to v3 ServerObject

armorbreak001:
## Problem

The AsyncAPI **3.0.0 specification** defines optional `title` and `summary` fields on the [Server Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#serverObject), but these fields were missing from the TypeScript type definition in `packages/parser/src/spec-types/v3.ts`.

This means users cannot properly type server objects that include these standard fields.

## Solution

Add `title?: string` and `summary?: string` to the `ServerObject` interface in v3.ts, matching the official specification.

## Changes

- `packages/parser/src/spec-types/v3.ts`: Added 2 optional string fields to `ServerObject`
  - `title?: string;`
  - `summary?: string;`

Fixes #1138

### DIFF
--- a/packages/parser/src/spec-types/v3.ts
+++ b/packages/parser/src/spec-types/v3.ts
@@ -44,6 +44,8 @@ export interface ServerObject extends SpecificationExtensions {
   protocol: string;
   pathname?: string;
   protocolVersion?: string;
+  title?: string;
+  summary?: string;
   description?: string;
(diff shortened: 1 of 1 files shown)


armorbreak001:
Closing to free up PR queue — will resubmit if still relevant.

Issue #1167, opened 2026-05-16
[MICROGRANT] Set of issues for resolution under the Microgrant Program 2026-06


AayushSaini101:
[MICROGRANT] Set of issues for resolution under the MICROGRANT Program 2026-05

This is an Aggregated MICROGRANT Issue consisting of GitHub issues
https://github.com/asyncapi/parser-js/issues/877
https://github.com/asyncapi/parser-js/issues/1137
https://github.com/asyncapi/parser-js/issues/1067

aeworxet:
### Microgrant Issue's service comment

Text labels: `microgrant/2026-06`, `microgrant/medium`, `microgrant/coding`
First assignment to regular contributors: [2026-05-23 00:00:00 UTC+12:00](https://time.lol/#2026-05-23T00:00:00+12:00)
End Of Life after: [2026-06-30 23:59:59 UTC-12:00](https://time.lol/#2026-06-30T23:59:59-12:00)

@asyncapi/microgrant_team

##### The Microgrant Program is not a Mentorship Program. The accepted level of Microgrant Program Participants is **Middle**/**Senior**.
##### Regular contributors should explain in clear terms how they will approach the resolution process when expressing interest in working on this Microgrant Issue.

AayushSaini101:
@aeworxet I want to work on the set of issues thanks

aeworxet:
Following the [expression of interest in working on this Microgrant Issue](https://github.com/asyncapi/parser-js/issues/1167#issuecomment-4483905174), @AayushSaini101 (`githubID`: [60972989](https://api.github.com/user/60972989)) is an AsyncAPI Maintainer specified in the file https://github.com/asyncapi/community/blob/master/MAINTAINERS.yaml, so they fall under the first category in the [prioritization list](https://github.com/asyncapi/community/blob/master/docs/010-contribution-guidelines/microgrant-program.md?plain=1#L42).

aeworxet:
### Microgrant Issue's Timeline
|Complexity Level|Assignment Date (by GitHub)|Start Date (by MP Rules)|End Date (by MP Rules)|Draft PR Submission|Final PR Merge Start|Final PR Merge End|
|----------------|---------------------------|------------------------|----------------------|-------------------|--------------------|------------------|
[...]

Issue #878, opened 2023-10-12
Parser fails to throw an error when i provide an invalid asyncapi file


AceTheCreator:
#### Describe the bug
The Parser doesn't throw any error when I provide an invalid asyncapi file but instead returns `undefined`

#### How to Reproduce
Here as you can see Included the `test:hello` which makes the following asyncapi doc invalid
```
asyncapi: 3.0.0
test: hello
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    address: user/signedup
    messages:
      subscribe.message:
        $ref: '#/components/messages/UserSignedUp'
operations:
  user/signedup.subscribe:
    action: send
    channel:
      $ref: '#/channels/user~1signedup'
    messages:
      - $ref: '#/components/messages/UserSignedUp'
    reply:
      channel:
        $ref: '#/channels/user~1signedup'
      messages:
        - $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
```

#### Expected behavior
I expect the Parser to return an error not `undefined`

kouzitech:
I have fixed this issue in PR #1168

The root cause was that `createUncaghtDiagnostic()` in `src/utils.ts` returns an empty array `[]` for non-Error values (strings, null, undefined, etc.). This means when the parser encounters an invalid document that causes a non-Error exception, the error is silently swallowed and the parser returns `undefined` instead of meaningful diagnostics.

The fix handles non-Error values by converting them to string diagnostic messages, ensuring the parser always returns diagnostic information.

I am participating in the MICROGRANT Program 2026-06 #1167

Guciolek:
Hi! I ran into this exact issue while integrating the parser into another project. I've opened a PR (#1182) that adds an opt-in `throwOnError` option to `Parser#parse()`. When enabled, the parser throws a new `AsyncAPIParseError` carrying the diagnostics instead of silently returning `{ document: undefined, diagnostics: [...] }`.

The default behavior is preserved (`throwOnError: false`), so this is fully backward compatible. The fix is small (~40 lines in `parse.ts` + 6 new tests, 2472 total pass). Happy to iterate on the design if the maintainers would prefer a different approach (e.g. flip the default and bump a major version, or expose it on `Parser` only and not on `parseOptions`).

PR #1168, opened 2026-05-17
fix: handle message bindings with $ref and return diagnostics for non-Error failures

### LINKED ISSUE #877: Parser returns `undefined` when there's bindings in message
#### Describe the bug

#### How to Reproduce

#### Expected behavior

cc @smoya

kouzitech:
Fixes #877 and #878

## Issue #877: Parser returns undefined when there's bindings in message

When an AsyncAPI v3 file has message bindings that include a `$ref`, the parser crashes or returns `undefined` because the `bindings()` function in `v3/mixins.ts` iterates over all entries of the bindings object, including the `$ref` key, treating it as a protocol name.

**Fix:** Filter out keys starting with `$` (like `$ref`) before processing binding entries.

## Issue #878: Parser fails to throw an error when provided invalid AsyncAPI file

When the parser encounters an invalid AsyncAPI document, internal errors that are not `Error` instances (e.g., strings, null, undefined) are silently swallowed by `createUncaghtDiagnostic()`, causing the parser to return `undefined` instead of meaningful diagnostics.

**Fix:** Handle non-Error values in `createUncaghtDiagnostic()` by converting them to string diagnostic messages, ensuring the parser always returns diagnostic information.

## Changes
- `packages/parser/src/models/v3/mixins.ts`: Filter out `$ref` entries in `bindings()` function
- `packages/parser/src/utils.ts`: Handle non-Error values in `createUncaghtDiagnostic()`

This PR is part of the [MICROGRANT Program 2026-06 #1167](https://github.com/asyncapi/parser-js/issues/1167).

### DIFF
--- a/packages/parser/src/models/v3/mixins.ts
+++ b/packages/parser/src/models/v3/mixins.ts
@@ -82,8 +82,12 @@ export abstract class CoreModel<J extends CoreObject = CoreObject, M extends Rec
 
 export function bindings(model: BaseModel<{ bindings?: BindingsObject }>): BindingsInterface {
   const bindings = model.json('bindings') || {};
+  // Filter out $ref entries (ReferenceObject) - they should not be treated as protocol bindings
+  // When bindings is a ReferenceObject like { $ref: '#/components/messageBindings/myBindings' },
+  // the $ref key should not be processed as a protocol name (fixes #877)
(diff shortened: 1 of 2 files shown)


princerajpoot20:
closing in reference to : https://github.com/asyncapi/parser-js/issues/877#issuecomment-4477611735

PR #1173, opened 2026-06-01
fix: filter out $ref keys in bindings() to prevent undefined parser output

### LINKED ISSUE #877: Parser returns `undefined` when there's bindings in message

q404365631:
## Description

Fixes #877

When AsyncAPI documents contain message bindings with `$ref` references (e.g., `{"$ref": "#/components/messageBindings/myBindings"}`), the parser returns `undefined` instead of properly parsing the document.

### Root Cause

The `bindings()` function in both `v2/mixins.ts` and `v3/mixins.ts` iterates over all entries of the bindings object using `Object.entries()`, including `$ref` keys. When a `$ref` key is present, it gets treated as a protocol name (like `kafka`, `amqp`, etc.), causing the `Binding` model to be created with invalid data, which ultimately causes the parser to crash or return `undefined`.

### Fix

Filter out keys starting with `$` (such as `$ref`) before processing binding entries, consistent with how the `extensions()` function filters keys using `EXTENSION_REGEX`.

### Changes

- `packages/parser/src/models/v3/mixins.ts`: Added `.filter(([key]) => !key.startsWith('$'))` before `.map()` in the `bindings()` function
- `packages/parser/src/models/v2/mixins.ts`: Same fix applied to the v2 `bindings()` function

### Testing

The fix ensures that:
1. Messages with `$ref` in bindings are parsed correctly instead of returning `undefined`
2. Normal protocol bindings (kafka, amqp, http, etc.) continue to work as expected
3. The `$ref` reference data is preserved in `originalData` for downstream processing

### Reproduction

Before fix:
```yaml
messages:
  myMessage:
    bindings:
      $ref: '#/components/messageBindings/myBindings'
    payload:
      type: object
```
Parser returns `undefined` for this message.

After fix:
Parser correctly creates the message model, with protocol bindings properly extracted (excluding the `$ref` key).

## MICROGRANT

This PR resolves issue #877 which is part of the [MICROGRANT Program 2026-06](https://github.com/asyncapi/parser-js/issues/1167).

### DIFF
--- a/.changeset/fix-bindings-ref-keys-877.md
+++ b/.changeset/fix-bindings-ref-keys-877.md
@@ -1,0 +1,17 @@
+---
+"@asyncapi/parser": patch
+---
+
+fix(parser): filter out `$ref` keys in `bindings()` to prevent undefined parser output
+
+When a v2 or v3 AsyncAPI message/server/channel/operation has a `bindings`
+field containing a `$ref` (e.g. `{"$ref": "#/components/messageBindings/kafka"}`),
+the parser would iterate over `bindings` and treat `$ref` as a protocol name,
+causing the `bindings()` method on the corresponding model to throw or
+return `undefined`, and the whole document parse to fail.
+
+This was reported in #877 (also part of the [MICROGRANT 2026-06 #1167](https://github.com/asyncapi/parser-js/issues/1167)
+aggregated issue). The fix filters out any key starting with `$` before
+processing binding entries.
+
+Fixes #877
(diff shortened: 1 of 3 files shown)


q404365631:
Hi! 👋 This PR fixes the issue where `$ref` keys in bindings cause the parser to return `undefined`. The fix filters out `$ref` keys before processing. Could someone please take a look? Thanks!

princerajpoot20:
the underline issue is part of microgrant program. please read the program guidlines https://github.com/asyncapi/community/blob/master/docs/010-contribution-guidelines/microgrant-program.md. thanks

PR #1174, opened 2026-06-03
feat(parser): add title() method to v3 Channel model

### LINKED ISSUE #1067: [FEATURE] Add ability to get title from the channel object
### Why do we need this improvement?

The [channel object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#channelAddressExpressions) contains a `title` (optional) field. I can't see anyway to get this title value using the parser?

Be nice to support this
```
const channels = document.allChannels();

for(const channel of channels){
       // The channel ID
        console.log('channel', channel.id())
       
       // does not work?
        console.log('title', channel.title())

      }
```

Maybe we can have `title` and `hasTitle` options?

### How will this change help?

Can help us write scripts to get the channel title from the Spec file.

### Screenshots

_No response_

### How could it be implemented/designed?

Adding `hasTitle` and `title` fields into the channel.

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

Yes I am willing to submit a PR!

q404365631:
## What
Adds a `title()` method to the v3 `Channel` model so consumers can read the `title` field from a channel object.

## Why
The AsyncAPI 3.0 spec defines `title` as an optional field on `ChannelObject` (and the parser already has it on the TypeScript type), but the `Channel` model had no accessor. Closing this gap so the parser exposes all the fields the spec allows.

Fixes #1067

## Changes
- **`packages/parser/src/models/channel.ts`**: Added optional `title?(): string | undefined` to the `ChannelInterface` (so v3 implements it, v2 doesn't have to).
- **`packages/parser/src/models/v3/channel.ts`**: Implemented `title()` returning `this._json.title`.
- **`packages/parser/test/models/v3/channel.spec.ts`**: Two tests — one verifies the value, one verifies `undefined` when unset.

3 files changed, 19 insertions. Follows the existing minimal pattern used by `address()`.

### DIFF
--- a/.changeset/feat-channel-title-1067.md
+++ b/.changeset/feat-channel-title-1067.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/parser": minor
+---
+
+feat(parser): add `title()` method to v3 Channel model so users can read the channel's title field from the parser API, matching the corresponding v3 spec field.
--- a/packages/parser/src/models/channel.ts
+++ b/packages/parser/src/models/channel.ts
@@ -8,6 +8,7 @@ import type { ServersInterface } from './servers';
 export interface ChannelInterface extends BaseModel, BindingsMixinInterface, DescriptionMixinInterface, ExtensionsMixinInterface {
   id(): string;
   address(): string | null | undefined;
+  title?(): string | undefined;
   servers(): ServersInterface;
   operations(): OperationsInterface;
   messages(): MessagesInterface;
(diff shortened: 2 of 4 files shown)


princerajpoot20:
the underline issue is part of microgrant program. please read the program guidlines https://github.com/asyncapi/community/blob/master/docs/010-contribution-guidelines/microgrant-program.md. thanks

PR #1196, opened 2026-07-07
fix: filter out $ref keys in bindings() to prevent undefined parser error

### LINKED ISSUE #877: Parser returns `undefined` when there's bindings in message

AayushSaini101:
Fixes #877

Related: #1167 

When bindings objects still contain JSON reference keys, treat them as references rather than protocol names so Binding models are not built from invalid data.

### DIFF
--- a/packages/parser/src/schema-parser/index.ts
+++ b/packages/parser/src/schema-parser/index.ts
@@ -45,10 +45,10 @@ export async function validateSchema(parser: Parser, input: ValidateSchemaInput)
   return schemaParser.validate(input);
 }
 
-export async function parseSchema(parser: Parser, input: ParseSchemaInput) {
+export async function parseSchema(parser: Parser, input: ParseSchemaInput): Promise<AsyncAPISchema> {
   const schemaParser = parser.parserRegistry.get(input.schemaFormat);
   if (schemaParser === undefined) {
-    throw new Error('Unknown schema format');
+    return input.data as AsyncAPISchema;
   }
   return schemaParser.parse(input);
 }
--- a/.changeset/fix-bindings-ref-filter.md
+++ b/.changeset/fix-bindings-ref-filter.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/parser": patch
+---
+
+fix: filter out $ref keys in bindings() to prevent undefined parser output
--- a/packages/parser/src/models/v2/components.ts
+++ b/packages/parser/src/models/v2/components.ts
@@ -136,8 +136,8 @@ export class Components extends BaseModel<v2.ComponentsObject> implements Compon
         Object.entries(bindingsData)
           .filter(([key]) => !key.startsWith('$'))
           .map(([protocol, binding]) => 
-          this.createModel(Binding, binding, { protocol, pointer: `${pointer}/${protocol}` })
-        ),
+            this.createModel(Binding, binding, { protocol, pointer: `${pointer}/${protocol}` })
+          ),
         { originalData: bindingsData as any, asyncapi, pointer }
       );
       return bindings;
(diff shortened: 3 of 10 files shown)


aeworxet:
@asyncapi/microgrant_team

princerajpoot20 (commented):
I tried the adeo file against this PR branch and still get document: undefined with the same schema-format error.

Can you try it on your end with this file and share the findings: https://github.com/asyncapi/spec/blob/master/examples/adeo-kafka-request-reply-asyncapi.yml

AayushSaini101:
Done @princerajpoot20 

image

**Tested via this script:**

```
import fs from 'node:fs';
import path from 'node:path';

import { AsyncAPIDocumentV3 } from '../../src/models';
import { Parser } from '../../src/parser';
import { filterLastVersionDiagnostics } from '../utils';

describe('Adeo case study spec', function() {
  const parser = new Parser();
  const specPath = path.join(__dirname, '../mocks/adeo-case-study.yaml');
  const spec = fs.readFileSync(specPath, 'utf8');

  it('should parse the Adeo case study document without registering AvroSchemaParser', async function() {
    const { document, diagnostics } = await parser.parse(spec, { source: specPath });

    expect(document).toBeInstanceOf(AsyncAPIDocumentV3);
    expect(document).toBeDefined();
    expect(filterLastVersionDiagnostics(diagnostics).length === 0).toEqual(true);

    const channel = document!.allChannels().get('costingRequestChannel');
    const responseChannel = document!.allChannels().get('costingResponseChannel');
    const operation = document!.allOperations().get('receiveACostingRequest');

    expect(document!.allServers()).toHaveLength(3);
    expect(document!.allServers().all().map(server => server.id())).toEqual([
      'production',
      'staging',
      'dev',
    ]);

    expect(channel?.address()).toEqual('adeo-{env}-case-study-COSTING-REQUEST-{version}');
    expect(channel?.bindings().get('kafka')?.json()).toEqual({
      replicas: 3,
      partitions: 3,
      topicConfiguration: {
        'cleanup.policy': ['delete'],
        'retention.ms': 60000000,
      },
    });

    expect(responseChannel?.address()).toBeNull();
    expect(responseChannel?.bindings().get('kafka')?.json()?.['x-key.subject.name.strategy']?.type).toEqual('string');

    expect(operation?.summary()).toContain('[COSTING]');
    expect(operation?.bindings().get('kafka')?.json()?.groupId?.type).toEqual('string');

    const requestMessage = document!.allMessages().find(msg => msg.name() === 'CostingRequestV1');
    expect(requestMessage?.title()).toEqual('Costing Request V1');
    expect(requestMessage?.payload()?.schemaFormat()).toEqual('application/vnd.apache.avro;version=1.9.0');
  });
});

```

packages/parser/src/schema-parser/index.ts
```diff
@@ -45,10 +45,10 @@ export async function validateSchema(parser: Parser, input: ValidateSchemaInput)
   return schemaParser.validate(input);
 }
 
-export async function parseSchema(parser: Parser, input: ParseSchemaInput) {
+export async function parseSchema(parser: Parser, input: ParseSchemaInput): Promise<AsyncAPISchema> {
   const schemaParser = parser.parserRegistry.get(input.schemaFormat);
   if (schemaParser === undefined) {
-    throw new Error('Unknown schema format');
```
AayushSaini101, line 51: this is the reason it was failing for avro schema parsing, if there is any spec containing avro schema parsing, we need to register manually otherwise it will raise an error cc: @princerajpoot20

princerajpoot20 (approved):
LGTM!

PR #1198, opened 2026-07-12
fix: validate v3 channel parameters against address placeholders

### LINKED ISSUE #875: Parser do not validate and throw error when `parameters` are provided but address is null

AayushSaini101:
closes: #875

### DIFF
--- a/packages/parser/src/ruleset/v3/functions/channelParameters.ts
+++ b/packages/parser/src/ruleset/v3/functions/channelParameters.ts
@@ -25,7 +25,7 @@ export const v3ChannelParameters = createRulesetFunction<
   },
   (targetVal, _, ctx) => {
     const { address, parameters } = targetVal;
-    if (!parameters || Object.keys(parameters).length === 0) return;
+    if (!parameters) return;
 
     const variables = parseUrlVariables(address ?? '');
     const results: IFunctionResult[] = [];
--- a/packages/parser/src/ruleset/v3/functions/channelParameters.ts
+++ b/packages/parser/src/ruleset/v3/functions/channelParameters.ts
@@ -1,0 +1,55 @@
+import { createRulesetFunction } from '@stoplight/spectral-core';
+
+import { getMissingProps, getRedundantProps, parseUrlVariables } from '../../utils';
+
+import type { IFunctionResult } from '@stoplight/spectral-core';
+
+export const v3ChannelParameters = createRulesetFunction<
+  { address?: string | null; parameters?: Record<string, unknown> },
+  null
+>(
+  {
+    input: {
+      type: 'object',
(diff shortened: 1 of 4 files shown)


packages/parser/src/ruleset/v3/ruleset.ts
```diff
@@ -82,5 +83,18 @@ export const v3CoreRuleset = {
...
+    'asyncapi3-channel-parameters': {
+      description: 'Channel parameters must be defined and there must be no redundant parameters.',
+      message: '{{error}}',
+      severity: 'error',
+      recommended: true,
+      given: [
+        '$.channels.*',
+        '$.components.channels.*',
+      ],
+      then: { // NOSONAR
```
princerajpoot20: What’s the NOSONAR for?

packages/parser/src/ruleset/v3/functions/channelParameters.ts
```diff
@@ -0,0 +1,55 @@
...
+          type: 'object',
+        },
+      },
+      required: ['parameters'],
+    },
+    options: null,
+  },
+  (targetVal, _, ctx) => {
+    const { address, parameters } = targetVal;
+    if (!parameters || Object.keys(parameters).length === 0) return;
```
princerajpoot20: early returning on `Object.keys(parameters).length === 0` 
means address: 'users/{userId}/signedUp' + parameters: {} 

never reports missing userId. 
how about removing that early return so empty parameters still runs getMissingProps?

princerajpoot20 (commented):
just few minor comments

aeworxet:
@asyncapi/microgrant_team

packages/parser/src/ruleset/v3/functions/channelParameters.ts
AayushSaini101: make sense done thanks

packages/parser/src/ruleset/v3/ruleset.ts
AayushSaini101: image 
this is the reason //No Sonar is also present for 'asyncapi3-channel-servers' ruleset as well

princerajpoot20 (approved):

PR #1217, opened 2026-08-06
fix: support native array methods on collections

### LINKED ISSUE #874: javascript map() function doesn't work on SecurityRequirements object.
#### Describe the bug
SecurityRequirements object acts as a javascript array in some ways but not in all of the cases. 
consider this example:
```ts
import Parser from "@asyncapi/parser"
const file = await Bun.file("asyncapi.yaml").text()

const parser = new Parser()

const { document, diagnostics } = await parser.parse(file)
const securityRequirements = document?.allServers()[0].security()[0]

if (!securityRequirements) {
  process.exit(0)
}
// Works fine
for (const security of securityRequirements) {
  console.log("In a for loop: ", security.jsonPath())
}

//Works fine
securityRequirements?.forEach((securityRequirement) => {
  console.log("in a forEach: ", securityRequirement.jsonPath())
})

//Throws error
securityRequirements?.map((securityRequirement) => {
  console.log(securityRequirement.jsonPath())
  return securityRequirement.jsonPath()
})
```

the last case will Throw:
```bash
1 | "use strict";
2 | Object.defineProperty(exports, "__esModule", { value: true });
3 | exports.Collection = void 0;
4 | class Collection extends Array {
5 |     constructor(collections, _meta = {}) {
6 |         super(...collections);
            ^
TypeError: Spread syntax requires ...iterable[Symbol.iterator] to be a function
      at new Collection (node_modules/@asyncapi/parser/cjs/models/collection.js:6:8)
      at new SecurityRequirements (:1:32)
      at map (:1:20)
```

AayushSaini101:
closes: #874 

## Summary

- Fix native Array methods (`.map()`, `.filter()`, `.slice()`) on `Collection` subclasses such as `SecurityRequirements`
- Root cause: Array transforms construct results via the subclass constructor with a length number; `Collection` then does `super(...collections)` and throws
- Fix: set `Symbol.species` to `Array` so transforms return plain arrays (same pattern as `all()` / `filterBy()`)

## Problem

```ts
securityRequirements.forEach(...) // works
securityRequirements.map(...)     // throws
// TypeError: Spread syntax requires ...iterable[Symbol.iterator] to be a function
```

Workaround today: `securityRequirements.all().map(...)`.

This affects every `Collection` subclass (`Servers`, `Channels`, `Messages`, `SecurityRequirements`, …), not only security.

## Solution

In `packages/parser/src/models/collection.ts`:

```ts
static get [Symbol.species]() {
  return Array;
}
```

Native transforms no longer reconstruct a `Collection` with an invalid constructor signature.

## Test plan

- [x] `collection.spec.ts` — `.map()` / `.filter()` / `.slice()` return plain arrays; source/`meta()` unchanged
- [x] `security-requirements.spec.ts` — `.map()` does not throw (issue reproduction)
- [x] Existing Collection helpers (`all`, `filterBy`, `get`, `has`) still pass
- [ ] CI green on PR

## Changeset

Patch for `@asyncapi/parser`: `fix: support native array methods on collections`

## References

- Closes https://github.com/asyncapi/parser-js/issues/874
- Related: https://github.com/asyncapi/parser-js/pull/1214


princerajpoot20 (approved):

aeworxet:
@asyncapi/microgrant_team

PR #1218, opened 2026-08-06
fix: allow unknown custom schema formats without validation errors

### LINKED ISSUE #1066: [BUG] Any non supported schema format should not make the doc invalid
### Describe the bug.

Current supported schema formats are available in https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject. 

If we try to parse a document containing an schema with, for example, `schemaFormat: "application/octet-stream"`, the parser gives a validation error: `Unknown schema format`.

E.g. 

```
Error thrown during AsyncAPI document parsing. Name: Error, message: Unknown schema format, stack: Error: Unknown schema format\n'
```

However, we find the following statement in the spec (same link as above):

> Custom values are allowed but their implementation is OPTIONAL.

### Expected behavior

Even though no custom schema parser is found for a given schema format, the document should still be valid as long as it is a string

### Screenshots

CleanShot 2024-11-21 at 17 19 36@2x

### How to Reproduce

```js
import { Parser } from "@asyncapi/parser";
const parser = new Parser();
const { document, diagnostics } = await parser.parse(`
asyncapi: 3.0.0
info:
  title: Test custom schema format
  version: 0.1.0
components:
  schemas:
    theSchema:
      schemaFormat: application/octet-stream # This is allowed by the spec
      schema: true
`);

if (document) {
  console.log(document.info().title());
} else {
  console.log(diagnostics);
}
```

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

None

AayushSaini101:
That rejected valid documents that use custom schema formats without a registered schema parser.

## Changes

- Skip schema validation when no registered parser matches the given `schemaFormat`
- Return the schema as-is during parse for unknown custom formats
- Require `schemaFormat` to be a string; reject non-string values
- Add tests for custom formats such as `application/octet-stream` and Avro without a registered parser

## Related issue

Closes #1066

## Test plan

- [ ] Parse a v3 document with `schemaFormat: "application/octet-stream"` and confirm it succeeds
- [ ] Parse a document with an unknown string `schemaFormat` and confirm no `Unknown schema format` diagnostic
- [ ] Confirm registered formats (e.g. AsyncAPI JSON Schema) still validate as before
- [ ] Confirm non-string `schemaFormat` is rejected


packages/parser/src/schema-parser/index.ts
```diff
@@ -27,27 +27,31 @@ export interface SchemaParser<D = unknown, M = unknown> {
...
+      const path = [...input.path];
+      path.pop(); // remove last element (e.g. 'payload' / 'schema')
+      return [
+        {
+          message: 'Schema format must be a string',
+          path: [...path, 'schemaFormat'],
+        },
+      ] as SchemaValidateResult[];
+    }
+    return [];
```
princerajpoot20: how about if  we emit an info/warning diagnostic like: "No schema parser registered for..."

packages/parser/src/schema-parser/index.ts
```diff
@@ -27,27 +27,31 @@ export interface SchemaParser<D = unknown, M = unknown> {
...
   }
   return schemaParser.validate(input);
 }
 
 export async function parseSchema(parser: Parser, input: ParseSchemaInput): Promise<AsyncAPISchema> {
   const schemaParser = parser.parserRegistry.get(input.schemaFormat);
+  // Instead of throwing "Unknown schema format", return the schema as-is when no parser is registered.
```
princerajpoot20:
```suggestion
  // No registered parser for this format — return the schema unchanged (validation is skipped separately).
```

packages/parser/test/custom-operations/parse-schema-v3.spec.ts
```diff
@@ -131,13 +131,41 @@ describe('custom operations for v3 - parse schemas', function() {
...
-    expect(diagnostics.length > 0).toEqual(true);
+    expect(diagnostics.some(d => d.message?.includes('Unknown schema format'))).toEqual(false);
     expect(((document?.json()?.channels?.channel as v3.ChannelObject).messages?.message as v3.MessageObject)?.payload?.schema).toEqual({
       type: 'object',
       'x-parser-schema-id': '<anonymous-schema-1>',
     });
   });
 
+  it('should allow custom schemaFormat application/octet-stream without invalidating the document', async function() {
+    const documentRaw = {
```
princerajpoot20, line 142: can you add a case that matches the issue reproduction more closely i.e Multi Format Schema under components.schemas

packages/parser/src/schema-parser/index.ts
```diff
@@ -27,27 +27,31 @@ export interface SchemaParser<D = unknown, M = unknown> {
...
   }
   return schemaParser.validate(input);
 }
 
 export async function parseSchema(parser: Parser, input: ParseSchemaInput): Promise<AsyncAPISchema> {
   const schemaParser = parser.parserRegistry.get(input.schemaFormat);
+  // No registered parser for this format — return the schema unchanged (validation is skipped separately).
   if (schemaParser === undefined) {
+    if (typeof input.schemaFormat !== 'string') {
+      throw new Error('Schema format must be a string');
```
princerajpoot20: umm..i guess, these non-string guards are not reachable on the public parse/validate path.

Call sites go through getSchemaFormat(), which already does in line no: 79

aeworxet:
@asyncapi/microgrant_team

packages/parser/src/schema-parser/index.ts
AayushSaini101: thanks for highligthing done

packages/parser/test/custom-operations/parse-schema-v3.spec.ts
AayushSaini101: done

packages/parser/src/schema-parser/index.ts
AayushSaini101: Added warning and updated the flow thanks

princerajpoot20 (approved):
LGTM

PR #1226, opened 2026-08-17
fix: ship LICENSE, NOTICE and README in the published packages

justFU7:
## Description

The published packages of this monorepo are missing their legal files, and two of the three are missing
their README as well.

Each manifest asks for them:

```json
"files": ["/esm", "/cjs", "/browser", "LICENSE", "README.md"]
```

but those paths resolve relative to `packages/<pkg>/`, and `LICENSE`, `NOTICE` and (for two of the three
packages) `README.md` exist only at the repository root, so npm has nothing to match and packs neither.

### Evidence from the published artifacts

Listing the current tarballs, everything outside `esm/`, `cjs/` and `browser/`:

| package | version | files at the tarball root |
| --- | --- | --- |
| `@asyncapi/parser` | 3.6.3 | `package.json` |
| `@asyncapi/multi-parser` | 2.4.0 | `package.json` |
| `@asyncapi/openapi-schema-parser` | 3.1.0 | `package.json`, `README.md` |

```
$ curl -sLO https://registry.npmjs.org/@asyncapi/parser/-/parser-3.6.3.tgz
$ tar tzf parser-3.6.3.tgz | grep -vE "^package/(esm|cjs|browser)/"
package/package.json
```

So no package ships a `LICENSE` or a `NOTICE`, and only `openapi-schema-parser` ships a `README`.

### Why it matters

Apache-2.0 section 4(a) requires a copy of the licence to accompany each distribution of the work, and 4(d)
requires the `NOTICE` attributions to travel with it. The npm tarball is how essentially every consumer
receives these packages, and it currently carries neither. Separately, `@asyncapi/parser` and
`@asyncapi/multi-parser` render with no README on npmjs.com.

## Changes

- new `scripts/copy-legal.js`, copying the root `LICENSE` and `NOTICE` into the package being published, plus
  the root `README.md` **only when the package does not ship one of its own** (so
  `openapi-schema-parser`’s README is left alone);
- each package runs it from `prepublishOnly`, so the files exist by the time npm packs;
- `NOTICE` added to each `files` array next to the `LICENSE` entry that was already there;
- the copied files are gitignored per package.

No new dependency - `prepublishOnly` already existed in all three manifests, this only prepends one step.

### Verifying

```
npm -w @asyncapi/parser run copy:legal && npm pack -w @asyncapi/parser --dry-run
```

`LICENSE`, `NOTICE` and `README.md` now appear in the pack listing.

### DIFF
--- a/.changeset/ship-legal-files.md
+++ b/.changeset/ship-legal-files.md
@@ -1,0 +1,9 @@
+---
+"@asyncapi/parser": patch
+"@asyncapi/multi-parser": patch
+"@asyncapi/openapi-schema-parser": patch
+---
+
+Ship LICENSE, NOTICE and README.md inside the published tarballs. All three were already listed in each
+package's `files` array, but those entries resolve inside `packages/<pkg>/` where the files do not exist,
+so npm packed none of them.
--- a/.changeset/ship-legal-files.md
+++ b/.changeset/ship-legal-files.md
@@ -4,6 +4,6 @@
 "@asyncapi/openapi-schema-parser": patch
 ---
 
-Ship LICENSE, NOTICE and README.md inside the published tarballs. All three were already listed in each
-package's `files` array, but those entries resolve inside `packages/<pkg>/` where the files do not exist,
-so npm packed none of them.
+Ship the Apache-2.0 `LICENSE` and the `NOTICE` file inside the published tarballs, and add the repository
+README to the two packages that had none. `LICENSE` and `README.md` were already listed in each package's
+`files` array, but those entries resolve inside `packages/<pkg>/`, where neither file exists.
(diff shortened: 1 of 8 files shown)


justFU7:
A note on scope, in case it helps triage.

This touches three packages because all three are affected, but the fix is independent per package. If a
three-package diff is awkward to review, I am happy to close this and open one PR per package instead - just say
which you prefer.

The evidence is reproducible in two commands per package:

```console
$ curl -sLO https://registry.npmjs.org/@asyncapi/parser/-/parser-3.6.3.tgz
$ tar tzf parser-3.6.3.tgz | grep -vE "^package/(esm|cjs|browser)/"
package/package.json
```

`@asyncapi/parser` and `@asyncapi/multi-parser` have nothing but `package.json` at the tarball root.
`@asyncapi/openapi-schema-parser` ships its own `README.md` but no `LICENSE` and no `NOTICE` - so the copy step
here deliberately preserves an existing README rather than overwriting it.

A changeset is included, since without one a merge would not cut a release and the published tarballs would stay
as they are.

Issue #1203, opened 2026-07-15
[MICROGRANT] Set of issues for resolution under the Microgrant Program 2026-09


AayushSaini101:
- https://github.com/asyncapi/parser-js/issues/863
- https://github.com/asyncapi/parser-js/issues/735

justanotherperson00:
Hi! I would like to be considered for #1203 under the 2026 Microgrant Program.

My proposed approach would be:

1. For #1066, reproduce the unsupported custom `schemaFormat` behavior with focused parser tests and trace the distinction between document validation, parser-registry lookup, and schema interpretation.

2. Preserve the existing behavior of built-in and registered custom schema parsers while ensuring that an unsupported but valid string format does not, by itself, invalidate the complete AsyncAPI document.

3. Review the earlier #1066 implementation attempts before choosing the result representation. I would confirm with the maintainers whether unsupported schemas should preserve raw schema data and format metadata, remain intentionally unresolved, or use another established model behavior rather than silently defining a new public API contract.

4. For #863, reproduce the nullable payload property failure directly in parser tests, identify the rule or JSONPath predicate that reads `schemaFormat` from a null or unrelated property named `message`, and apply the narrowest type, null, path, or structural guard that preserves validation of actual AsyncAPI Message Objects.

5. Add regression coverage for unsupported string formats, invalid non-string formats, registered formats, nullable and non-null payload examples, unrelated property names, and relevant v2/v3 behavior.

6. Run the requested parser tests, browser-facing tests where applicable, build, lint, and release checks before opening a focused PR.

Before assignment, could you confirm:

* whether #1203 includes both #1066 and #863;
* the intended parser/model representation for an unsupported schema format;
* the final reward and deadline; and
* whether equivalent v3 coverage is expected for #863?

I have also expressed interest in #1204. I currently have capacity to complete both microgrants if I am selected for both, subject to confirmation of their final scopes and deadlines. I would keep the work in separate focused branches and follow the required progress-update cadence for each assignment.

I will not start implementation or submit a pull request unless I am explicitly assigned the microgrant.

DPS0340:
I would like to be considered for this Microgrant under the 2026-08 program. I will not start implementation or open a PR unless the parser-js maintainers explicitly assign it to me.

My proposed approach:

[...]

PR #1247, opened 2026-09-12
fix: allow payload property named message with null value

### LINKED ISSUE #863: Payload property named message breaks with null value

AayushSaini101:
closes[#863](https://github.com/asyncapi/parser-js/issues/863):
Related: #1203 
 parsing/validating an AsyncAPI 2.x document no longer crashes when a payload schema defines a property named `message` with a `null` example value.

The issue occurred because several v2 Spectral rule JSONPath filters used expressions like:

```ts
[?(@property === 'message' && @.schemaFormat === void 0)]
```

`jsonpath-plus` evaluates these filters while traversing the document tree. Any property named `message` matched the filter — including payload schema fields, not just AsyncAPI Message Objects. When that value was `null`, accessing `@.schemaFormat` threw:

```
Cannot read properties of null (reading 'schemaFormat')
```

This fix adds `!@null` guards before property access on `@`, consistent with other filters already in the ruleset.

## Changes

- Add `!@null &&` guards to 8 JSONPath filters in `packages/parser/src/ruleset/v2/ruleset.ts`
  - `asyncapi2-message-examples`
  - `asyncapi2-schema-default`
  - `asyncapi2-schema-examples`
  - `asyncapi2-message-messageId`
- Add `!@null &&` guards to 2 JSONPath filters in `messageExamples-spectral-rule-v2.ts`
- Add regression tests based on the gist from the issue
- Add integration test in `validate.spec.ts` using `parser.parse()`

## Test plan

- [x] Parse a document with `payload.properties.message` allowing `null` and `examples[0].payload.message: null`
- [x] Confirm no `Cannot read properties of null` diagnostic
- [x] Confirm no `uncaught-error` diagnostic
- [x] Confirm existing rule tests still pass for `asyncapi2-message-examples`, `asyncapi2-schema-examples`, `asyncapi2-schema-default`, and `asyncapi2-message-messageId`

Closes #863

### DIFF
--- a/packages/parser/src/ruleset/v2/functions/messageExamples-spectral-rule-v2.ts
+++ b/packages/parser/src/ruleset/v2/functions/messageExamples-spectral-rule-v2.ts
@@ -17,9 +17,9 @@ export function asyncApi2MessageExamplesParserRule(parser: Parser): RuleDefiniti
     recommended: true,
     given: [
       // messages
-      '$.channels.*.[publish,subscribe][?(@property === \'message\' && @.schemaFormat !== void 0)]',
+      '$.channels.*.[publish,subscribe][?(@property === \'message\' && !@null && @.schemaFormat !== void 0)]',
       '$.channels.*.[publish,subscribe].message.oneOf[?(!@null && @.schemaFormat !== void 0)]',
-      '$.components.channels.*.[publish,subscribe].message[?(@property === \'message\' && @.schemaFormat !== void 0)]',
+      '$.components.channels.*.[publish,subscribe].message[?(@property === \'message\' && !@null && @.schemaFormat !== void 0)]',
       '$.components.channels.*.[publish,subscribe].message.oneOf[?(!@null && @.schemaFormat !== void 0)]',
       '$.components.messages[?(!@null && @.schemaFormat !== void 0)]',
       // message traits
--- a/packages/parser/test/ruleset/rules/v2/asyncapi2-schema-examples.spec.ts
+++ b/packages/parser/test/ruleset/rules/v2/asyncapi2-schema-examples.spec.ts
@@ -396,4 +396,52 @@ testRule('asyncapi2-schema-examples', [
       },
     ],
   },
+
+  // See https://github.com/asyncapi/parser-js/issues/863
+  {
+    name: 'valid case with payload property named message and null example value',
+    document: {
+      asyncapi: '2.4.0',
+      info: {
+        title: 'Messages',
+        version: '1.1.0',
+      },
+      defaultContentType: 'application/json',
+      channels: {
+        'test/message': {
+          subscribe: {
+            message: {
+              $ref: '#/components/messages/transferReturned',
+            },
+          },
+        },
+      },
+      components: {
+        messages: {
+          transferReturned: {
+            title: 'Transfer Returned',
+            payload: {
+              type: 'object',
+              properties: {
+                message: {
+                  type: ['string', 'null'],
+                  maxLength: 50,
+                },
+              },
+              required: ['message'],
+            },
+            examples: [
+              {
+                name: 'return',
+                payload: {
+                  message: null,
+                },
+              },
+            ],
+          },
+        },
+      },
+    },
+    errors: [],
+  },
 ]);
--- a/.changeset/fix-payload-message-null.md
+++ b/.changeset/fix-payload-message-null.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/parser": patch
+---
+
+fix: allow payload property named message with null value (#863)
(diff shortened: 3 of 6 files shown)


aeworxet:
@asyncapi/microgrant_team

packages/parser/src/ruleset/v2/functions/messageExamples-spectral-rule-v2.ts
```diff
@@ -17,9 +17,9 @@ export function asyncApi2MessageExamplesParserRule(parser: Parser): RuleDefiniti
     recommended: true,
     given: [
       // messages
-      '$.channels.*.[publish,subscribe][?(@property === \'message\' && @.schemaFormat !== void 0)]',
+      '$.channels.*.[publish,subscribe][?(@property === \'message\' && !@null && @.schemaFormat !== void 0)]',
       '$.channels.*.[publish,subscribe].message.oneOf[?(!@null && @.schemaFormat !== void 0)]',
-      '$.components.channels.*.[publish,subscribe].message[?(@property === \'message\' && @.schemaFormat !== void 0)]',
+      '$.components.channels.*.[publish,subscribe].message[?(@property === \'message\' && !@null && @.schemaFormat !== void 0)]',
```
princerajpoot20, line 22: This selector still never matches a message under `components.channels`.
The filter is attached after `.message`, so it looks for a child field named `message` inside the message object. That is the payload field, not the message object itself.
I checked with nimma 0.2.3, for a normal `components.channels.*.subscribe.message` that has `schemaFormat` set:
- this path returns no hits, with or without `!@null`
- the same path without the extra `.message` returns that message

Could you please have a look at it once and let us know

packages/parser/test/ruleset/rules/v2/asyncapi2-schema-examples.spec.ts
```diff
@@ -396,4 +396,52 @@ testRule('asyncapi2-schema-examples', [
       },
     ],
   },
+
+  // See https://github.com/asyncapi/parser-js/issues/863
+  {
+    name: 'valid case with payload property named message and null example value',
```
princerajpoot20, line 402: This case passes on the unfixed code, so it does not lock in the fix.
`asyncapi2-schema-examples` uses selectors like:
`$.channels[*][publish,subscribe][?(@property === 'message' && @.schemaFormat === void 0)].payload.examples^`

I ran that expression without `!@null` through nimma 0.2.3 against this document and it does not throw. The crash comes from selectors that end at the filter, such as the one in `asyncapi2-message-examples`:
`$.channels.*.[publish,subscribe][?(@property === 'message' && @.schemaFormat === void 0)]`
This fixture also puts `examples` on the message. This rule reads `payload.examples` (examples on the schema). Message examples are already covered by `asyncapi2-message-examples.spec.ts` and by the new case in `validate.spec.ts`, and those do exercise the crash

could you have a look at it and let us know

